PR #8, opened 2020-04-03
feat: provide responsive grid system

feedm3:
Provide a 12 column grid.

The grid container can be created with `<Grid />`, each item with `<GridItem />`. A demo can be viewed on `http://localhost:8000/grid-test`.

### DIFF
--- a/src/components/grid/grid.tsx
+++ b/src/components/grid/grid.tsx
@@ -1,26 +1,85 @@
 import styled, { CSSProp } from 'styled-components';
 import { up } from '../breakpoint/breakpoint';
 
+type GridItemSize = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+
 const GRID_GAP_DESKTOP: CSSProp = '24px';
 const GRID_GAP_MOBILE: CSSProp = '12px';
 
-export const Grid = styled.div`
-  max-width: 1280px;
-  margin: 0 auto;
-  padding: 0 ${GRID_GAP_MOBILE}; // outer gap
+interface GridProps {
+  disableGap?: boolean;
+}
 
+/**
+ * This container provides a grid that has 12 columns.
+ *
+ * Grids can also be nested, as well as their gap can be disabled via prop.
+ */
+export const Grid = styled.div<GridProps>`
   display: grid;
   grid-template-columns: repeat(12, 1fr);
-  grid-column-gap: ${GRID_GAP_MOBILE};
+
+  grid-column-gap: ${(props) => (props.disableGap ? 0 : GRID_GAP_MOBILE)};
+  padding: 0 ${(props) => (props.disableGap ? 0 : GRID_GAP_MOBILE)}; // outer gap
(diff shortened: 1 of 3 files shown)


src/components/grid/grid.tsx
```diff
@@ -1,26 +1,85 @@
 import styled, { CSSProp } from 'styled-components';
 import { up } from '../breakpoint/breakpoint';
 
+type GridItemSize = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+
 const GRID_GAP_DESKTOP: CSSProp = '24px';
 const GRID_GAP_MOBILE: CSSProp = '12px';
```
miffili, line 7: Where are these values coming from? The notion suggests the value is not changeable, in the design however I see no *one* gap value for desktop and mobile respectively.

src/components/grid/grid.tsx
```diff
@@ -1,26 +1,85 @@
...
+}
+
+/**
+ * The grid item can be placed inside the Grid component. If the amount of items in one row is larger then 12, the
+ * overflowing grid item will be placed in the next row.
+ *
+ * By default, a grid item takes the whole row (12 columns) on xs.
+ */
+export const GridItem = styled.div<GridItemProps>`
+  /* on mobile an item takes the whole row itself by default */
```
miffili, line 49: Redundant comment, as it's already mentioned in line 48. 🤓

src/pages/grid-test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </Grid>
+    <Grid>
+      <GridItem sm={6}>
+        <Card>xs=12 | sm=6</Card>
+      </GridItem>
+      <GridItem sm={6}>
+        <Card>xs=12 | sm=6</Card>
+      </GridItem>
+      <GridItem sm={12} md={4}>
+        <Card>xs=default | sm=12 | md=3</Card>
```
miffili, line 31: Not sure if this page will be kept in the project or was just meant for illustration purposes.. but the configuration if the Grid Item & the Card content doesn't always match.

If this page is kept for documentation purposes, please check values. Also check if the code could be slimmed down/cleaned up. Otherwise it's a lot of repeating code.

miffili (approved):
Nice, thorough change. 👍

erinfm (approved):

src/components/grid/grid.tsx
feedm3: They are currently not used. I just added it as other grid implementations (like the one from material-ui) also allow to set a grid spacing. If we figure out that we don't need it, we will throw it out.

src/pages/grid-test.tsx
feedm3: This page will be thrown out as soon as we have a component documentation like storybook or docz. I added it to have a visual feedback if the grid is even working 🙈

src/components/grid/grid.tsx
feedm3: removed

PR #9, opened 2020-04-24
feat: add headings to homepage

feedm3:
This PR:
- introduces headings on the front page
- add some placeholder for the other components
- uses the grid on the page to give the components the correct width
- throws out the grid test page

<details>
 <summary>Mobile</summary>
Bildschirmfoto 2020-04-24 um 10 53 12
</details>

<details>
 <summary>Desktop</summary>
Bildschirmfoto 2020-04-24 um 10 53 03
</details>


src/pages/index.tsx
```diff
@@ -3,34 +3,75 @@ import React from 'react';
...
+  margin-top: 80px;
+  margin-bottom: 80px;
+
+  ${up('md')} {
+    margin-top: 160px;
+    margin-bottom: 120px;
+  }
+`;
+
+const DesktopOnlyBreak = styled.br`
```
georgiee: I would remove this. We don't need responsive breaks. That's an antipattern. You can't control them if the text is coming from any external source anyway. This is just too much.

src/components/typography/typography.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { HTMLAttributes } from 'react';
+import styled from 'styled-components';
+import { up } from '../breakpoint/breakpoint';
+
+const H1 = styled.h1`
```
georgiee: I think we should stay semantic with the headlines. Clarify how to name H1/H2. It would be much more readable to if they are named like "PageTitle". "SectionTitle" etc. h1/h2 are kind of implementation details although we need to ensure a correct order for accessibility later of course.

src/pages/index.tsx
```diff
@@ -3,34 +3,75 @@ import React from 'react';
...
+      <GridItem xs={0} md={2} />
+      <GridItem xs={12} md={8}>
+        <Title>
+          <Typography variant="h1">
+            Full Stack <br />
+            Digital Service <br />
+            Agency
+          </Typography>
+        </Title>
+        <Subtitle>
```
georgiee: Hmm Subtitle being a div and then the nested Typography tag is weird. I see why you have h1/h2 now as you used the "semantic" titles here. But something is off here?

src/pages/index.tsx
```diff
@@ -3,34 +3,75 @@ import React from 'react';
...
-            Prozess von Ideation bis zur Implementierung des letzten
-            performanten Funnels und der letzten Zeile wunderschönen Codes.
-          </div>
-          <button>All Services</button>
-        </div>
-        <p>We are into relationships, not one-night-stands.</p>
+      <GridItem xs={0} md={2} />
+      <GridItem xs={12} md={8}>
+        <Title>
+          <Typography variant="h1">
```
georgiee: What's the goal of "Typography" ? Only Headlines? any other tag? I don't know if this is really readable and scaleable. This could be an internal generalized tool but I would have expected something like:

```
<Title>ABC</Title>
<Subtitle>CDE</Subtitle>
```

Whatever you need to do in Title. But it's much more readable., don't you think?

src/components/typography/typography.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const Body = styled.div``;
+
+type Variant = 'h1' | 'h2';
+
+interface TypographyProps extends HTMLAttributes<{}> {
+  variant: Variant;
+}
+
+export const Typography: React.FC<TypographyProps> = (props) => {
```
georgiee: I don't think such a "Factory" makes sense. You could just import the according component instead.
What's the purpose of this?

georgiee (changes requested):
Thanks for that. I have some comments 🙌

src/pages/index.tsx
```diff
@@ -3,34 +3,75 @@ import React from 'react';
...
+            relationships, not <DesktopOnlyBreak />
+            one-night-stands.
+          </Typography>
+        </Subtitle>
+        <p>Card</p>
+        <p>List</p>
+        <Subtitle>
+          <Typography variant="h2">
+            We are on your <DesktopOnlyBreak />
+            side - and the <DesktopOnlyBreak />
```
erinfm: do we need all the line breaks? could we get the the same visual result with fixed widths on the Subtitle and Title components? and would this have a11y implications?

erinfm (approved):
i'm not 100% on if we need all the line breaks in the headings (are there any screenreader implications there, for example?)

src/pages/index.tsx
feedm3: tested it: it breaks the voice over, as it reads "We are on your **GROUP** side...."

src/pages/index.tsx
feedm3: Wanted to make our designers super happy :D But yeah, it also breaks accessibility so it doesn't work anyway.

src/components/typography/typography.tsx
feedm3: Good idea, done.

src/pages/index.tsx
feedm3: 👍

src/components/typography/typography.tsx
feedm3: I took the typography component from material-ui as role model, but at this development stage it's probably simpler to make separate components each. Done.

src/pages/index.tsx
feedm3: resolved within the other comments

src/pages/index.tsx
```diff
@@ -3,34 +3,45 @@ import React from 'react';
 import Layout from '../components/layout/layout';
 import SEO from '../components/seo';
 import { Grid, GridItem } from '../components/grid/grid';
+import { PageTitle, SectionTitle } from '../components/typography/typography';
+import styled from 'styled-components';
+import { up } from '../components/breakpoint/breakpoint';
+
+const Title = styled(PageTitle)`
```
georgiee, line 10: Confusing as you mix general styles (PageTitle) with local components that sound similar general (Title). But good for a first round 💪

georgiee (approved):
Let's merge this 👌

PR #12, opened 2020-05-08
feat: add placeholder pages

feedm3:
This pull request will add an empty placeholder page to all sites we need.

This will be the basis for the pages and template. It also has the nice sideeffect that the GatsbyLink will not print errors into the console, when pages do not exists to a link.


erinfm (approved):
cool 👍. guess we will also need impressum and datenschutz pages at some point. we could add those placeholders too?

feedm3:
> cool 👍. guess we will also need impressum and datenschutz pages at some point. we could add those placeholders too?

Yep, will add 💪

PR #17, opened 2020-05-15
feat: add client list component

miffili:
Code changes/additions include:

✓ add **client list** component to project
✓ test coverage for **client list**
✓ implement on index page (replacing placeholder)
✓ add some colours to theme palette (incl. types)

### DIFF
--- a/src/@types/styled-components.d.ts
+++ b/src/@types/styled-components.d.ts
@@ -20,6 +20,11 @@ declare module 'styled-components' {
         defaultLight: string;
         header: string;
         headerLight: string;
+        darkLinkColor: {
+          default: string;
+          hover: string;
+        };
+        darkDefault: string;
       };
       background: {
         bodyLight: string;
--- a/src/components/client-list/client-list.tsx
+++ b/src/components/client-list/client-list.tsx
@@ -1,0 +1,153 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import { up } from '../breakpoint/breakpoint';
+import { Link } from 'gatsby';
+import { theme } from '../layout/theme';
+
+interface ClientListProps {
+  clients: [ClientListEntryProps];
+}
+
+interface ClientListEntryProps {
+  title: string;
+  timestamp: string;
+  link: string;
+}
+
+const linkStyles = css`
+  color: ${theme.palette.text.darkLinkColor.default};
+  font-size: 20px;
+  line-height: 110%;
+  text-decoration: none;
+
+  &:hover {
+    color: ${theme.palette.text.darkLinkColor.hover};
+  }
+`;
+
+const Wrapper = styled.div`
+  /* 32px = 2 x 16px (padding of page layout) */
+  width: calc(100% + 32px);
+  margin-top: 80px;
+  margin-bottom: 80px;
+  margin-left: -16px;
+
+  ${up('md')} {
+    width: 100%;
+    margin-top: 120px;
+    margin-bottom: 160px;
+    margin-left: 0;
+  }
+`;
+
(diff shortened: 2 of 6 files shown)


src/@types/styled-components.d.ts
```diff
@@ -20,6 +20,12 @@ declare module 'styled-components' {
         defaultLight: string;
         header: string;
         headerLight: string;
+        headerLight: string;
```
feedm3: This was already in, can be removed

src/components/client-list/client-list.tsx
```diff
@@ -0,0 +1,146 @@
...
+  line-height: 110%;
+  text-decoration: none;
+
+  &:hover {
+    color: ${theme.palette.text.darkLinkColor.hover};
+  }
+`;
+
+const Wrapper = styled.div`
+  width: 100vw;
```
feedm3: It looks like `100vw` ignores the scroll bar. Don't ask me why 🙈 If you make the width `calc(100% + 32px)` it also works with the scroll bar.

image

feedm3 (changes requested):
Looks great! 🙂  Especially the tests are awesome 🎉

src/@types/styled-components.d.ts
miffili: Whoopsies...

src/components/client-list/client-list.tsx
miffili: What do you mean with "ignores the scrollbar"? That the chevrons are almost touching the scroll bar?

EDIT: Fixed it under the assumption, you meant that. 🤓 By giving the chevron a bit more space to the right. Since checking it with mobile devices (in chrome) & other browsers seems fine so far.

feedm3: Ah, you are right, on the browsers mobile view it works. But if you use the desktop view and just make the window smaller, there is now a vertical scroll bar. This is becuase of the `width: 100vw`.

<details>
 <summary>width: 100vw</summary>
CleanShot 2020-05-26 at 09 07 41@2x
</details>

<details>
 <summary>width: calc(100% + 32px)</summary>
CleanShot 2020-05-26 at 09 06 41@2x
</details>

miffili: You mean the horizontal scroll bar? Where does this calculation come from?
To be honest a width of `100vw` is unlikely to cause a horizontal scroll bar.

EDIT: And testing it with Chrome, Firefox & Edge, I don't get a horizontal scroll bar.

@erinfm @georgiee could you please check? 🙂

erinfm: i don't see the horizontal scroll bar either :)

erinfm: i retract my previous statement 😅: i get the horizontal scroll bar at smaller widths, while using a mouse (not trackpad)

Screenshot 2020-05-29 at 12 24 21

feedm3: Very crazy behaviour: When using a mouse, Chrome treats scroll bars differently then when using trackpad. There are also a bunch of articles online about this: https://jaketrent.com/post/100vw-not-scrollbar-aware/

I would recommend using `width: calc(100% + 32px)` as this "just works".

miffili: But the calculation seams really arbitrary.. where does this come from?

feedm3: 100% = grid width, 32px = grid padding on the left + grid padding on the right (16px each)

miffili: Done.

feedm3 (approved):

PR #22, opened 2020-06-19
Feat: Tabs component SYW-008

erinfm:
This PR adds the tabs, and tab content, for the Services page. The tabs use Reach UI as a base.

Features:
- The selected tab is highlighted
- On mobile the tab overflow is scrollable.

Notes: Currently, the tab hash is not implemented.

<details>
<summary>Mobile</summary>
Screenshot 2020-06-19 at 08 47 13

</details>

<details>
<summary>Mobile (scrolled)</summary>

Screenshot 2020-06-19 at 08 48 25
</details>

<details>
<summary>Desktop</summary>

Screenshot 2020-06-19 at 08 48 43
</details>


src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,49 @@
...
+  transform: translateX(-14px);
+  white-space: nowrap;
+  width: 100vw;
+
+  ${up('md')} {
+    height: 100%;
+  }
+`;
+
+export const StyledTab = styled(Tab)`
```
feedm3: This file can be removed, as the components here are never used.

src/components/tabs/services-tabs.tsx
```diff
@@ -0,0 +1,143 @@
...
+  margin-bottom: 12px;
+`;
+
+const StyledTabList = styled(TabList)`
+  background: none;
+  padding-bottom: 18px;
+  overflow-x: scroll;
+  transform: translateX(-14px);
+  white-space: nowrap;
+  width: 100vw;
```
feedm3: `100vw` often makes problems as it doesn't work that well with our grid (on desktop with mouse). Better use`100%`, should also work (if even necessary)

src/components/tabs/services-tabs.tsx
```diff
@@ -0,0 +1,143 @@
...
+`;
+
+const StyledLi = styled.li`
+  margin-bottom: 12px;
+`;
+
+const StyledTabList = styled(TabList)`
+  background: none;
+  padding-bottom: 18px;
+  overflow-x: scroll;
```
feedm3: `scroll` shows a constant scroll bar, even if it is not used (desktop with mouse). `auto` should fix that problem

CleanShot 2020-06-19 at 11 44 49@2x

src/components/tabs/services-tabs.tsx
```diff
@@ -0,0 +1,143 @@
...
+      border-radius: 1px;
+    }
+  }
+`;
+
+const StyledTabPanels = styled(TabPanels)`
+  line-height: 1.5;
+`;
+
+const ServicesTabs: React.FC = () => {
```
feedm3, line 69: It's better to separate the presentational component and the data it should show as it makes the component reusable. So passing in the actual tabTitles and tabContents as properties instead of hardcoding.

Would look like this then:
```
interface ServicesTabsProps {
  tabs: {
    title: string;
    content: string;
  }[];
}

const ServicesTabs: React.FC<ServicesTabsProps> = (props) => {
  return (
    <Tabs defaultIndex={0}>
      <StyledTabList>
        {props.tabs.map((tab) => (
          <StyledTab key={tab.title}>{tab.title}</StyledTab>
        ))}
        ...
```

Let's make a separate ticket for this improvement :)

feedm3 (changes requested):

src/components/tabs/services-tabs.tsx
erinfm: the 100vw (and the transform) are there so that the tabs stretch the full page width on mobile as in the design. if we change it to 100% or remove it, the mobile tabs look like this:

Screenshot 2020-06-19 at 13 42 00

vs design: 

Screenshot 2020-06-19 at 13 45 37

what's the best way to implement that design, without the vw?

src/components/tabs/services-tabs.tsx
erinfm: thanks for the catch, trackpad vs mouse strikes again 😅

src/components/tabs/services-tabs.tsx
erinfm: agreed, making it into a reusable component will be much better :)

feedm3 (approved):

PR #23, opened 2020-06-19
feat: use cloudinary image for blog posts

feedm3:
To be able to add images to a blog post, this PR integrated Cloudinary.

Preview URL: https://deploy-preview-23--satellytes-website-new.netlify.app/blog/how-to-do-a-blog-post

Todos:
- add an effect to only load necessary width of an image
  - if the browser gets resized, only load an image if the browser was resized larger
  - debounce reload to not load an image every millisecond
- add description in Readme on how to use
- add warning if image url in markdown is not a cloudinary url

### DIFF
--- a/src/components/markdown/renderers.tsx
+++ b/src/components/markdown/renderers.tsx
@@ -71,3 +71,29 @@ export const InlineCodeRenderer: React.FC<ReactMarkdownCodeRendererProps> = (
 ) => {
   return <InlineCode>{props.value}</InlineCode>;
 };
+
+/*
+ * Image
+ *
+ */
+const StyledImage = styled.img`
+  width: 100%;
+`;
+
+interface ReactMarkdownImageProps {
+  src: string;
+  alt: string;
+}
+
+export const CloudinaryImageRenderer: React.FC<ReactMarkdownImageProps> = (
+  props,
+) => {
+  // todo:
+  // - useEffect to check the what image width is needed
+
+  const srcParts = props.src.split('/');
+  const imageId = srcParts[srcParts.length - 1];
+  const transformations = 'w_814';
+  const src = `https://res.cloudinary.com/satellytes/image/upload/${transformations}/satellytes-website/${imageId}`;
+  return <StyledImage src={src} alt={props.alt} />;
+};
--- a/src/components/markdown/renderers.tsx
+++ b/src/components/markdown/renderers.tsx
@@ -1,8 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { SectionTitle, SubTitle, Text } from '../typography/typography';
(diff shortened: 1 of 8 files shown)


src/components/markdown/renderers.tsx
```diff
@@ -71,3 +74,70 @@ export const InlineCodeRenderer: React.FC<ReactMarkdownCodeRendererProps> = (
...
+
+const isCloudinaryUrl = (src: string): boolean => {
+  const url = new URL(src);
+  return (
+    url.hostname.includes('cloudinary.com') &&
+    url.pathname.startsWith('/satellytes')
+  );
+};
+
+const transformCloudinaryUrl = (src: string, width: number): string => {
```
georgiee: cloudinary sounded so magic but here we go, that's everything need 🙌

src/components/markdown/renderers.tsx
```diff
@@ -71,3 +74,70 @@ export const InlineCodeRenderer: React.FC<ReactMarkdownCodeRendererProps> = (
...
+
+/*
+ * Image
+ *
+ */
+const StyledImage = styled.img`
+  width: 100%;
+`;
+
+const isCloudinaryUrl = (src: string): boolean => {
```
georgiee: this is kind of cluttering the renderer file. I think everything around the cloudinary would fit in a separate file very good?

src/components/markdown/renderers.tsx
```diff
@@ -71,3 +74,70 @@ export const InlineCodeRenderer: React.FC<ReactMarkdownCodeRendererProps> = (
...
+      // we only change the width if the picture gets better/larger
+      if (newImageWidth > imageWidth) {
+        setImageWidth(newImageWidth);
+      }
+    },
+    250,
+    [currentWindowSize],
+  );
+
+  const src = transformCloudinaryUrl(props.src, imageWidth);
```
georgiee, line 125: is it a good idea to pass any given size? will cloudinary handle this or create dozens of different sizes when you have a width of 100,101,102 etc?

georgiee (approved):
Thanks for this. We already serve images, is cloudinary useful for all the different transforms instead of resizing locally? Is that why you use it here? 

I left a few comments but otherwise good with really few bits of code 👍

src/components/markdown/renderers.tsx
feedm3: I debounced it: It will only create a new URL after 250ms and it will also only create a new URL if the image width gets larger (once you have the large image the browser can also use it when you resize to a smaller width)

src/components/markdown/renderers.tsx
feedm3: good idea, makes it also simpler for testing 👍

PR #27, opened 2020-06-24
Feat: Make tab component more reusable

erinfm:
Code additions/changes:
- Services tabs now take data via props
- Tab panel content is now wrapped in Markdown tags, to support complex tab content (with bullet points, etc)


feedm3 (approved):
Really nice 👍  Next step would be to get the data out of the services page into it's own file, but let's do this later on in a new PR

PR #30, opened 2020-06-30
feat: add blog-cards & layout blog overview

miffili:
* add blog-card (incl. all shenanigans)
  * hover-state was discussed privately in slack
  * ~'vor 5 Tagen' not realised, instead _today_ & _yesterday_ get displayed, otherwise the~ date in a pretty format
* test blog-card
* placeholderImage for blog-cards (preview images are a whole other story
* blog overview layout

_note_: Swoosh is added in [another PR](https://github.com/satellytes/satellytes.com-new/pull/29).

### URL

https://deploy-preview-30--satellytes-website-new.netlify.app/blog/

### Screenshots (quick look)

<details>
 <summary>Mobile</summary>
Blog – Mobile
</details>

<details>
 <summary>Desktop
</summary>
Blog – Desktop
</details>


feedm3:
Added the IntelliJ Github integration on my machine. "Close" button works better than expected 🙈 😄

src/components/cards/blog-card.tsx
```diff
@@ -0,0 +1,150 @@
+import React from 'react';
+import GatsbyImage, { FluidObject } from 'gatsby-image';
+import styled, { css } from 'styled-components';
+import { Link } from 'gatsby';
+import { theme } from '../layout/theme';
+import { up } from '../breakpoint/breakpoint';
+import { CardProps, CardWrapper, CardTitle, CardText } from './card';
+
+type sizes = 'S' | 'L';
```
feedm3: 1. Types should be uppercase, like interfaces and class names
2. If you use it like an enum, it should be Singular. It's easier to read `Size.S` than `Sizes.S`, as it always referes to one item, not multiple ones.

src/components/cards/blog-card.tsx
```diff
@@ -0,0 +1,150 @@
...
+  ${(props) =>
+    props.image &&
+    css`
+      padding: 24px 16px;
+    `}
+`;
+
+const BlogCardImage = styled(GatsbyImage)`
+  width: calc(100% + 2 * 16px);
+  margin: -24px -16px 0;
```
feedm3: Never seen margin with only 3 values. Would probably be simpler to read if it gets merged with the `margin-bottom` bellow to a "standard" 4 value margin.

src/pages/blog.tsx
```diff
@@ -1,34 +1,91 @@
...
 
+const BlogPageTitle = styled(PageTitle)`
+  margin-bottom: 48px;
+
+  ${up('md')} {
+    margin-bottom: 80px;
+  }
+`;
+
+const getDateString = (date: string): string => {
```
feedm3: This function is quite complicated to understand, here are some tips:

1. Dates are always a hassle. Especially if you also consider that different browsers treat them differently. Therefore, using a date library like `date-fns` is very much appreciated. They have simple and easy to read helper methods. So I would suggest you add this library to this project and use those. Example:
1.1 You could replace the whole method with https://date-fns.org/v2.14.0/docs/formatRelative (not sure if that does exactly what we need)
1.2 If the `formatRelative` doesn't work/ fit, you can replace the calculation of `isToday` and `isYesterday` with https://date-fns.org/v2.14.0/docs/isYesterday and https://date-fns.org/v2.14.0/docs/isToday.

2. The naming of "getDateString" is quite general. Maybe a more specific name like "formatToRelative" is more clear.

3. The `date` parameter is of type string. As a date can have a bunch of different string representations (ISO or custom format like `dd-mm-yyyy`) it should be made clear in a comment above what format is expected.

feedm3 (changes requested):
The cards look very good on all devices, good job 🥳  Just some technical improvements as feedback to make this feature "round"

src/components/cards/blog-card.tsx
Miffili: I disagree. This form is a common shorthand.
https://developer.mozilla.org/de/docs/Web/CSS/margin

src/pages/blog.tsx
miffili: 1. I disagree. To pull in a library to handle one date is overkill, when this function achieves the same result.
2. Sure, I'll rename the function, but `formatToRelative` is not entirely correct, because the method not only determines "Gestern" & "Heute", which wasn't even asked for. But also parses the date to an understandable, nicely readable date format. In detail yyyy-mm-dd is parsed to 01. August 2020, as the input date format, can be confusing (US/BE).
3. Okay.

src/components/cards/blog-card.tsx
miffili: Can you please explain point 2 in more detail? So I know, what exactly to change.

src/components/cards/blog-card.tsx
miffili: Thanks for pointing out the `margin-bottom` below.

src/pages/blog.tsx
miffili: _Edit_: I simplified the method. In my honest opinion, this was somewhat of a suggestion (not required) and isn't of importance at the moment. 🙂

src/components/cards/blog-card.tsx
miffili: Nevermind. Fixed the issue.

feedm3 (approved):

src/pages/blog.tsx
feedm3: Yeah, there should always be valid reasons to import a library. For dates, most of the use cases are a valid reason to import one. Even with the simple method here, `date-fns` parse method would be better. Why? Because using a string to initiate a date is highly discourage (see "3. Timestamp string" of the spec: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date). 

Technically, it's not an overkill. As date-fns uses ES6 modules, only the functions that are imported will be used, all other code of the library is untouched. Also test coverage and documentation will probably most of the time better with such (let's call it) "high quality libraries". 

But as for our simple usecase everything is working fine I'm good to go 👍

PR #31, opened 2020-07-01
feat: enhance career page & add career grid

miffili:
* add content to career page
* implement design/layout
* simply mock data for career grid
* add career grid, using cards

https://deploy-preview-31--satellytes-website-new.netlify.app/career/

### DIFF
--- a/src/pages/career.tsx
+++ b/src/pages/career.tsx
@@ -1,29 +1,136 @@
 import React from 'react';
+import styled from 'styled-components';
 
-import Layout from '../components/layout/layout';
 import SEO from '../components/seo';
-import { PageTitle } from '../components/typography/typography';
+import Layout from '../components/layout/layout';
+import { theme } from '../components/layout/theme';
+import { up } from '../components/breakpoint/breakpoint';
+import {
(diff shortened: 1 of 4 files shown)


src/pages/career.tsx
```diff
@@ -1,29 +1,138 @@
...
+        </GridItem>
+
+        <GridItem xs={12}>
+          <CareerGrid careerData={careerData} />
+        </GridItem>
+
+        <GridItem xs={12}>
+          <Note>
+            Oder schicke deine Bewerbung an:{' '}
+            <StyledLink to="mailto:team@satellytes.com">
```
erinfm: think this needs to be a standard a tag/external link, because isn't Gatsby Link only for internal page links?

erinfm (approved):
i think the mailto link at the bottom need to be changed to a standard external link. then we're good to go 🥳

src/pages/career.tsx
miffili: 🙏
I took the mailto out for now, as we actually don't know what is expected here & since mailto isn't necessarily best practise.

PR #36, opened 2020-07-10
feat: add contact page

feedm3:
This PR will add the contact page. This also includes the contact form.

- For the contact form, the Netlify forms feature will be used

Preview URL: https://deploy-preview-36--satellytes-website-new.netlify.app/contact

Screenshots:
<details>
 <summary>Full Desktop</summary>
CleanShot 2020-07-10 at 15 52 15@2x
</details>

<details>
 <summary>Full Mobile</summary>
CleanShot 2020-07-10 at 15 52 15@2x
</details>

<details>
 <summary>Form empty state</summary>
CleanShot 2020-07-10 at 15 52 15@2x
</details>

<details>
 <summary>Form valid state</summary>
CleanShot 2020-07-10 at 15 53 21@2x
</details>

<details>
 <summary>Form success state</summary>
CleanShot 2020-07-10 at 15 46 42@2x
</details>

<details>
 <summary>Netlify dashboard</summary>
CleanShot 2020-07-10 at 15 54 52@2x
</details>


src/components/form/contact.tsx
```diff
@@ -0,0 +1,170 @@
+import React, { ChangeEventHandler, FormEventHandler, useState } from 'react';
+import styled from 'styled-components';
+import { Text } from '../typography/typography';
+
+const InputContainer = styled.div`
+  display: flex;
+  flex-direction: row;
```
miffili: By default, flex-direction is always row, so this is a bit redundant.

src/components/form/contact.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return (
+    <form
+      name="contact"
+      method="POST"
+      data-netlify="true"
+      onSubmit={handleSubmit}
+    >
+      <InputContainer>
+        <Input
+          placeholder="Your name"
```
miffili, line 140: This is a weird Denglisch mix on the whole website. Since the whole contact page is in German, I would change these to German.

src/components/form/contact.tsx
```diff
@@ -0,0 +1,170 @@
...
+        />
+        <Input
+          placeholder="Your email"
+          type="email"
+          name="email"
+          value={formData.email}
+          onChange={handleInputChange}
+        />
+      </InputContainer>
+      <TextArea
```
miffili, line 154: I would only allow resizing in vertical direction, as it can be messy in the FE, if it is arbitrarily possible.

src/components/form/contact.tsx
georgiee: Do we have an agreement on the default language to begin with? Current website is German so I would also try to put German coy in all places 🙏 Makes it easier to translate later too.

src/components/form/contact.tsx
```diff
@@ -0,0 +1,170 @@
...
+        name="message"
+        value={formData.message}
+        onChange={handleInputChange}
+      />
+      <div>
+        <SendButton
+          disabled={!hasValidInput()}
+          type="submit"
+          title={
+            hasValidInput()
```
georgiee, line 165: I can't see any validation happening (Chrome/Firefox). No message and the submit button also seems to be not clickable. Is this working in the preview for you?

georgiee: Oh lord, now I see it. The button is disabled (barely visible to me) and only active when all fields are filled. I would prefer to have an appearing messsage the moment I click on the submit button. Is this not in the design? Is this expected like this? I think that's pretty confusing. Can you tackle it within this PR or later?

src/components/header/header.tsx
```diff
@@ -4,17 +4,20 @@ import { up } from '../breakpoint/breakpoint';
...
-const StyledHeader = styled.header<{ light: boolean }>`
+const StyledHeader = styled.header<{ light: number }>`
   height: ${HEADER_HEIGHT};
   display: flex;
   justify-content: space-between;
   align-items: center;
   padding: 0 16px;
   border-bottom: ${(props) =>
-    props.light
+    props.light === 1
```
georgiee, line 20: I'm curious why the change to a number ? I didn't expect it to be a number.

src/components/header/hero-image.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const StyledBackgroundImage = styled(BackgroundImage)`
+  height: calc(344px - ${() => HEADER_HEIGHT});
+
+  ${up('md')} {
+    height: calc(560px - ${() => HEADER_HEIGHT});
+  }
+`;
+
+export const HeroImage: React.FC<{ fluid?: FluidObject }> = (props) => {
```
georgiee, line 16: background works 👍

src/components/layout/layout.tsx
```diff
@@ -44,7 +46,8 @@ const Layout: React.FC<LayoutProps> = (props) => {
       <GlobalStyle isIndexPage={props.isIndexPage} />
       <Header
         siteTitle={data.site.siteMetadata.title}
-        light={props.isIndexPage}
+        light={props.isIndexPage || Boolean(props.heroImage)}
```
georgiee, line 49: light is a number and is given a boolean 🤔

src/pages/contact.tsx
```diff
@@ -2,16 +2,86 @@ import React from 'react';
...
+    <Layout heroImage={data.imagePlaceholder.childImageSharp.fluid}>
       <SEO title="Contact" />
-      <Grid>
-        <GridItem>
-          <PageTitle>Contact</PageTitle>
+      <Grid center>
+        <GridItem xs={0} md={2} />
+        <GridItem xs={12} md={8}>
+          <ContactSubTitle short>Adresse</ContactSubTitle>
+          <div>
```
georgiee, line 45: For my clarification. This will be replaced by markdown in the future and mapped to the very same components you use here. Is this correct?

georgiee (approved):
Thanks for this 👍
The form doesn't fit on mobile on my side. How did you make the screenshot of mobile and in which size?

Screenshot 2020-07-14 at 13 47 07

I would have expected a single column instead the two inputs share a line and that's the reason why it doesn't fit. Is this the design that imposes this layout? We don't have much horizontal space so I would like to see it distribute vertically.

I approve this PR as it works as promised and brings the basic tech. There are some details I marked. Feel free to tackle them here or in following PRs. I'm happy to review any addition within this PR. Thanks!

src/components/form/contact.tsx
feedm3: First version will go live with english only. Text is currently in progress, once we have the final version we will update all text. Until then I use the text from figma.

src/components/form/contact.tsx
feedm3: Go catch, will remove it :)

src/components/form/contact.tsx
feedm3: Nice, didn't thought about that. Will do 👍

src/components/form/contact.tsx
feedm3: 1. Single column: Yep, makes sense. Will do that. For you error: The design doesn't break on Chrome, but on Firefox only. It even works on Safari 🧐  But as we change to single column we can ignore that problem.
2. States: There are no states in the design so I tried my best and improvised :D I will leave it as is for now, create a follow up ticket and ask for a proper design

src/components/header/header.tsx
feedm3: If you use a styled component with the boolean you get this error: https://github.com/styled-components/styled-components/issues/1198.

Interestingly, the error only appears with the `SiteTitle` component in this file, and not the `StyledHeader` where I did this change 🤷 . But to have a bit more consistency in this file, I still changed the prop here to a number.

src/components/layout/layout.tsx
feedm3: See answer above: Number only needed for styled components.

src/pages/contact.tsx
feedm3: Yes, you are totally right.

PR #38, opened 2020-08-06
Post signature

katerynabugaieva:
This PR about adding signature for posts and date

Update: add author summary (design will be corrected by Verena soon)
Screenshot 2020-08-07 at 12 13 31


georgiee:
Just checked the errors. Install fails because of npm https://status.npmjs.org/incidents/cksjqc1w11v5

src/components/signature/signature.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    author: 'John Doe',
+    date: new Date().toString(),
+  },
+  {
+    author: 'John Doe',
+    date: '2020-08-01',
+  },
+];
+
+describe(Signature.name, () => {
```
georgiee: Please use a static name I don't see the point of using some variable reference here. Makes it hard to find in the output.

src/components/signature/signature.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  describe('days label', () => {
+    it('should render label "Tag" if date today', () => {
+      const { findByText } = render(
+        <Signature author={testData[1].author} date={testData[1].date} />,
+      );
+      expect(findByText('Tag')).toBeTruthy();
+    });
+
+    it('should render label "Tagen" if date earlier than today', () => {
```
georgiee: that is covered by the following tests (as you check for "vor 5 Tagen"). I think you can remove this test.

src/components/signature/signature.tsx
```diff
@@ -0,0 +1,44 @@
+import differenceInCalendarDays from 'date-fns/differenceInCalendarDays';
```
georgiee: our package.json misses the explicit dependency on `date-fns`. Please add it if you need this helper.

src/components/signature/signature.tsx
```diff
@@ -0,0 +1,44 @@
...
+import React from 'react';
+import styled from 'styled-components';
+
+export interface SignatureProps {
+  author: string;
+  date: string;
```
feedm3: If you make the date a string the component needs to know how to parse it. This makes it dependent of the context in which it gets used (and vice versa), so less reusable. 
I would propose to make the property of a type "Date". Then, the component doesn't need to take care of the format at all.

src/components/signature/signature.tsx
```diff
@@ -0,0 +1,44 @@
...
+  font-weight: bold;
+`;
+
+const SignatureDaysPassed = styled.p`
+  font-weight: normal;
+`;
+
+const Signature: React.FC<SignatureProps> = ({ author, date }) => {
+  const daysPassed = differenceInCalendarDays(new Date(), new Date(date));
+  const daysText = daysPassed > 2 ? 'Tagen' : 'Tag';
```
feedm3: If `daysPassed` is 0, it should be "Heute" and without the `vor `

src/components/signature/signature.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+      );
+      expect(findByText('Tagen')).toBeTruthy();
+    });
+
+    it('should render correct number of days', () => {
+      const { getByText } = render(
+        <Signature author={testData[2].author} date={testData[2].date} />,
+      );
+      const date = getByText('vor 5 Tagen');
+      expect(date).toBeTruthy();
```
feedm3: This test will only work today ;-) 

You can use the calculation method you already use in the component to get the number of days, and check if the component returns the same amount.

feedm3 (changes requested):
Well done, I like it 👍  The structure of the component is simple and clear and the tests are good. Just some small improvements, then it's good to go 🚀

src/templates/blog-post.tsx
```diff
@@ -8,6 +8,7 @@ import { Markdown } from '../components/markdown/markdown';
 import { SectionTitle } from '../components/typography/typography';
 import styled from 'styled-components';
 import { up } from '../components/breakpoint/breakpoint';
+import Signature from '../components/signature/signature';
```
georgiee: Maybe we can rename from "signature" to "byline" as it's the usual name for the author addition to an article.

georgiee (approved):
Thanks for the change.

Later/Ideas:
I would consider to move the signature (or byline as suggested) to the top (more accessible?) and include an optional summary of the author (Frontend Engineer at Satellytes or so). Can we check if those suggestions make sense? Does this need a design change? Please align with @feedm3

src/components/signature/signature.test.tsx
georgiee: oh I missed that 😬 Actually I would mock the Date.now to a specific date. 
Put this line at the top of the test will allow to nail the date `Date.now = jest.fn(() => new Date(Date.UTC(2000, 0, 1)).valueOf());` wouldn't it ?

feedm3: Can also be done like this, yep. Aditionally, I would put it into `beforeEach` and unmock it in `afterEach` to prevent nasty side effects.
https://github.com/facebook/jest/issues/2234#issuecomment-324868057

src/components/signature/signature.tsx
katerynabugaieva: Thanks for suggestion, I wanted to be consistens, as soon as we have date as a string in .md-files
I would like to keep it as `date: Date | string`

feedm3: The thing is: The signature component should not care about the type of string in the markdown file. They  shouldn't be dependent of each other. It's the job of `blog-template.tsx` to get the data and transform it in the correct format, so you can parse the date there and push it as normal date object to the signature.

src/templates/blog-post.tsx
```diff
@@ -16,6 +17,8 @@ interface BlogArticleTemplateProps {
         date: string;
         title: string;
         image?: string;
+        author: string;
```
feedm3: as the author could be optional it should be typed with `?` (`author?: string`). same for the `authorSummary`

feedm3 (changes requested):

src/templates/blog-post.tsx
katerynabugaieva: sure, thanks, it's nice idea

src/templates/blog-post.tsx
```diff
@@ -44,6 +47,13 @@ const BlogArticleTemplate: React.FC<BlogArticleTemplateProps> = ({ data }) => {
             {data.markdownRemark.frontmatter.title}
           </BlogPostTitle>
           <Markdown data={data.markdownRemark.rawMarkdownBody} />
+          {data.markdownRemark.frontmatter.author && (
+            <Byline
+              author={data.markdownRemark.frontmatter.author}
+              date={new Date(data.markdownRemark.frontmatter.date)}
```
feedm3: please use https://date-fns.org/v2.15.0/docs/parseISO for that, as our date string is not 100% compatible with ISO and the date-fns methods catches that

feedm3 (changes requested):

feedm3 (approved):

PR #39, opened 2020-08-10
Shared panel 

katerynabugaieva:
this PR about adding panel for sharing-buttons for post author.

`react-share` is used


package.json
```diff
@@ -42,6 +42,7 @@
     "react-dom": "16.13.1",
     "react-helmet": "6.1.0",
     "react-markdown": "^4.3.1",
+    "react-share": "^4.2.1",
```
feedm3, line 45: Please install with yarn and remove the `package-lock.json`

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+  cursor: pointer;
+  &:last-of-type {
+    margin-right: 0;
+  }
+`;
+
+export const SharedPanel: React.FC<AuthorCredentialsProps> = ({
+  facebookId,
+}) => {
+  const shareUrl = 'https://satellytes-website-new.netlify.app';
```
feedm3: Share URL will be the URL where the share panel is used, so use `window.location.href`

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const title = 'Satellytes';
```
feedm3: Title will be the title of the current page. You can get this via window object somehow, if I remember correctly

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+  cursor: pointer;
+  &:last-of-type {
+    margin-right: 0;
+  }
+`;
+
+export const SharedPanel: React.FC<AuthorCredentialsProps> = ({
+  facebookId,
+}) => {
+  const shareUrl = typeof window !== `undefined` ? window.location.href : '';
```
feedm3: instead of `typeof window !== 'undefined'` you can also use `process.browser ? ...`. it's a bit more readable

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <SharedPanelContainer>
+      <SharedPanelText>Artikel teilen</SharedPanelText>
+      <SocialLinks>
+        <SocialLinkItem>
+          <WhatsappShareButton url={shareUrl} title={title}>
+            <IconWhatsapp />
+          </WhatsappShareButton>
+        </SocialLinkItem>
+        <SocialLinkItem>
+          <FacebookMessengerShareButton appId={facebookId} url={shareUrl}>
```
feedm3: I would remove this. We probably won't create an app on Facebook Messanger, also the Facebook Share icon below can be used for Facebook stuff

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+} from 'react-share';
+import styled from 'styled-components';
+import { up } from '../breakpoint/breakpoint';
+import { IconEmail } from '../icons/social/email';
+import { IconFacebook } from '../icons/social/facebook';
+import { IconMessenger } from '../icons/social/messenger';
+import { IconTwitter } from '../icons/social/twitter';
+import { IconWhatsapp } from '../icons/social/whatsapp';
+
+export interface AuthorCredentialsProps {
```
feedm3: As this interface defines the props of the SharedPanel, it should be called `SharedPanelProps`

src/components/shared-panel/shared-panel.tsx
```diff
@@ -0,0 +1,92 @@
...
+const SocialLinkItem = styled.li`
+  display: inline-block;
+  margin-right: 16px;
+  cursor: pointer;
+  &:last-of-type {
+    margin-right: 0;
+  }
+`;
+
+export const SharedPanel: React.FC<AuthorCredentialsProps> = ({
```
feedm3: Why `SharedPanel` and not `SharePanel`?

src/components/shared-panel/shared-panel.tsx
katerynabugaieva: cool, thanks, because I was wondering if we want to let users contact the Author directly

src/components/shared-panel/shared-panel.tsx
katerynabugaieva: As far as we don't need facebookId, we can remove it completely 🌻

src/components/shared-panel/shared-panel.tsx
katerynabugaieva: thanks for the idea... will take a look

UPD:
sorry, I have objection:

`process.browser` seems to be  defined by webpack and I am not sure we can rely on it. 

not sure if I understand clearly, how it save the day:
> It also exposes a "browser" member (i.e. process.browser) which is true in this implementation but undefined in node. This can be used in isomorphic code that adjusts it's behavior depending on which environment it's running in.

and `typeof window !== undefined` is not too ugly to me

feedm3 (approved):

PR #41, opened 2020-08-25
Implement correct states for the contact form

katerynabugaieva:
this PR about:

*  add error messages
*  enable `send`-button
* if the user send is a success, the initial empty state is taken with a success message
* if user sends correct form, `send->`button turns into `sent `-button
* if anything changes, `sent` button turns back into `send ->`
* if user clicks on button, and there are some invalid fields, message `Bitte füllen Sie alle Felder aus` appears
*  if user clicks on button, all empty/invalid fields are marked red 

Screenshot 2020-08-26 at 12 44 30
Screenshot 2020-08-26 at 12 46 05

### DIFF
--- a/src/components/form/contact.tsx
+++ b/src/components/form/contact.tsx
@@ -3,6 +3,8 @@ import { useForm } from 'react-hook-form';
 import styled from 'styled-components';
 import { up } from '../breakpoint/breakpoint';
 import { Text } from '../typography/typography';
+import { RightArrowIcon } from '../icons/buttons-icons/right-arrow';
+import { CheckmarkIcon } from '../icons/buttons-icons/checkmark';
 
 const InputContainer = styled.div`
   display: flex;
@@ -51,6 +53,10 @@ const Input = styled.input`
 const InputError = styled(Input)`
   background: #f8cdd5;
   color: #dc052d;
+  ::placeholder {
+    color: #dc052d;
+    opacity: 1;
+  }
 `;
 
 const TextArea = styled.textarea`
@@ -70,6 +76,10 @@ const TextArea = styled.textarea`
 
 const TextAreaError = styled(TextArea)`
   background: #f8cdd5;
+  ::placeholder {
+    color: #dc052d;
+    opacity: 1;
+  }
 `;
 
 const Button = styled.button`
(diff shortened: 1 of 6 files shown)


src/components/form/contact.tsx
```diff
@@ -13,7 +16,24 @@ const InputContainer = styled.div`
...
+    margin-right: 24px;
+
+    ${up('md')} {
+      margin-right: 24px;
+    }
+  }
+`;
+
+interface InputTextProps {
+  hasError?: any;
```
feedm3: Why not a `boolean`?

src/components/form/contact.tsx
```diff
@@ -25,22 +45,30 @@ const Input = styled.input`
...
+    color: #dc052d;
+    ::placeholder {
+      color: #dc052d;
+      opacity: 1;
+    }
+  `}
 `;
 
-const TextArea = styled.textarea`
+const TextArea = styled.textarea<InputTextProps>`
```
feedm3: As you use the `InputTextProps` on multiple places, I would rather call it `ValidationProps` to make it more clear that it's not just about the technical input text. Just an idea. If you like your name more, keep it :)

src/components/form/contact.tsx
```diff
@@ -118,59 +174,88 @@ export const ContactForm: React.FC = () => {
...
+        <InputWrapper>
+          <Input
+            placeholder="Your email"
+            type="email"
+            name="email"
+            value={formData.email}
+            onChange={handleInputChange}
+            ref={register({
+              required: true,
+              pattern: /^(([^<>()\]\\.,;:\s@"]+(\.[^<>()\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
```
feedm3: If we put this into it's own variable `IS_EMAIL_REGEX` it's a bit more clear what this pattern does. Also put a link to where you found it.

src/components/form/contact.tsx
```diff
@@ -50,29 +78,65 @@ const TextArea = styled.textarea`
...
 `;
 
-const SendButton = styled.button`
+const Button = styled.button`
+  margin-top: 40px;
   cursor: pointer;
   padding: 13px 18px;
 
   font-size: 20px;
+  font-family: CocoGothic !important;
```
feedm3: What is the reason to have the font defined here?

feedm3 (changes requested):

feedm3:
Only some small changes, nice work! 💪

feedm3 (approved):

PR #65, opened 2021-03-24
feat: add career pages from personio

feedm3:
First draft of implementing Personio XML API. Code needs some cleanup, but is working for now.

URL: https://satellytescom-featpersonio.gtsb.io/career

**Todos**
- Check meta title and description if it's correct. Maybe we also set that in Personio?
- Check if we need a description field. Like the "short description", but as text before the other sections (like in the design)
- Check hardcoded text. It needs to fit for all job positions, not just developers
- Check hardcoded contact email.

PR #66, opened 2021-03-25
feat: introduce career form (personio flavour)

georgiee:
This is based on the changes in #65 and therefore include those changes too. Let's review the other PR first. You can still checkout the preview though and give some feedback.

Included:
+ Use the excellent `react-form-hooks` ✨
+ I decided to use axios in order to get easily access to the upload progress and a better error handling (compared to fetch). We can also offer a cancellation while the upload is running with axios through their cancelation token.
+ I moved the controls in the existing contact form into a separate file to reuse some of them.
+ Form expects some props around the API connection (company id & access token) plus the job id that is different per job.
+ Collected some components outside of the form itself to make it a little bit more readable. It still feels like a mess and I wonder if this is react as is or if I could improve some parts.

+ Form Behavior:
   + Validation only after click on submit
   + Display a progress bar for the running upload
   + Disable the form while submitting
   + Hide the form when completed and display a proper message
   + Everything is German but we should think about translating it soon to make it more accessible to non-German folks.
   + When an error happens on the API side the form is unlocked again and a `try again` button is displayed. Click it to reveal the submit button again once you have read the error message.

### DIFF
--- a/src/components/career-form/career-components.tsx
+++ b/src/components/career-form/career-components.tsx
@@ -21,7 +21,11 @@ interface InputFieldProps {
   type?: 'text' | 'file';
 }
 
-export const InputField: React.FC<InputFieldProps> = (props) => {
+export const Container = styled.div`
+  margin-top: 32px;
+`;
+
+export const InputField = (props: InputFieldProps) => {
   return (
     <InputContainer>
       <InputWrapper>
--- a/src/components/career-form/career-components.tsx
+++ b/src/components/career-form/career-components.tsx
@@ -21,7 +21,7 @@ interface InputFieldProps {
   type?: 'text' | 'file';
 }
 
-export const Container = styled.div`
+export const CareerFormStyled = styled.form`
   margin-top: 32px;
 `;
 
--- a/src/components/career-form/career-components.tsx
+++ b/src/components/career-form/career-components.tsx
@@ -1,0 +1,124 @@
+import React from 'react';
+import {
+  ButtonText,
+  ErrorMessage,
+  Input,
+  InputContainer,
+  InputWrapper,
+  SendButton,
+  SentButton,
+} from '../form/controls';
+import { CheckmarkIcon } from '../icons/buttons-icons/checkmark';
+import styled from 'styled-components';
+import { rgba } from 'polished';
+import { RightArrowIcon } from '../icons/buttons-icons/right-arrow';
+
+interface InputFieldProps {
+  placeholder: string;
+  name: string;
+  error: any;
+  inputRef: any;
+  type?: 'text' | 'file';
+}
+
+export const InputField: React.FC<InputFieldProps> = (props) => {
+  return (
+    <InputContainer>
+      <InputWrapper>
+        <Input
+          placeholder={props.placeholder}
+          type={props.type ?? 'text'}
+          name={props.name}
+          ref={props.inputRef}
+          hasError={props.error}
+        />
+        <FormError error={props.error} />
+      </InputWrapper>
+    </InputContainer>
+  );
+};
+
+export const FormError = ({ error }) => {
+  if (!error) {
+    return null;
+  }
+
+  return (
+    <ErrorMessage>
+      <span>{error.message}</span>
+    </ErrorMessage>
+  );
+};
+
+export const SuccessMessage = () => {
+  return (
+    <div>
+      <p>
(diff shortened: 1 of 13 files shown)


src/components/career-form/career-components.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+interface InputFieldProps {
+  placeholder: string;
+  name: string;
+  error: any;
+  inputRef: any;
+  type?: 'text' | 'file';
+}
+
+export const InputField: React.FC<InputFieldProps> = (props) => {
```
feedm3: `React.FC` is outdated, typing the props and return type separate is now how things go: https://github.com/facebook/create-react-app/pull/8177

```
export const InputField = (props: InputFieldProps): JSX.Element => {
```

src/components/career-form/career-form.tsx
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    const payload = {
+      ...apiData,
+      ...formValues,
+    };
+
+    const formData = new FormData();
+
+    for (const [key, value] of Object.entries(payload)) {
```
feedm3, line 61: I don't really get that for loop, could you add some comments?

src/components/career-form/career-form.tsx
```diff
@@ -0,0 +1,191 @@
...
+          formValues.documents[0],
+          formValues.documents[0].name,
+        );
+      }
+      formData.append(key, value as any); // formdata doesn't take objects
+    }
+    // await new Promise(resolve => setTimeout(resolve, 2000));
+    await axios
+      .post(API_ENDPOINT, formData, {
+        onUploadProgress: (progressEvent) =>
```
feedm3, line 74: nice

src/components/career-form/career-form.tsx
```diff
@@ -0,0 +1,191 @@
...
+} from './career-components';
+
+interface CareerFormProps {
+  recruiting_channel_id: string;
+  job_position_id: string;
+  access_token: string;
+  company_id: string;
+}
+
+type FormData = {
```
feedm3: Why a `type` and not an `interface`?

I usually only use `type` for primitive types like enums 
```
type Day = 'Monday' | 'Tuesday'
```

src/templates/career-details.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <div key={name}>
+              <SectionTitle>{name}</SectionTitle>
+              <PersonioHtml dangerouslySetInnerHTML={{ __html: value }} />
+            </div>
+          ))}
+        </GridItem>
+
+        <GridItem xs={12} md={8}>
+          <SectionTitle>Bewirb dich jetzt</SectionTitle>
+          <CareerForm
```
feedm3, line 103: As this is a `Grid` it should be put outside of the parent grid here, so that there is no other `Grid` around it. Too much margin otherwise.

Also, `margin-top` is missing. Should be `32px`.

image

feedm3 (approved):
Nice, it works! 🚀  Have some small suggestions for you but overall really good.

src/components/career-form/career-components.tsx
georgiee: makes so much more sense. I just picked up this (familiar) concept from another place in the repo. Thanks!

src/components/career-form/career-form.tsx
georgiee: This is to construct the form data object. I iterate of the `entries` iterator to get my hands on the key & values easily.
Inside the loop I can then differentiate between the `documents` entry (which needs special handling) and the other "atomic" values I can just append as is.

src/components/career-form/career-form.tsx
georgiee: I'm constantly torn between interfaces and types 
and I don't have a proper "best practice" yet. I have discussed with other colleagues a similar thought. It's not about primitives but about the type composition.

This boils down to: Use interface whenever possible and type in any other case `type` like for composition. I will change it here because I randomly picked one of the two ways here and I would like to align 👍

src/templates/career-details.tsx
georgiee: oh thanks, I ignored the grid in grid fact somehow entirely.

georgiee: I had to keep the grid but I extended it with a `nested` property to allow it to nest without the overhead of spacing.

PR #73, opened 2021-05-05
feat: improve font loading

### LINKED ISSUE #70: Improve font loading with the gatsby-omni-font-loader plugin
After a reload, the font on the page flickers. There is a new plugin that can probably handle this problem. Let's try it out:

https://github.com/codeAdrian/gatsby-omni-font-loader

feedm3:
This PR started with a simple fix and ended in a detailed analysis 😄 

The changes of this PR should hopefully fix the problem:
- correct caching headers to show the font immediately when reloading or revisiting the page
- using `font-display: auto` instead of `swap` to let the browser decide which font to show (for the first visit when the font is not cached yet)

## Current situation

1. We already [preload](https://web.dev/optimize-webfont-loading/#preload-your-webfont-resources) the `normal` and `bold` font with highest priority (not `italic`, as this font is only used on some special occasions). It also gets cached correctly: 
image
2. If you run `yarn build` on localhost and then `yarn serve`, you will see the correct font immediately. This is the case for both `font-display: swap` and `font-display: block`. 

What is the difference between Prod and Localhost?

It's the waiting time for the response:
On Production, the browser has to wait 170ms for the response
image
Caching header: `public, max-age=0, must-revalidate`

On localhost, it's only 12ms
image
Caching header: `Cache-Control: public, max-age=0`

## Possible solutions

**improve caching headers**
Netlify sets the `must-revalidate` header. This header means, that the browser asks the server first if the resource has changed before it is used (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#revalidation_and_reloading). As the font never changes, we can omnit this header.

Correct caching header for fonts:
`cache-control: public, max-age=31536000, immutable` ([Gatsby Docs](https://www.gatsbyjs.com/docs/caching/#static-files))

-> will do

**better fallback font**
We can improve our fallback font with https://meowni.ca/font-style-matcher/ 
This will not remove the flicker, but reduces it to a minimum.

- could do: this would improve the font swapping for first time visitors. let's see how `font-display: auto` will work and if this is still necessary

**[omni-font-loader plugin](https://github.com/codeAdrian/gatsby-omni-font-loader)**
Using a different font loading mechanism won't make any difference. The font is already loaded with highest priority.

-> won't do: doesn't solve the problem

**font-display block**
Using `font-display: block` will make the page still flicker, as the font is not displayed initially (but all other elements). It's just a different flicker as it is right now.

> This PR introduces the `font-display: block`. You can see the behaviour in the preview. Keep in mind that the preview environment disables caching in the browser

-> won't do: doesn't solve the problem

## Changing the Netlify headers

Netlify headers can be customized within the `_headers` file: https://docs.netlify.com/routing/headers/
There is also a Gatsby Netlify plugin which will add some all the headers we need out of the box: https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-netlify

**Result**
Font is in the static folder. Marking those files as `immutable` should make them instantly available in the browser.
image

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -72,5 +72,8 @@ module.exports = {
         exclude: ['/imprint', '/data-privacy', '/blog/**'],
       },
     },
+    {
+      resolve: `gatsby-plugin-netlify`,
+    },
   ],
 };
--- a/src/components/layout/fonts/fonts.css
+++ b/src/components/layout/fonts/fonts.css
@@ -7,7 +7,7 @@
     font-family: 'CocoGothic';
     font-style: normal;
     font-weight: 400;
-    font-display: block;
+    font-display: auto;
     src: local('CocoGothic'),
     url("CocoGothic.woff2") format('woff2'),
     url("CocoGothic.woff") format('woff'),
@@ -19,7 +19,7 @@
     font-family: 'CocoGothic';
     font-style: italic;
     font-weight: 400;
-    font-display: block;
+    font-display: auto;
     src: local('CocoGothic Italic'),
     url("CocoGothic-Italic.woff2") format('woff2'),
     url("CocoGothic-Italic.woff") format('woff'),
@@ -31,7 +31,7 @@
     font-family: 'CocoGothic';
     font-style: normal;
     font-weight: 700;
-    font-display: block;
+    font-display: auto;
     src: local('CocoGothic Bold'),
     url("CocoGothic-Bold.woff2") format('woff2'),
     url("CocoGothic-Bold.woff") format('woff'),
--- a/src/components/layout/fonts/fonts.css
+++ b/src/components/layout/fonts/fonts.css
@@ -7,7 +7,7 @@
     font-family: 'CocoGothic';
     font-style: normal;
     font-weight: 400;
-    font-display: swap;
+    font-display: block;
     src: local('CocoGothic'),
     url("CocoGothic.woff2") format('woff2'),
     url("CocoGothic.woff") format('woff'),
@@ -19,7 +19,7 @@
     font-family: 'CocoGothic';
     font-style: italic;
     font-weight: 400;
-    font-display: swap;
+    font-display: block;
     src: local('CocoGothic Italic'),
     url("CocoGothic-Italic.woff2") format('woff2'),
     url("CocoGothic-Italic.woff") format('woff'),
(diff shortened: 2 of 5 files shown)


ghost:
**satellytes.com**

:tada: Your build was successful! [See the Deploy preview here.](https://satellytescom-featfontloading.gtsb.io)

## Build Details

[View the build logs here.](https://gatsbyjs.com/dashboard/a7744c7a-07a1-4266-88c4-b236495f6f4b/sites/2611ae8b-872e-4dc5-9f18-b2b73f263818/builds/d90003d1-279d-40d6-b585-95eaa97319e9/details)

:clock1: Build time: **1m**

## Performance

### Lighthouse report

| Metric | Score |
| ----- | ---- |
| Performance |   :large_orange_diamond: 77   |
| SEO |   :green_heart: 90   |

:link: [View full report](https://build-d90003d1-279d-40d6-b585-95eaa97319e9.gtsb.io/reports/lighthouse/index.html)

src/components/layout/fonts/fonts.css
```diff
@@ -7,7 +7,7 @@
     font-family: 'CocoGothic';
     font-style: normal;
     font-weight: 400;
-    font-display: swap;
+    font-display: auto;
```
georgiee, line 10: 👍

gatsby-config.js
```diff
@@ -72,5 +72,8 @@ module.exports = {
         exclude: ['/imprint', '/data-privacy', '/blog/**'],
       },
     },
+    {
+      resolve: `gatsby-plugin-netlify`,
```
georgiee, line 76: wow how much code goes into that plugin?
https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-netlify/src

I'm a fan of cherry picking the small bits we need. Can't be that much in this case. Would this be possible instead of pulling in a plugin that does ton of things we don't realize (which could bite us in the future).

It's okay for this PR though, I'm just wondering if you would be willing to share my opinion about plugins like this.

georgiee (approved):
wow, it's absolutely noticeable 🥳 Let's merge this already!

Here the obligatory before/after:

bad:

https://user-images.githubusercontent.com/1701755/117126400-fd3af380-ad9a-11eb-931c-6cfa4e8d45d1.mp4

good:

https://user-images.githubusercontent.com/1701755/117126415-00ce7a80-ad9b-11eb-950e-cb0878fc70d1.mp4

gatsby-config.js
feedm3: Actually I prefer the plugin over copying code into our repo:
- As it's an official plugin, it's pretty save that it will always be up-to-date with Gatsby 
- It creates a quite sophisticated `_headers` file which covers all files and not just the fonts. I wouldn't want to implement and maintain all of the logic on my own

PR #80, opened 2021-05-17
feat: add office page

georgiee:
This PR adds the office page with our current set of photos.

## Description
I used the new gatsby image plugin. See the [docs](https://www.gatsbyjs.com/plugins/gatsby-plugin-image/) and the [migration guide](https://www.gatsbyjs.com/docs/reference/release-notes/image-migration-guide) to compare it with the legacy edition.

I was able to use the `StaticImage` component to skip the graphql part for this page 👌
I used the maximum width of `1200px` together with the constrain layout to let it shrink in width.

In addition I also installed a plugin to check markdown links and transform them into gatsby links in order to use the react router for our internal links.

The image quality is really bad with the default of 50 but even 80 doesn't look much better. I think the purposely added grain in the image post-processing doesn't fit well with the anticipated high compression. It's not super bad so let's keep it as is for now as we want to update the office images soon anyway.

I also wanted to add a little bit of text after the images. I linked to the career page as the interests of "checking the office" and having a career at Satellytes correlates a little bit. I also linked back to the office from the career page intro.

## Review
Check the screenshot and the [deployed version](https://satellytescom-gkoffice.gtsb.io/)
Screen Shot 2021-05-18 at 14 02 56-fullpage


feedm3 (approved):
Nice, looks really good 👍  There are some unused imports in `office.tsx` (Github shows them directly in the PR https://github.com/satellytes/satellytes.com/pull/80/files#diff-8fa577ceab9c8ef8974d3131740c17c033d1b76675ebb5c6cfee85651f057baa), would be cool to remove those before the merge as they will otherwise land in the webpack bundle. Other than that, let's merge.

georgiee:
Thanks for the feedback. I will check the imports. I expected the linting to catch this, but I maybe maybe I have committed with `no-verify` quickly before going to lunch 😬 I will check it, thanks.

georgiee:
I have removed the dangling code including an obsolete graphql query and added some missing spacing too. It's now merged. Thanks.

PR #81, opened 2021-05-17
feat: add customized social image

georgiee:
Let's create share images with more value than an empty tile with logo.
Here some examples what is being created:

## Preview
The preview should work with the deployed url once available. I will check with some meta checker then.
out

Automation of the day:
I used the wonderful image magick to quickly tile them into a single image.
(`montage -density 300 -tile 2x0 -geometry +5+50 -border 10 *.jpg out.png`)

## Description
I have created a [node-canvas](https://github.com/Automattic/node-canvas) based script that does three things:
1. Start with our default background
2. Add the title
3. Add the author (optional)

In order to do that I had to load custom fonts and use a text wrapping algorithm as the canvas is pretty bare bone (even with the node specific re-implementation).

Once I had the working tool I was able to integrate it in the gatsby-node configuration. The first naive approach was to run `generateCard` directly before `createPage` and pass in the image url through the context.

A much more flexible approach is to hook into `onCreateNode` in order to create the image per node (like every markdown node or whatever content we have as the base for our pages). This doesn't work entirely because our personio API access happens directly before `onCreatePage` which is not very gatsby-like. We should create all nodes from the API response first (our career position nodes) and later query those to create the pages. That way I could process everything from inside `onCreateNode`. Let's start with this version and then improve later 👌


ghost:

:triangular_flag_on_post: Your build failed. [See the build logs here](https://gatsbyjs.com/dashboard/a7744c7a-07a1-4266-88c4-b236495f6f4b/sites/2611ae8b-872e-4dc5-9f18-b2b73f263818/builds/6a59e27e-ae5c-4e6f-b36a-d5a59b41bee4/details)

georgiee:
Work on my machine but fails on gatsby cloud. `gatsby build` works locally, I can't access the build log. @feedm3 please help 🙏

feedm3:
This is the build log
<details>
<summary>Screenshot</summary>
grafik
</details>

It looks like something is broken with the  `gatsby-config.js`. Looking at the code, the only change is the 
```
const { siteMetadata } = require('./gatsby-config');
```
in the `gatsby-node.js`. Doesn't look suspicious to me, but maybe try to hardcode the `siteMeta`?

Those cards look super cool 😍

georgiee:
Hey Meta is not working, loads forever.
https://www.heymeta.com/url/satellytescommain-gkshareimage.gtsb.io/career/325493-senior-uxui-designer-(mwx)
but the career page works:
https://www.heymeta.com/url/satellytescommain-gkshareimage.gtsb.io/career

So probably a problem with the URL containing special chars like `(` and their service.

 https://metatags.io/ and twitter card validator works🥳
Screenshot 2021-05-18 at 12 19 46
Screenshot 2021-05-18 at 12 20 38

Let's merge and then improve 🙌

.eslintrc.js
```diff
@@ -37,5 +37,5 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
   },
-  ignorePatterns: ['gatsby-*.js', 'jest', '*.md'],
+  ignorePatterns: ['gatsby-*.js', 'jest', '*.md', 'tooling/**'],
```
georgiee, line 40: It's linting my js file with ts rules which is causes issues I can't fix so here is the workaround. Should this work or what's wrong with our setup ?

data/pages/career.md
```diff
@@ -1,3 +1,7 @@
+---
+title: Karriere
```
georgiee, line 2: the `onCreateNode` approach I uses requires some meta data for the markdown file. This title is currently only used fro the preview images but I think it would be good to use it for the actual headline too? It's otherwise a pretty inconsistent markdown handling.

gatsby-config.js
```diff
@@ -7,7 +7,7 @@
  */
 const extractGatsbyCloudPreviewUrl = () => {
   const PRODUCTION_BRANCH_NAME = 'main';
-  const DOMAIN_NAME = 'satellytescom';
+  const DOMAIN_NAME = 'satellytescommain';
```
georgiee, line 10: the new gatsby installation I provided has changed the name

gatsby-node.js
```diff
@@ -9,15 +11,52 @@ const CAREER_DETAILS_TEMPLATE_PATH = path.resolve(
...
+    generateCard({ title: post.title }, outputFile).then((filename) => {
+      createNodeField({
+        node,
+        name: `socialCard`,
+        value: publicUrl,
+      });
+    });
+  }
+};
+
```
georgiee, line 45: I think this approach is really beautiful. I stole it from an existing plugin:
https://github.com/andrioid/gatsby-plugin-social-cards/blob/0a18b30f50041b09721c3839c5c5d377781ecab6/gatsby-node.js#L10

We can't use the plugin as we have special requirements (two fonts for example) and a much simpler drawing solution (they use jimp we use node-canvas).

gatsby-node.js
```diff
@@ -26,20 +65,10 @@ const createCareerPages = async ({ actions }) => {
   const jobsParse = xmlParser.parse(jobsXml);
   const positions = jobsParse['workzag-jobs'].position;
 
-  const slugifyPositionName = (name) => {
```
georgiee, line 29: I installed `slugify`. Actually it should be installed already (implicite) because it's a dependency of the remark plugin. I think it makes sense to swap our custom code withe the "standard" library of creating slugs.

gatsby-node.js
```diff
@@ -71,11 +100,26 @@ const createCareerPages = async ({ actions }) => {
       ? description.value || ''
       : '';
 
+    /**
+     * Ideally we create first the nodes from the API and then create the pages
```
georgiee, line 104: yeah, I think I'm going to refactor our personio api approach in that direction soon so we can have a single point where we generate our cards.

tooling/generate-card/generate-card.js
```diff
@@ -0,0 +1,107 @@
...
+
+  context.fillText(line, x, y);
+};
+
+/**
+ * Generate a card image (jpeg format)
+ * by composing a background with a title
+ * and an optional author.
+ */
+async function generateCard({ title, author }, file) {
```
georgiee, line 67: that's the core of creating the preview images 👌

src/templates/career-details.tsx
```diff
@@ -90,8 +91,9 @@ const CareerPage = ({ pageContext }: CareerPageProps): JSX.Element => {
       <Aurora type={AuroraType.Pink} />
       <Layout siteTitleUrl="/career" transparentHeader={true}>
         <SEO
+          imageUrl={pageContext.socialCardImage}
           title={`Karriere - ${pageContext.position.name} | Satellytes`}
-          description={`Wir suchen nach ${pageContext.position.name}! Schaue Dir unsere offene Stelle an. Wir freuen uns auf Deine Bewerbung.`}
+          description={`Bewirb dich jetzt als ${pageContext.position.name}! Schaue Dir unsere anderen offenen Stelle an. Wir freuen uns auf Deine Bewerbung.`}
```
georgiee, line 96: Make the sentence more valid.

src/templates/career-details.tsx
```diff
@@ -69,6 +69,7 @@ export const PersonioHtml = styled.div`
 interface CareerPageProps {
   pageContext: {
     position: PersonioJobPosition;
+    socialCardImage: string;
```
georgiee, line 72: image comes through the context as we can't query for personio nodes right now that we could attach a field.

package.json
```diff
@@ -8,6 +8,11 @@
     "type": "git",
     "url": "git@github.com:satellytes/satellytes.com.git"
   },
+  "engines": {
+    "node": ">= 14.0.0",
```
georgiee, line 12: to express our modern local environment in code.

package.json
```diff
@@ -8,6 +8,11 @@
...
+    "npm": "please-use-yarn",
```
georgiee, line 13: we used  that in another project to prevent people using npm to install things. But it's actually not always recognized so it's more like an information to read by humans at least.

georgiee (commented):
here some comments and I checked all files so it's ready for a review.

.eslintrc.js
feedm3: I never dived into this problem, don't know yet.

data/pages/career.md
feedm3: Definitely!

gatsby-node.js
feedm3: Nice usage of gatsby nodes, really good 👍

tooling/generate-card/generate-card.js
feedm3: Straight forward code, simple to read

feedm3 (approved):

PR #82, opened 2021-05-18
feat: add contact leaflet map

georgiee:
This will add a map as the header on the contact page✨

+ Use leaflet with tile layers from mapbox.
+ Custom Mapbox Style. Use the mapbox account stored in our company password vault to access and modify
+ Integrated two tile layers with a control. One with our dark style but I also want to offer a normal style if people want to explore the environment. A detail I would love to see on a page so I added it to ours
+ I'm not sure but I think the usage of Leaflet + the tiles from mapbox should be free forever. Their pricing is about usign the framework JS from mapbox and not the open source leaflet ? (https://www.mapbox.com/pricing/). Anyway should be fine for our traffic in any case
+ We have not touched any google product with this addition 💪 Feels good to protect our users from them.
+ I have included a little detail if people get lost (aka moved too far away from the office) to rescue them. We could use a small and nice animation/gif here. I love details.
+ Added the custom marker from the design but I also added a text and a circle too highlight the area. It looked nice while tinkering around with leaflet so maybe that's something worth to keep. Depends on the designers opinion partially though.
+ Scrollwheel is disabled to prevent annoying scroll complications. People have the control bottom left to zoom.

Note:
I updated the hero vs layout/header structure. Before the Hero wrapped the entire header which made it difficult to understand the DOM. I'm not sure why this was done like that because they both live perfectly side by side in the DOM without any visual difference. This change was required to let me pass a custom hero element (here: the map).

Preview will be available soon for this PR. Here two screenshots.

<details>

Screen Shot 2021-05-19 at 18 14 27-fullpage

Screen Shot 2021-05-19 at 18 14 18-fullpage

</details>


src/pages/contact.tsx
```diff
@@ -21,11 +22,13 @@ const ContactPage: React.FC = () => {
   return (
     <>
       <Aurora type={AuroraType.Pink} />
-      <Layout transparentHeader={true}>
+
+      <Layout transparentHeader={true} hero={<Leaflet />}>
```
georgiee, line 23: ne option `hero` allows us to pass in any component to be placed between the fixed header and the content.

src/components/leaflet/leaflet.tsx
```diff
@@ -0,0 +1,138 @@
...
+  background-color: ${theme.palette.text.contrastLink.default};
+  border: none;
+  font-size: 1.5em;
+  padding: 0.3em 0.5em;
+`;
+
+/**
+ * Create a Leaflet map to display the o
+ */
+export const Leaflet = () => {
```
georgiee, line 59: I love how easy it is to work with Leaflet🤩

src/components/layout/layout.tsx
```diff
@@ -123,9 +131,11 @@ const Layout: React.FC<LayoutProps> = (props) => {
         siteTitle={data.site.siteMetadata.title}
         siteTitleUrl={props.siteTitleUrl}
         $lightTheme={isLight}
-        heroImage={props.heroImage}
```
georgiee, line 126: I moved the hero from inside he header into the layout. Makes it much easier to understand and to extend.

src/components/header/header.tsx
```diff
@@ -128,30 +126,28 @@ const Header: React.FC<HeaderProps> = (props) => {
   });
 
   return (
-    <HeroImage image={props.heroImage}>
```
georgiee, line 131: Why was it structured like that? Hero nesting the header feels up side down.

gatsby-ssr.js
```diff
@@ -22,4 +22,13 @@ export const onRenderBody = ({ setPostBodyComponents }) => {
       }}
     />,
   ]);
+
+  const link = React.createElement('link', {
```
georgiee, line 26: required to be added manually. I found this way to be used by gatsby folks. I wanted to add it to gatsby-browser first but her, it's ok ?

georgiee (commented):
That was fun, here some comments.

src/components/leaflet/sy-marker.tsx
```diff
@@ -0,0 +1,26 @@
+import SyMarkerImage from './sy-marker.png';
+import Leaflet from 'leaflet';
+
+const isBrowser = typeof window !== 'undefined';
+
+/**
+ * Leaflet doesn't support SSR, so we run into issues
```
georgiee, line 7: this took me some time. When the build first failed I chose to "quick fix" this with some if-else logic and push it without testing it locally until it failed on gatsby cloud. I tried this approach 4 or 5 times and then decided to do it properly. Being lazy is never good and this solution works.

The problem here is that we access the Leaflet Icon directly. React Leaflet is wrapping leaflet properly (it seems) but access Leaflet directly creates issues that we have to solve by our own.

src/components/leaflet/sy-marker.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * with window if leaflet is loaded directly. Check file gatsby-node
+ * for the webpack config injection with replaces it with the null loaders.
+ * This also means Leaflet will be null here in this case. Handle accordingly.
+ *
+ * Related:
+ * https://github.com/PaulLeCam/react-leaflet/issues/495
+ */
+let SatellytesMarkerIcon: any = undefined;
+
+if (isBrowser) {
```
feedm3, line 17: When excluding components from the SSR, the `isBrowser` check should be done by using `useEffect`. 

```
const [isBrowser, setIsBrowser] = useState(false);
useEffect(() => {
	setIsBrowser(true)
})
```

The reason for this is, that during hydration the result must be the same on the first render. If you put the browser check outside of a react, this is not the case anymore and you get hydration errors in the console.

I checked the code on my machine, as Github doesn't show all changes:
- for the `SatellytesMarkerIcon` I guess the `window !== 'undefined'` check is ok, as the `Leaflet` component also has a browser check
- the leaflet component should use the `useEffect` check, for optimal hydration

feedm3 (approved):

feedm3:
General question: How does Leaflet and Mapbox work together? Doesn't Mapbox have their own SDK?

georgiee:
First of all thanks for your review. I will check the SSR more carefully, I have not much experience with React/SSR so I just made it work and I hoped for a feedback to clarify things 🙏

> General question: How does Leaflet and Mapbox work together? Doesn't Mapbox have their own SDK?

Mapbox is a tile provider and they also ship a custom advanced map render engine called [Mapbox GL JS](https://docs.mapbox.com/mapbox-gl-js/). They also maintain [Mapbox.js](https://docs.mapbox.com/mapbox.js/api/v3.3.1/) which builds on top of leaflet.

You can easily stick to Leaflet because you can just consume the map tiles. 
You can spot the following url.

```
https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/{z}/{x}/{y}?access_token=
```

which gets passed to leaflet. There are dozens of providers out there. You can also use google maps:

```
http://{s}.google.com/vt/lyrs=m&x={x}&y={y}&z={z}'
```

Pretty nice that this is so flexible. That also allows you to "easily" integrate custom tiles. Like a cat world for example 😅
https://leafletjs.com/examples/extending/extending-2-layers.html

src/components/leaflet/sy-marker.tsx
georgiee: thank you for the useEffect hint. What's the best way to test the outcome? Do I always have to build or is it enough to disable javascript? Or as you commonly demonstrate, check the network pane for the html payload ? 

I will work on the Leaflet component to make SSR better here.

gatsby-ssr.js
feedm3: yep 👍

src/components/header/header.tsx
feedm3: Don't ask :D

src/components/layout/layout.tsx
```diff
@@ -99,7 +101,7 @@ function overrideDarkFromQuery() {
   return params.has('dark');
 }
 
-const Layout: React.FC<LayoutProps> = (props) => {
+const Layout: FunctionComponent<LayoutProps> = (props) => {
```
feedm3: Recommendation by react team: Use `JSX.Element`.

```
const Layout = (props: PropsType): JSX.Element => (...)
```

src/components/leaflet/leaflet.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const [mapInstance, setMapInstance] = useState(null);
+  const [isBrowser, setIsBrowser] = useState(false);
+  useEffect(() => {
+    setIsBrowser(true);
+  });
+
+  // we don't want to render leaflet outside of the browser (SSR)
+  if (!isBrowser) {
+    return null;
```
feedm3: Ask already said in slack: Empty `div` with the same `height` as the `MapContainerWithHeight`

src/components/leaflet/sy-marker.tsx
feedm3: Most of the time, I just check the outcome in the network tab. Disabling JS is a good idea, never tried it.

feedm3 (approved):

src/components/layout/layout.tsx
georgiee: I think this is rule number one with react & TS: You will always pick the wrong type if you don't look it up once to learn the correct one 😬Thanks.

georgiee:
I pushed some changes:
+ Enable SSR all the time via `DEV_SSR`
+ Enable placeholder in office (piggy back change, because I discovered this while debugging the preview for the map)
+ I chose not to use any gatsby image component for the placeholder. A plain img is just fine and I load a static preview of the selected map.
+ I use a different tile size (512 instead of 256) to make the labels bigger. By default it's only `256`. To increase it we have to offset the zoom level. We kind of pretend to be in zoom level + 1 (which matches the next bigger tile size). that way we get tiles with bigger labels.

I also double checked the details of the free tier. The tiles itself are of course also limited. It's visible within the mapbox account. Should be fine for us.
Screenshot 2021-05-19 at 11 28 20


Ready to be merged if nothing else blocks.

src/components/layout/layout.tsx
georgiee: Can you link to the recommendation?
`React.FC` and `React.FunctionComponent` are the same types and they refer to functional components and include the children property. When I switch to the `JSX.Element` type I have to declare the children manually which is awkward. It's working so we can keep it and try it out in the next PR.

feedm3: Declarin children manually is intentional. Not all components have children.

- There is a more detailed blog post about that here: https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/
- And the React PR that even removes the `React.FC` from types: https://github.com/facebook/create-react-app/pull/8177 (this is from 2020 and it's still in the code, so I don't know where exactly they removed it... 😄 )

feedm3:
Small details:
- The image on mobile is to large
- The image looks a bit different then the rendered Leaflet component. Obviously, because it's just on static image and not an image for every size. I'm not sure, but maybe using a transparent placeholder looks and feels better?

src/components/leaflet/leaflet.tsx
georgiee: after going back and forth it's now an empty div with the correct media queries to expose the correct placeholder height for any screen.

src/components/layout/layout.tsx
georgiee: Thanks for the links, I'm finally convinced and I will be explicit about everything including children in the future 🤓🙌

src/components/layout/layout.tsx
```diff
@@ -99,7 +102,7 @@ function overrideDarkFromQuery() {
   return params.has('dark');
 }
 
-const Layout: React.FC<LayoutProps> = (props) => {
+const Layout = (props: LayoutProps): JSX.Element => {
```
georgiee, line 105: there we are 🙏

PR #83, opened 2021-05-19
feat: working satellite (draft)

georgiee:
This is a draft

---

First a preview video as the build will fail due to SSR errors with threejs/threejs fiber.

https://user-images.githubusercontent.com/1701755/118773621-4c961f00-b885-11eb-880c-c6f231d283ad.mp4

This was surprisingly easy with threejs fiber that wraps threejs to use it in react and act on it like threejs is a dom.
Unfortunately this causes quite some issues in gatsby. HMR interferes with it you get often a suspended error. I understand fibers and the suspended concept only high level. Mixed with all those hooks I felt like a novice putting strange things together until it worked. Well it worked at least.
 
Anyway I stayed pragmatic and added some post processing and frame by frame animation. Looks good already.
I stopped myself though to focus on priorities first.

Next steps:
+ Make canvas full stage so we can place the satellite behind the text and don't have content that is cut by the limited bounding box.
+ Fix the build errors. I think we might have to go back to vanilla threejs. That was my original plan but it's a lot of more manual work to make this work. Before doing this we need to make sure that gatsby will not fail with this.
+ Workaround Opportunity. Like with the leaflet map #82  we could introduce a null loader to remove the entire satellite component on SSR. But I'm not sure if this will work on the special 404 page then (is it always dynamic or statically rendered?)


georgiee:
I don't think we will make the changes anytiem soon to bring the satellite into the 404 page. The PR can stay as a reference but no need to keep it open, not even as a draft.

PR #84, opened 2021-06-08
feat: new blog post about the injector

georgiee:
During the day I took constant small breaks to write this small blog article.
I would like to see how it feels and what we are missing right now.

Our codebase is a little mess and I try to refactor code from the beginning bit by bit with every blog post. Naturally this firs blog post caused some more changes I wanted to have fixed.

+ The blog image was a cloudinary url directly in frontmatter. I would like to use a local image. This must be possible
+ I wanted to highlight lines in our code blocks. Did not find an easy way although the [documentation ](https://www.gatsbyjs.com/plugins/gatsby-remark-prismjs/) tells it should work, but without an example.
+ I tried out to use the `diff-` highlighter. Works okay'ish, maybe the theme is not ideal supporting? Because colors are weirdly mixed.
+ A header image is missing. Or any image at all 😅 That makes it a really rough reading experience and I would like to add something soon.
+ ~There might be some typos, I still have to read it a second time.~
+ Updated the hero image handling. I think the one integrated in the `Layout` is now obsolete but I didn't dare to remove it just yet as I don't know about the implications for content outside of the blog.
+ Unlocked the blog in the navigation
+ Added reading time


georgiee:
@feedm3
FYI, I had to reactivate the gatsby runner manually because it was suspended due to inactivity. Fair enough for the free tier though.

Screenshot 2021-06-08 at 15 32 37

Screenshot 2021-06-08 at 15 32 57

georgiee:
The preview looks quite good (via https://metatags.io/) but the excerpt is missing. I will add the change.
Screenshot 2021-06-10 at 10 06 28

src/components/util/use-media-query.ts
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
+
+/**
+ *
+ * via https://www.netlify.com/blog/2020/12/05/building-a-custom-react-media-query-hook-for-more-responsive-apps/
+ */
+export function useMediaQuery(query) {
```
feedm3: Checking the media query with Javascript has one big downside: The initial render will always be either mobile or desktop. If it's desktop and the user opens the page on mobile, it will flicker.

In other projects we are actively removing the media query Javascript hook and replacing it with pure CSS to improve the UX. A standard way to do this, is to add both the desktop and the mobile HTML to the DOM tree, and hide either one or the other with a media query.

It makes the DOM tree lager but in most cases this downside is worth the upside of having a perfect page on the initial render.

feedm3 (approved):

src/components/util/use-media-query.ts
georgiee: Thanks for the hint. Makes sense and I actually prefer the CSS only solution. I got dragged into that direction by forcing myself to think "more react-like" whatever this means but it caused me to use JavaScript rather than CSS. I will remove it already again 👌

georgiee:
That looks great now too ☀️

Screenshot 2021-06-10 at 10 32 40

feedm3:
It really cool, I like it. Looking forward to more blog posts 💪 

What I don't like is the code formatting. It's just my taste but the Github code formatting looks way better. I created a small example gist. Maybe we can use this formatting as inspiration. Doesn't matter for merging this PR though, more of a future improvement.

https://gist.github.com/feedm3/c366ac7f00e3cbd5de899e78bae404e4

> edit: github includes the gist right into the preview, nice

georgiee:
I fully agree with the syntax highlighting. I added the default syntax from prism. Much better than the "coy" version I used for the light design (see https://prismjs.com/) but the `diff` highlight is still off. Let's go with this and have a dedicated session for the syntaxt highlight then. There must be a github flavoured prism style out there already.


Screenshot 2021-06-10 at 10 50 20

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+title: "The \"Inject the Injector\" pattern"
+featuredImage: images/georgios-kaleadis-aBTfTMsOCOI-unsplash.jpg 
+imageCredits: We want to credit people for their work
+shortSummary: A coding pattern to prevent breaking changes when dealing with injections in base classes used in distributed libraries
+author: "Georgios Kaleadis"
+authorSummary: "CTO at Satellytes"
+---
+ 
+
+We maintain a successful proprietary enterprise library based on Angular. One challenge while doing so are breaking changes. Those occur naturally while we improve and extend the library. There was one particular type of breaking change that caused us some trouble, and we want to show you how we tackled it for good.
```
SyFlo:
```suggestion
We maintain a successful proprietary enterprise library based on Angular. One challenge while doing so is how to deal with breaking changes. Those occur naturally while we improve and extend the library. There was one particular type of breaking change that caused us some trouble, and we want to show you how we tackled it for good.
```

just thinking out loud, maybe the adjective "successful" sounds a bit like patting ourselves on the shoulder in the first sentence of an introductory paragraph. maybe something along the lines of "in widespread use within a large enterprise organization" could be considered as an alternative. definitely not a big issue though.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+    @Inject(Injector) injector: Injector,
+    @Inject(MyService) myService: MyService
+  ) {
+    super(injector)
+  }
+}
+```
+
+## Our starting point
+In order to understand the problem we will look at the following Angular base class that act as the functional foudnation for derived explicit components.
```
SyFlo:
```suggestion
In order to understand the problem we will look at the following Angular base class that acts as the functional foundation for derived concrete components.
```
two little typos and I changed "explicit" to "concrete" because that might be a slightly better choice of words. However, I think "derived components" without the additional adjective would be clear enough. imo the additional adjective doesn't add anything in terms of meaning.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  ngOnInit(){
+    this.myTrackingService.trigger();
+  }
+  
+  subscribe() {
+    this.mySubscribeService.doSomething();
+  }
+}
+```
+That base class provides a set of default functionality for any other component extending from it in the future. This not only saves repeated work on the side of the component authors. This acts as an alignment & contract between all derived components.
```
SyFlo:
```suggestion
That base class provides default functionality for any other component extending from it in the future. This not only saves repeated work on the side of the component authors, but also acts as an alignment & contract between all derived components. 
```
Just a suggestion that is a little easier on the eye (at least at mine).

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  
+  subscribe() {
+    this.mySubscribeService.doSomething();
+  }
+}
+```
+That base class provides a set of default functionality for any other component extending from it in the future. This not only saves repeated work on the side of the component authors. This acts as an alignment & contract between all derived components. 
+
+The `MyAbstractBaseComponent` will be delivered through a core library and extended by dozens of other components.
+Like the following imaginary subscription component:
```
SyFlo: I'd suggest to merge this into one sentence ."..by dozens of other components like the following imaginary subscription component:"
It's not that I don't **like** the imaginary component that follows ;-) just that the word "like" should rather be interpreted as "for example" in this case.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  selector: 'my-subscription',
+  template: `<button (click)="subscribe()">subscribe now</button>`
+})
+export class SubscriptionComponent extends MyAbstractBaseComponent {
+}
+```
+
+The `SubscriptionComponent` is guaranteed to invoke the one service through `ngOnInit` as described by the parent class plus it can access the `subscribe` method from the parent class as the service `mySubscribeService` is readily available to be invoked by `subscribe()`.
+
+Things get complicated the moment library authors starts to get a little more advanced.
```
SyFlo:
```suggestion
Things get complicated the moment library authors start to get a little more advanced. 
```
Just a little typo. However, the first sentence tripped me up a bit. Nothing to put my finger on, but the choice of words "authors getting advanced" seems a bit odd and I'm not sure everyone will immediately understand that you're talking about the programming experience of authors. Just a suggestion, maybe something along the lines of "Things get complicated the moment library authors implement slightly more advanced use cases"?

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  
+  unsubscribe() {
+    this.myUnsubscribeService.trigger();
+  }
+}
+```
+
+The class `UnsubscribeToolComponent` is in desperate need to access a custom service `MyUnsubscribeService` whose functionality is not provided by the base class. Fair enough, they chose to inject that service. The injection itself looks complicated though, because the author needs to repeat the injection to forward the two service instances `mySubscribeService` & `myTrackingService` to the parent class. 
+
+Otherwise, the solution works pretty well. Great.
```
SyFlo:
```suggestion
Besides, the solution works pretty well. Great.
```
"Otherwise" could maybe be interpreted as "it would, but it doesn't".

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+
+## The problem
+
+The core team decides to extend the functionality of the base class. They want to ship a new service `MyRemoteHomeService` to all subclasses and offer a new method `lightsOff` ready to be used. This is adding some new feature and shouldn't cause much trouble, should it ?
```
SyFlo:
```suggestion
The core team decides to extend the functionality of the base class. They want to ship a new service `MyRemoteHomeService` to all subclasses and offer a new method `lightsOff` ready to be used. This is adding some new feature and shouldn't cause much trouble, should it?
```

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
++  }
+}
+```
+You can clearly see the appearance of `MyRemoteHomeService` and the method `lightsOff()` using it.
+
+The team decides not to mark it as a `BREAKING CHANGE`, because, well it can't break anything because it **adds** functionality. They will ship it as version `V1.1`. Soon after the release the maintainers of the advanced component `UnsubscribeToolComponent` complain that their component fails to compile.
+
+The reason gets obvious pretty quickly. The local team didn't know about the service added by the core team. On the other side they rely on repeating the constructor signature in order to provide all expected dependencies.
+
+The team's fix is simple, but it came by surprise, and it created a lot of confusion.
```
SyFlo:
```suggestion
The team's fix is simple, but it came by surprise and it created a lot of confusion.
```
Not sure if there's a grammar rule for that, but imo it reads more fluently without a comma preceding the conjunction "and".

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  }
+  
+  unsubscribe() {
+    this.myUnsubscribeService.trigger();
+  }
+}
+```
+What did they do? They had to repeat the injection of `MyUnsubscribeService` to forward the instance through the `super()` invocation. 
+
+The core team created a breaking change by adding an innocent new feature because they did not recognize that slightly more advanced users of their library are overriding the constructor. The team could stop here, because they are now aware of the problem, and the next change of the constructor will be marked as a breaking change.
```
SyFlo:
```suggestion
The core team created a breaking change by adding an innocent new feature because they did not recognize that slightly more advanced users of their library are overriding the constructor. The team could stop here, because they are now aware of the problem and the next change of the constructor will be marked as a breaking change.
```
same suggestion as in the comment above

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+    this.mySubscribeService = injector.get<AclService>(MySubscribeService)!;
+    this.myTrackingService = injector.get<AclService>(MyTrackingService)!;
+    this.myRemoteHomeService =  injector.get<AclService>(MyRemoteHomeService)!;
+  }
+}
+```
+
+Can you see the elegance here? We inject the injector, which is the engine of the dependency injection (DI) system in Angular itself and then request the singleton instances of our desired services to assign them to the local variables as before.
+
+The injector has the correct typing, that the result might be undefined, that's why we convince TypeScript with the [non-null assertion operator (!)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) that we guarantee to receive a value. That's possible because we control the environment, and the services are guaranteed to be available as they are provided in the root (`@Injectable({providedIn: 'root' })`).
```
SyFlo: The first sentence reads a bit awkward. Maybe if we turned it around, it would be easier to understand. What do you think about the following suggestion?

We need to use the non-null assertion operator (!) to tell TypeScript that we guarantee to receive a value, because the typing of the injector correctly states that the result might be undefined.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+
+In case we decide to add a fourth of fifth service in the base class, we can now add it and request it directly from the injector without breaking any subclass as the constructor signature stays the same. See how many lines of random services the team are not even interested in, can now be replaced:
```
SyFlo:
```suggestion
In case we decide to add a fourth or fifth service in the base class, we can now add it and request it directly from the injector without breaking any subclass as the constructor signature stays the same. See how many lines of random services, in which the feature teams are not even interested in, can now be replaced:
```
one little typo and a suggestion to fix the use of "are" in conjunction with the singular "team" in the last part of the sentence.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
++    super(injector)
+  }
+  
+  unsubscribe() {
+    this.myUnsubscribeService.trigger();
+  }
+}
+```
+
+resulting in this much more compact version of a constructor.
```
SyFlo: This is not a complete sentence. Not a problem per se, but maybe not in perfect harmony with style of writing in this article. How about "The result is a much more compact version of the original constructor."?

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+  ) {
+    super(injector)
+  }
+  
+  /*...*/
+}
+```
+
+## Conclusion
+Handling breaking changes is an act of empathy 💛. You want to protect your users from struggling with your changes. The "inject the injector" pattern we have introduced here helped us a lot and made it very easy to extend our base class without breaking things anymore.
```
SyFlo:
```suggestion
Handling breaking changes is an act of empathy 💛. You want to protect your users from struggling with your changes. The "inject the injector" pattern we have introduced here helped us a lot and made it very easy to extend our base class without breaking things.
```
I think I might just omit the last word "anymore".

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
...
+
+It's always advisable to be more specific instead of using generalized concepts like described in this blog post. Your code will be less abstract, better readable and maintainable. Use this approach only if you have very good reasons to do so 👍 This is a highly specific pattern for our distributed library project. In case you're developing individual components that are not distributed to other developers you most probably should not care for breaking changes and patterns like that.
```
SyFlo: "It's always advisable to be more specific instead of using generalized concepts like described in this blog post." - I'm not sure I understand what you mean to say by that. Is it synonymous to 
"It's always advisable to be as specific as possible rather than using generalized concepts like the one we describe in this blog post"?
The last sentence in this paragraph suggests that this is supposed to be disclaimer of sorts. I think I get what you're saying, but maybe something along the lines of
"In case you're developing individual components that are not distributed to other developers you most probably have no need to prevent the occurrence of breaking changes by means of patterns like that"
would be a better choice of words? What developer in their right mind would not care for all sorts of patterns and breaking changes? I think what we want to say is that unless you find yourself working with a large codebase and distributed teams in a typical enterprise organization, dealing with breaking changes is much easier and introducing a pattern like this might not be necessary or even beneficial.

src/components/cards/blog-card.tsx
```diff
@@ -115,14 +103,7 @@ export const BlogCard: React.FC<BlogCardProps> = ({
...
     >
-      {image && placeholderImage && (
-        <BlogCardImage
-          fluid={{
-            ...placeholderImage,
-            aspectRatio: large ? 1.714 : 1.56,
-          }}
-        />
-      )}
+      {image && <BlogCardImage alt={''} image={image} />}
```
SyFlo:
```suggestion
      {image && <BlogCardImage alt="" image={image} />}
```
just a matter of code style preference though, so I make the suggestion just this once.

src/templates/blog-post.tsx
```diff
@@ -58,25 +68,72 @@ const BlogHeader = ({ frontmatter }) => {
...
+const HeroImageSquared = styled(GatsbyImage)`
+  display: none;
+  @media (max-aspect-ratio: 0.75) {
+    display: block;
+  }
+`;
+
+const BlogHeroImage = ({ wideImage, squareImage }) => {
+  /**
+   * A screen is tall once the height is larger then the width, which means the aspect ratio dips below 1 (0-1)
```
SyFlo: The comment says "A screen is [considered as] tall..." but I don't see any mentioning of tallness in the code below. As a casual reader this puzzles me a bit, maybe the comment could rather explain that the output of either one of HeroImageDefault or HeroImageSquared is visually hidden depending on the viewport aspect ratio. This gets apparent by looking at the styled components directly above, but giving a reference to "aspect-ratio" rather than "tallness" might save the reader a second or two.

src/templates/blog-post.tsx
```diff
@@ -58,25 +68,72 @@ const BlogHeader = ({ frontmatter }) => {
...
   return (
-    <Layout
-      heroImage={data.markdownRemark.frontmatter.image}
-      siteTitleUrl={'/blog'}
-      light
-    >
+    <Layout transparentHeader siteTitleUrl={'/blog'} light hero={heroImage}>
+      {/*
+       * SEO Notes:
+       * Recommended meta description length this days is 120 - 158 characters. The lower number is relevant for mobile devices.
```
SyFlo:
```suggestion
       * Recommended meta description length these days is 120 - 158 characters. The lower number is relevant for mobile devices.
```

SyFlo (approved):
a very pleasant and interesting read 😃 I made some suggestions, mostly about wording and correcting the odd typo.

blog-posts/angular-inject-the-injector.md
```diff
@@ -0,0 +1,252 @@
+---
+path: "/blog/angular-inject-the-injector"
+date: "2021-06-08"
+title: "The \"Inject the Injector\" pattern"
+featuredImage: images/georgios-kaleadis-aBTfTMsOCOI-unsplash.jpg 
+imageCredits: We want to credit people for their work
```
SyFlo: should this actually list people to credit for images?

georgiee: yes, then I switched plans and just took my own image but actually I would like to ship this feature with this PR so I try to push a change along the text update. Thank for the hint!

blog-posts/angular-inject-the-injector.md
georgiee: now reading this I recognize that `explicit` might be more related to content suited for adults 😁 but it's perfectly harmless in the dictionary. Thanks for the catch. Changes accepted

blog-posts/angular-inject-the-injector.md
georgiee: thank you, I like the change and accepted it.

blog-posts/angular-inject-the-injector.md
georgiee: I love short sentences but sometimes they get too short. I like your change and accept it.

blog-posts/angular-inject-the-injector.md
georgiee: I **like** it and it's prepare on my local changes 😄

blog-posts/angular-inject-the-injector.md
georgiee: I keep the successful for now, let's see how it fells in the end.

blog-posts/angular-inject-the-injector.md
georgiee: Oh you are right. Accepted

blog-posts/angular-inject-the-injector.md
georgiee: In the local changes batched like this. Thanks.
> Things get complicated the moment library authors implement slightly more advanced use cases, like in the following component.

blog-posts/angular-inject-the-injector.md
georgiee: I ALWAYS do this. I don't know when I learned that wrong pattern of inserting a space before the question mark in a sentence. Thanks!

blog-posts/angular-inject-the-injector.md
georgiee: Commas are something I don't understand in English. It's almost random for me how people use it. If there a commas I might have blindly accepted suggestions from my local IDE and if you have an opinion here I prefer the human before the machine 🤓

blog-posts/angular-inject-the-injector.md
georgiee: yes indeed. I think the sentence is a left over how I developed the sentence in my mind where I'm coming from the problem and then justify it by the operator. Sounds much better the other way around.

blog-posts/angular-inject-the-injector.md
georgiee: thank you

blog-posts/angular-inject-the-injector.md
georgiee: I added the following changes locally. Thank you

> The result is a much more compact version of the original constructor.

blog-posts/angular-inject-the-injector.md
georgiee: You understood me correctly. Actually I wrote this as a word of warning at the beginning of the blog post thinking of the people taking such a pattern and over complicating their simple project. I wanted to clarify that this is only suitable for distributed libraries. I like your last summary and here my updated sentence:


> Introducing a pattern like this might not be necessary or even beneficial if you are not working with a large codebase and distributed teams in a typical enterprise organization. Never use this pattern only to save a few lines of code, you will make your code less understandable.

src/components/cards/blog-card.tsx
georgiee: My IDE is doing that as auto completion when starting an attribute all the time and I'm being too lazy handling it correctly. It's as if my IDE would do this in Angular `[alt]="''"` which I would never accept. Thanks for the catch and motivation to handle this more carefully in the future 👍

georgiee (commented):
comments for the pending review, but I think they are too late because I just accepted the batch of changes I commented on 😄

georgiee:
I added attribution data to the markdown so we can use images with the required attribution, which is just fair to the creator 👍

```
attribution:
  creator: upklyak
  source: https://www.freepik.com/free-vector/space-landscape-night_13748451.htm
```

Screenshot 2021-06-15 at 10 39 15
Screenshot 2021-06-15 at 10 39 19

Links are nofollow and blank too. Let's finally merge this blog post to write the next one 👍

georgiee:
Thanks for your reviews, very much appreciated 💛

PR #86, opened 2021-06-15
feat: put blog posts into sitemap and update dependencies

feedm3:
- All dependencies are now up-to-date, except for `react-hook-form`. Will fix this in a separate PR
- I added the blog posts to the `sitemap.xml` to make them visible for Google


gatsby-config.js
```diff
@@ -132,7 +132,7 @@ module.exports = {
     {
       resolve: `gatsby-plugin-sitemap`,
       options: {
-        exclude: ['/imprint', '/data-privacy', '/blog/**'],
+        excludes: ['/imprint', '/data-privacy'],
```
georgiee, line 135: oh did that never work as we used the wrong keyword?

georgiee (changes requested):
Thanks for thinking about this 🙌
1. The local http://localhost:8000/sitemap.xml doesn't show the blog posts? Wrong place to check?
2. The tests are failing, probably because it's outdated only?

gatsby-config.js
feedm3: - the sitemap plugin api changed due to the dependency update
- the sitemap only gets created on build, not with the dev server. you can see the sitemap here https://satellytescommain-choredependencies.gtsb.io/sitemap.xml
- will have a look for the tests

feedm3: test is now fixed, pipeline is green. thanks for the review :)

georgiee: I can't see the blog posts in this file though? (same as on localhost by the way).
I thought it's about the individual posts?
Screenshot 2021-06-15 at 16 27 09

feedm3: Oh, nice catch, didn't notice it.

The new sitemap plugin has a new path for the sitemap, it's `sitemap/sitemap-index.xml`. I updated the URL to our `robots.txt`. Maybe you can also do a double check in the Google Search Console and update the path to the sitemap there. Full path should be `https://satellytes.com/sitemap/sitemap-index.xml` after the merge.

- https://satellytescommain-choredependencies.gtsb.io/sitemap/sitemap-index.xml
- https://satellytescommain-choredependencies.gtsb.io/sitemap/sitemap-0.xml

The root sitemap isn't generated anymore (missing on my local environment). It's probably on the gatsby deployment due to file caching.

georgiee: clarified and answered 👍

georgiee (approved):
Let's merge that beauty of housekeeping

PR #89, opened 2021-06-18
Feat: updated career details

WiebkeFr:
pr contains:

- updated react-hook-from
- drag-and-drop field for files
- refactor: text-field in new file

todo: checkbox and clear-button and checkmark styling, loading file

test on: [https://satellytescommain-featupdatedcareerdetails.gtsb.io/career/338504-frontend-engineer-(wmx)](https://satellytescommain-featupdatedcareerdetails.gtsb.io/career/338504-frontend-engineer-(wmx))


src/components/career-form/career-file-input.tsx
```diff
@@ -0,0 +1,105 @@
...
+  error,
+  children,
+}: FileInputProps) => {
+  const onDrop = useCallback(
+    (droppedFiles) => {
+      // dropped files are appended to already selected files
+      const dataTransfer = new DataTransfer();
+      if (selectedFiles) {
+        for (let i = 0; i < selectedFiles.length; i++)
+          dataTransfer.items.add(selectedFiles[i]);
```
feedm3: Bei `for` und `if` immer `{}` danach verwenden, ist einfacher zu lesen

src/components/career-form/career-file-input.tsx
```diff
@@ -0,0 +1,105 @@
...
+    [selectedFiles, setValue],
+  );
+
+  const { getRootProps, getInputProps } = useDropzone({
+    onDrop,
+    accept: '.pdf',
+  });
+
+  return (
+    <InputContainer>
```
feedm3: Wir brauchen noch ein hover Style, wenn man die Datei in den Browser zieht. Damit der User weiß, wo der drop gemacht werden muss. So wie hier: https://react-dropzone.js.org/#!/Styling%20Dropzone

src/components/career-form/career-form.tsx
```diff
@@ -38,12 +44,24 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
   } = useForm();
   const [uploadProgress, setUploadProgress] = useState(0);
+  const selectedFiles = watch('documents');
+  const privacyChecked = watch('privacy');
 
   const onSubmit = async (formValues: FormData): Promise<void> => {
+    if (!privacyChecked) {
+      setError(
+        'privacy',
+        { type: 'manual', message: 'Hier fehlt noch ein Häckchen' },
```
feedm3: `Deine Zustimmung fehlt`

feedm3 (changes requested):

src/components/career-form/career-file-input.tsx
feedm3: Am besten kurz mit Verena checken, ob die dazu noch ein Design hinzufügen kann

WiebkeFr: Ich hab ein isDragActive bei dem Feld eingefügt. Das Styling ist bisher nur ein Beispiel

package.json
```diff
@@ -67,7 +67,7 @@
     "react-dom": "17.0.2",
     "react-dropzone": "^11.3.2",
     "react-helmet": "6.1.0",
-    "react-hook-form": "^6.15.8",
+    "react-hook-form": "^7.8.7",
```
georgiee, line 71: Was this required or just a housekeeping step ? Anything that broke or is of interest for us ? Did this cause the update in the contact form (breaking changes?)

src/components/career-form/career-form.tsx
```diff
@@ -112,102 +143,72 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
-          <GridItem xs={12} md={6}>
-            <InputField
-              required={true}
-              inputRef={register({ required: 'Dein Vorname fehlt' })}
-              error={errors.first_name}
-              name="first_name"
-              label="Vorname"
-            />
-          </GridItem>
+          <CareerTextFields register={register} errors={errors} />
```
georgiee, line 146: 👌

src/components/career-form/career-form.tsx
```diff
@@ -112,102 +143,72 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
+                  Drop files to upload or <span>browse</span>
+                </div>
+              </>
+            </FileInput>
+            <FormError error={errors.documents} />
           </GridItem>
 
-          {/*CV File */}
+          {/*File-Review*/}
+          {selectedFiles &&
```
georgiee: For me file review + file input could be a single component "Document Upload" because I will never use the parts in isolation so I'm happy to encapsulate them and hide the logic from the form component (similar to the text fields component you created within this PR).

src/components/career-form/career-textfields.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import { GridItem } from '../grid/grid';
+import { InputField } from './career-components';
+import { SIMPLE_EMAIL_PATTERN } from './career-form';
+
+export const CareerTextFields = ({ register, errors }) => {
```
georgiee: 🙌

src/components/form/contact.tsx
```diff
@@ -70,7 +74,7 @@ export const ContactForm: React.FC = () => {
         <GridItem xs={12} md={6}>
           <InputField
             required={true}
-            inputRef={register({ required: 'Ihr Name fehlt' })}
+            inputRef={register('name', { required: 'Ihr Name fehlt' })}
```
georgiee, line 77: Did that work before because it's implicitly given by the name attribute ?

georgiee (changes requested):
Looks awesome and it works! 💪

+ ✅ I was able to upload a single PDF file.
+ ⛔️ I was able to upload two pdf files but it's not visible in Personio. I only see the second file. Have you tested this scenario? Because in the past we only had one file that can be uploaded but it would be awesome to send mutiple ones. I really would like to support that. Please involve the personio api docs to accomplish it. This can be a second PR though.


Below are some things in general I found while tinkering around with the exmaple.

+ No hover styles at all when I interact with the mouse
+ I can add the same file multiple times, can we prevent this?
+ I think we should introduce a limit of max 3 files. I don't see anyone needing more files. We need to print a message "maximum amount of files reached"
+ I was able to upload a pdf file of 15MB. We need to agree for a limit (10MB?), check the limits of personio and the enforce with an error message if I try to upload such a file (you can query the size client-side I'm sure)
+ I can drop jpg/png files but it's simply ignored. We need information what is allowed and if I try to drop something that is not allowed tell me about it. Error message or forbid the drop (with a changed mouse cursor). Is this possible?
+ I don't know what I can upload nor what I should upload. A small text on the component would help to clarify this:

> Lade hier bitte dein relevanten Dokumente hoch, wie zb Lebenslauf (CV), Motivationschreiben oder Referenzen. 
Erlaubt sind ausschließlich PDF Dateien, die maximale Größe pro Datei beträgt 5MB.

+ (No action required yet) We need to think about this later: Anschreiben is mandatory while I might want to upload a formatted PDF as my Anschreiben. 


Regarding the PR itself, you ship changes to the contact form. That should usually be a separate PR, but it's okay for because it's small enough and probably required by the react form hook update? 🙏

georgiee:
Hello,
how is the progress? Did my feedback help or did I create more problems 😬

src/components/form/contact.tsx
WiebkeFr: The register() broke because of the update from "6.15.8" to "7.8.7" (zum Vergleich: [https://react-hook-form.com/api/useform/register](https://react-hook-form.com/api/useform/register))

src/components/career-form/career-file-preview.tsx
```diff
@@ -3,16 +3,37 @@ import React from 'react';
 import { Clear } from '../icons/clear';
 import { Checkmark } from '../icons/checkmark';
 import { theme } from '../layout/theme';
+import { up } from '../breakpoint/breakpoint';
+import { CareerSelect } from './career-select';
 
-export const FilePreview = ({ index, file, onClick }) => {
+export const FilePreview = ({
```
feedm3: Is there a reason why there are no type definitions like `FilePreviewProps`?

feedm3 (changes requested):
Looks already pretty good 👍 

Noticed some small bugs:
- after a successful upload, the success message should be in the window (scroll to the message)
- the line height on the info text looks wrong
Screenshot 2021-06-28 at 13 40 06

Looking at the code I don't have much to complain other then 2 more general things:
- some components are missing type definitions. I personally type everything to make it more clear of what data is allowed and what not
- some components do "prop drilling", which means some props are passed down several components until the variable is finally used: `CareerForm > FileUpload > FilePreview > CareerSelect (a.e. watch)`. The from component is quite complex, so this isn't a big issue and I expect some complexity and time to understand it. Nevertheless there are some resources and tricks to avoid this. You can check this links, maybe you find some inspiration on how to make those components a bit more simpler:
  - https://kentcdodds.com/blog/prop-drilling
  - https://twitter.com/mjackson/status/1195431511417208834
  - https://www.youtube.com/watch?v=3XaXKiXtNjw
  - https://epicreact.dev/soul-crushing-components/

package.json
WiebkeFr: Yes, the others were already updated. It broke the `register`-function of the input-fields. (The name should be added to the function itself, not only optional validation and error-messages as before)
For reference [https://react-hook-form.com/api/useform/register](https://react-hook-form.com/api/useform/register)

src/components/career-form/career-file-preview.tsx
WiebkeFr: No, i added it 👇

WiebkeFr:
> Noticed some small bugs:
> 

Resolved:
- scroll after successful submit was added 
- general error message below submit `Benötigte Felder fehlen`
- prop-drilling reduced : `CareerForm > Upload > Select`
- icon next to progress bar 
- gender (divers for all) and prop interfaces
- success message

feedm3 (approved):

src/components/career-form/career-components.tsx
```diff
@@ -139,10 +166,33 @@ export const Actions = ({ tryAgainFn, error, isSubmitting }) => {
...
+    <ButtonWrapper>
+      <SendButton type="submit" disabled={isSubmitting}>
+        <ButtonText>Senden</ButtonText>
+        {!isSubmitting && <RightArrowIcon />}
+        {isSubmitting && <span>...</span>}
+      </SendButton>
+      {Object.keys(fieldErrors).length > 0 && (
+        <ErrorWrapper>
+          <FormError
+            error={{ message: 'Bitte füllen Sie alle benötigten Felder aus' }}
```
georgiee: Wir dutzen überall in Karriere

src/components/career-form/career-components.tsx
```diff
@@ -61,14 +61,22 @@ export const InputField = (props: InputFieldProps) => {
     </InputContainer>
   );
 };
-
-export const FormError = ({ error }) => {
+interface FormErrorProps {
```
georgiee, line 64: typing 👍

src/components/career-form/career-components.tsx
```diff
@@ -61,14 +61,22 @@ export const InputField = (props: InputFieldProps) => {
...
+  marginBottom?: number;
```
georgiee, line 65: that feels awkward because it's so specific to the presentation layer. I would rather define a specific error component class instead. We take whatever we need to make it work for now, so I'm fine with that of course 🙏

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+  register: UseFormRegister<FieldValues>;
+  watch: UseFormWatch<FieldValues>;
+  name: string;
+  selectedFiles: FileList;
+  errors: FieldErrors;
+  children: ReactNode | ReactNode[];
+}
+
+const MAX_NUMBER = 3;
+const MAX_SIZE = 20000000; // maximal size per document of personio
```
georgiee, line 35: Screen Shot 2021-07-02 at 13 50 50

confirmed with a 30MB file 👌

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+  clearErrors: UseFormClearErrors<FieldValues>;
+  register: UseFormRegister<FieldValues>;
+  watch: UseFormWatch<FieldValues>;
+  name: string;
+  selectedFiles: FileList;
+  errors: FieldErrors;
+  children: ReactNode | ReactNode[];
+}
+
+const MAX_NUMBER = 3;
```
georgiee, line 34: Screen Shot 2021-07-02 at 13 52 50

Confirmed as well 👌

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+      }
+      for (let i = 0; i < Math.min(maximalOfFiles, droppedFiles.length); i++) {
+        dataTransfer.items.add(droppedFiles[i]);
+      }
+      setValue(name, dataTransfer.files, { shouldDirty: true });
+      clearErrors(name);
+
+      // in case not all file could be added, an error is set
+      if (
+        Math.min(maximalOfFiles, droppedFiles.length) !== droppedFiles.length
```
georgiee, line 74: I love the many details now covered by the form upload

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+  errors,
+  children,
+  watch,
+}: FileUploadProps) => {
+  const onDrop = useCallback(
+    (droppedFiles) => {
+      // maximal number of file being dropped
+      const maximalOfFiles = selectedFiles
+        ? MAX_NUMBER - selectedFiles.length
+        : MAX_NUMBER;
```
georgiee, line 53: next time a if-else condition works wonder to keep the code readable. The ternary operator is send from hell in terms of readability 👹

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+      // dropped files are appended to already selected files
+      const dataTransfer = new DataTransfer();
+      if (selectedFiles) {
+        for (let i = 0; i < selectedFiles.length; i++) {
+          dataTransfer.items.add(selectedFiles[i]);
+        }
+      }
+      for (let i = 0; i < Math.min(maximalOfFiles, droppedFiles.length); i++) {
```
georgiee: good knowledge: the part `Math.min(maximalOfFiles, droppedFiles.length);` is never changing but by writing it into the loop it's being evaluated as often the loop is running. Instead you could initialize the limit in the first part of the loop (where `i=0` is set) as it's run only once. Super tiny detail which is not worth to optimize here though.

georgiee: You calculate that value ` Math.min(maximalOfFiles, droppedFiles.length)` at least two times in this component so you can also just calculate once and store it at the beginning too

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+    },
+    [selectedFiles, setValue],
+  );
+
+  const validator = (file) => {
+    clearErrors(name);
+    if (selectedFiles) {
+      // check if file was already uploaded
+      for (let i = 0; i < selectedFiles.length; i++) {
+        if (JSON.stringify(selectedFiles[i]) === JSON.stringify(file)) {
```
georgiee: `JSON.stringify` can get pretty expansive  and hang peoples browser for large files, I would simply use the file name 😬 You never know what gigantic file people try to upload (although I would question if they.. well other topic)

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+    }
+    // check maximal size of documents
+    if (file.size > MAX_SIZE) {
+      return {
+        code: 'file is too big',
+        message: `Maximal filesize of 20 MB is reached.`,
+      };
+    }
+    return null;
+  };
```
georgiee, line 123: thank you for the many details you are covering know

src/components/career-form/career-file-upload.tsx
```diff
@@ -0,0 +1,238 @@
...
+  flex-direction: column;
+  justify-content: center;
+  ${({ selectedFiles }) =>
+    (!selectedFiles || selectedFiles.length === 0) && `align-items: center;`}
+  border-radius: 4px;
+  ${({ isDragActive }) =>
+    isDragActive && `border: 2px solid ${theme.palette.primary.main};`}
+  padding: ${({ selectedFiles }) =>
+    !selectedFiles || selectedFiles.length === 0 ? `56px` : `24px`};
+  margin-bottom: 2px;
```
georgiee: I love styled component but that is just ugly and barely readable  😑 I see myself writing code like that and maybe it would help to pre-caculate `!selectedFiles || selectedFiles.length === 0)` into a boolean `empty` and pass that instead as a prop?

src/components/career-form/career-form.tsx
```diff
@@ -60,14 +82,20 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
-          'documents',
-          formValues.documents[0],
-          formValues.documents[0].name,
-        );
+        for (let i = 0; i < formValues.documents.length; i++) {
+          const keyName = `categorised_documents[${i}][file]`;
+          const fileName = formValues.documents[i].name.split('.')[0];
+          const category = formValues.category_select[fileName];
+          formData.append(keyName, formValues.documents[i]);
+          const nameCategory = `categorised_documents[${i}][category]`;
```
georgiee, line 91: nice that you figured out the personio api format. I don't like their API description and it's difficult to find the correct combination but you succeeded here👍

src/components/career-form/career-form.tsx
```diff
@@ -78,6 +106,7 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
       .then((json) => {
         if (json.success) {
           // all good
+          props.scrollToStart();
```
georgiee, line 110: nice. sometimes I think we should not do this, because the user is the owner of the scrollbar. But I have seen Fabian's comment and I think it's okay here to do that.

src/components/career-form/career-form.tsx
```diff
@@ -110,119 +139,162 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
+  }
+`;
+
+const CheckboxLabel = styled.label`
+  display: flex;
+  position: relative;
+  flex-direction: row;
+  cursor: pointer;
+
+  div {
```
georgiee: It's always good to name things. You should never target html elements in a specific css statement like you write here

src/components/career-form/career-form.tsx
```diff
@@ -110,119 +139,162 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
+    top: -3px;
+    width: 20px;
+    height: 20px;
+    margin-right: 20px;
+    background: rgba(122, 143, 204, 0.3);
+    border-radius: 4px;
+  }
+
+  &:hover {
+    &:before {
```
georgiee: **First**
pseudo classes and pseudo elements have actually different amount of colons. 
+ pseudo classes like hover: `:hover`
+ pseudo elements like before: `::before`

Here some links:
+ https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements
+ https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

Why does it still work here? Because of this:
> most browsers support both syntaxes for the original pseudo-elements.

This compatibility mesure creates the confusion around the colons I guess (and I share that confusion too)

**Second**
Do yourself a favor and keep nesting to a minimum. It helps the readability and you can perfectly write `&:hover::before {}` here

src/components/career-form/career-select.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface CareerSelectProps {
+  setValue: UseFormSetValue<FieldValues>;
+  clearError: UseFormClearErrors<FieldValues>;
+  register: UseFormRegister<FieldValues>;
+  watch: UseFormWatch<FieldValues>;
+  errors: FieldErrors;
+  file: File;
+}
+
+export const CareerSelect = ({
```
georgiee: So cool to have the category of the document supported. I think the name is kind of misleading. It's more a `CareerFileUploadType` instead of some random career select

src/components/form/controls.tsx
```diff
@@ -57,6 +57,10 @@ export const Input = styled.input<ValidationProps>`
   border: 0;
   background: rgba(122, 143, 204, 0.2);
 
+  &:hover {
+    background: rgba(122, 143, 204, 0.5);
```
georgiee, line 61: you should use the available `polish` (it's a styled component companion) and the function `rgba` to use real values. The rgba values are not natural and you won't find them when you are going to replace/update color values in the future.

georgiee (approved):
Thank you very much for all the changes. Great job everyone, I'm really thankful that we have finally a superb career form 💪 And I hope @WiebkeFr learned a lot while working on this form.

I have left quite some comments, partially with some random knolwedge for Wiebke but also some things I would like to see improved. Let's merge this PR because it's good now and open for too long. Please address the few changes if possible in a small suffix PR.

Thank you 🥳


https://user-images.githubusercontent.com/1701755/124274039-1533b700-db41-11eb-8204-966d189ec6ec.mp4

PR #91, opened 2021-07-07
feat: update all dependencies on a regular schedule

feedm3:
It's time to automate some housekeeping: Updating dependencies 🧹

This Github Action will update all our dependencies to their latest version, commit it to a new branch and open a pull request. This again will trigger the Gatsby Cloud build, so all we have to do is to click on "Merge" if the build is successful.

The Github Action will run every monday at 2am.

### DIFF
--- a/.github/workflows/scheduled-dependency-update.yml
+++ b/.github/workflows/scheduled-dependency-update.yml
@@ -1,0 +1,33 @@
+name: Scheduled Dependency Update
+
+on:
+  # run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+  # every monday at 02:00
+  schedule:
+    - cron: '0 2 * * 1'
+
+jobs:
+  dependency-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+
+      - name: Upgrade dependencies
+        run: yarn upgrade --latest
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          branch: chore-dependency-update
+          delete-branch: true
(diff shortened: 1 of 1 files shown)


.github/workflows/scheduled-dependency-update.yml
```diff
@@ -0,0 +1,33 @@
+name: Scheduled Dependency Update
+
+on:
+  # run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+  # every monday at 02:00
```
georgiee, line 7: thanks for the comment 👍 server time is CET ? That usually bites me. The time doesn't matter actually as it's not holding any service. Just curious.

.github/workflows/scheduled-dependency-update.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          branch: chore-dependency-update
+          delete-branch: true
+          title: "chore: dependency update"
+          commit-message: "chore: dependency update"
+          body: |
+            This PR was automatically created.
+            - all dependencies are updated
+            - this PR will be created every monday at 2am and force pushed
```
georgiee, line 33: force pushed in case the branch gets outdated in the meantime ?

georgiee (approved):
Thanks👍 This is the same Felix created for our insurance project via github actions (enterprise). Great to see how easy dependency automation got after projects like renovator or greenkeeper.io (which already dead)

.github/workflows/scheduled-dependency-update.yml
feedm3: It's UTC. I didn't even think of the timezone, good hint to do so 🙈

.github/workflows/scheduled-dependency-update.yml
feedm3: Just to minimize the error rate, for example if someone did forget to merge the last PR or something changed in the `package.json` in the meantime.

PR #92, opened 2021-07-07
feat: file-type select for career form

WiebkeFr:
zum testen:
[https://satellytescommain-featcareertypeselect.gtsb.io/career/338504-frontend-engineer-(wmx)](https://satellytescommain-featcareertypeselect.gtsb.io/career/338504-frontend-engineer-(wmx))


src/components/career-form/career-file-upload-type.tsx
```diff
@@ -29,40 +40,159 @@ export const CareerFileUploadType = ({
 }: CareerFileUploadTypeProps) => {
   const fileName = file.name.split('.');
   const name = `category_select.${fileName[0]}`;
-  const state = watch(name);
+  const state = watch(name, 'default');
```
feedm3: `state` is quite general. maybe something like `selectedFileType`?

feedm3 (approved):
This is so good, excellent! 🥳

package.json
```diff
@@ -27,6 +27,7 @@
     "precommit": "lint-staged"
   },
   "dependencies": {
+    "@reach/listbox": "^0.15.2",
```
georgiee: I checked the accessbility and I was surprised how perfect the implementation worked. Then I saw that you have pulled in reach ui. Superb library and I have seen that they indeed put a focus on accessibility.

I checked how the dropdown opens/closes, how you can navigate and how the labels are rendered in the accessibility object model👌

src/components/career-form/career-file-upload-type.tsx
georgiee: I can only second this, makes it much more readable.

src/components/career-form/career-file-upload-type.tsx
```diff
@@ -29,40 +40,159 @@ export const CareerFileUploadType = ({
...
+      </StyledListboxInput>
       {errors.category_select?.[fileName[0]] && (
-        <FormError
-          error={{ message: 'Deine Auswahl fehlt' }}
-          lineHeight={100}
-        />
+        <FormErrorWrapper>
+          <FormError
+            error={{ message: 'Deine Auswahl fehlt' }}
+            lineHeight={90}
```
georgiee: We are drifting into Denglish with our mix of German and English 😅 This is not a problem here but I'm looking forward to our proper internationalization support

src/components/career-form/career-file-upload-type.tsx
```diff
@@ -29,40 +40,159 @@ export const CareerFileUploadType = ({
...
+  &:hover{
+    background: ${rgba('#ffffff', 0.2)};
+  }
+  
+  ${up('md')} {
+    width: 144px;   
+    height: 32px;
+  }
+  
+  [data-reach-listbox-button] {
```
georgiee: I guess that's the recommended way to style reach ui ? I didn't expect that they support approaches like styled component although it would have been nice here instead of squeezing in so many rules. 

It's quite some css, do we have any recommendations when and how to split css specific componnets into separate files @feedm3 ?

src/components/career-form/career-file-upload-type.tsx
```diff
@@ -29,40 +40,159 @@ export const CareerFileUploadType = ({
...
+      background: ${rgba('#303B5C', 0.9)};
+    }
+    
+    &[aria-disabled="true"]:hover{
+      background: ${rgba('#212940', 1)};          
+    }
+    
+    &[data-current-selected]{
+        padding-left: 0px;  
+        &::before{
```
georgiee: double colon for pseudo elements and single colon for pseudo selectors, thank you for hearing my feedback 👍

src/components/career-form/career-form.tsx
```diff
@@ -270,7 +270,7 @@ const CheckboxLabel = styled.label`
     content: '';
     display: inline-block;
     position: relative;
-    top: -3px;
```
georgiee, line 273: those magic numbers will bite us in the future, because we won't know if you read this from a design or if this came from some tinkering around values in the browser. I'm interested, where does this come from? I'm fine with the change of course here.

src/components/form/controls.tsx
```diff
@@ -150,6 +150,8 @@ export const SendButton = styled(Button)`
 
 export const Checkbox = styled.input`
   opacity: 0;
+  width: 1px;
+  height: 1px;
```
georgiee: I guess this is to visually hide the checkbox.
Common practices are using an empty clip path or setting opacity to 0. width/height are usually not used, although I don't know why. Why the addition to opacity? Did you have some layout shifts? then position absolute could save you too here. Again nothing to change if it works of course. Just asking for details

georgiee (approved):
Looks great and I had already notice the native dropdown in the current version. Actually I was fine with that, but this nails it. Thanks for the PR. Approved🤘

One thing: Your PR description is empty. Although your closes team members will know the task it's just good practice to reflect on your task, write about the problem and how you approached it and what challenges you had while doing so 👍 Please write a better PR description in your next PR. Thanks.

src/components/career-form/career-file-upload-type.tsx
feedm3: Yep, that CSS is a lot. But actually I have no idea right now on how to split the CSS. It looks like it's necessary to style the reach-ui components. If thats recommended by reach-ui it would be ok for me.

src/components/form/controls.tsx
WiebkeFr: Yes it worked with opacity to 0 and position: absolute 👍

feedm3:
Nice. I tested it and it works flawlessly 💪 Feel free to merge it 🚀

PR #95, opened 2021-07-09
feat: re-add workshop blog post from 2018

georgiee:
Some content from 2018 that is probably not the best blog post but adds some relevance regarding Angular & Workshops to our content collection plus we link the github project.

The workshop itself was amazing and I think we created a really unique experience for a workshop that we should repeat with Angular or React by offering to our clients or interested companies.


georgiee:
I tried to embed to youtube video but the plugins available did not work properly (cache errors in gatsby)
plus they don't respect privacy so resources are loaded directly from the remote server. I think we should just embed a preview image and link to yotube. Right now it's a link.

feedm3 (approved):

PR #98, opened 2021-07-12
feat: improve google indexing

feedm3:
I checked out Google Search Console and found out that we have a bunch of pages that are either not crawled correctly or marked as duplicate content. 

One problem is that amount of link at wich our content can be reach. In worst case, we have 4 different URLs that all point to the same content:

- `http` redirects to `https`
- `www` redirect to an URL without `wwww`
- missing trailing `/` will redirect to `/`

Screenshot 2021-07-12 at 08 31 11

1. For performance reasons, it's best to always point to the URL that doesn't need a redirect. That's why I made sure that all links in the `sitemap.xml` and also all internal links have a trailing slash. This makes all internal links consistent, which is also recommended by the documentation https://developers.google.com/search/docs/advanced/guidelines/duplicate-content

2. Our old `https://satellytes.com/sitemap.xml` was still served, although it's not part of the output folder anymore. It turn out the Gatsby Cloud still adds the root `sitemap.xml`. I started a clean build an hope that it gets removed. Super strange, as this file shouldn't be generated anymore since weeks. Edit: It's gone now.

3. I'm not sure about canonical URLs. The Google Search Console still lists some `www.` URLs, but they all redirect with a 301 to the original file. Due to the documentation this shouldn't be a problem anymore: https://developers.google.com/search/docs/advanced/guidelines/duplicate-content 

There is a discussion in which the author goes in detail on what he did to improve the Google ratings:
- https://github.com/gatsbyjs/gatsby/discussions/27889

**Canonical URLs**
The author in the discussion recommends to set a canonical URL. I tried it with the [gatsby-plugin-canonical-urls
](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls). This plugin added a canonical URL to every page, but not with consistent trailing slashes (for blog posts they were missing). After that I tried [gatsby-plugin-react-helmet-canonical-urls](https://github.com/NickCis/gatsby-plugin-react-helmet-canonical-urls), but it had the same issue. The plugins don't provide a simple function to overwrite the canonical URL so I didn't want to invest anymore time on it. If we want to have it, we need to build it on our own.
I'm not sure if this is really necessary, as all redirect are working properly and should therefore make the canonical URL obsolete.  

**Backlinks**
Users will probably link to our pages without the trailing slash, as this gets removed in the address bar by default. I have no idea if this has a negative impact because we are using only trailing slashes in the `sitemap.xml` and for internal links. Like with the canonical URLs I would think that this shouldn't be a problem as we have proper redirect, but maybe we can ask some SEO expert.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -117,6 +117,12 @@ module.exports = {
     'gatsby-plugin-styled-components',
     'gatsby-plugin-typescript',
     {
+      resolve: 'gatsby-plugin-canonical-urls',
+      options: {
+        siteUrl: 'https://satellytes.com',
+      },
+    },
+    {
       resolve: 'gatsby-plugin-robots-txt',
       options: {
         host: 'https://satellytes.com',
@@ -135,8 +141,13 @@ module.exports = {
       options: {
         excludes: ['/imprint', '/data-privacy'],
         resolvePagePath: ({ path }) => {
-          // trailing slashes are direct links without a redirect
-          return path.endsWith('/') ? path : `${path}/`;
+          if (!path.endsWith('/')) {
+            console.warn(
+              'Path of the page does not end with a slash! For SEO reasons all paths should end with a slash:',
+              path,
+            );
+          }
+          return path;
         },
       },
     },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -134,6 +134,10 @@ module.exports = {
       resolve: `gatsby-plugin-sitemap`,
       options: {
         excludes: ['/imprint', '/data-privacy'],
+        resolvePagePath: ({ path }) => {
+          // trailing slashes are direct links without a redirect
+          return path.endsWith('/') ? path : `${path}/`;
+        },
       },
     },
     {
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -114,7 +114,7 @@ const createCareerPages = async ({ actions }) => {
     generateCard({ title: position.name }, outputFile);
 
     createPage({
-      path: position.satellytesPath,
+      path: appendTrailingSlash(position.satellytesPath),
       component: CAREER_DETAILS_TEMPLATE_PATH,
       context: {
         position,
@@ -124,7 +124,7 @@ const createCareerPages = async ({ actions }) => {
   });
 
   createPage({
-    path: '/career',
+    path: '/career/',
     component: CAREER_TEMPLATE_PATH,
     context: {
       positions: positions,
@@ -158,7 +158,7 @@ const createBlogPages = async ({ actions, reporter, graphql }) => {
   // create a page for each markdown file
   markdownBlogPages.data.allMarkdownRemark.nodes.forEach((node) => {
     createPage({
-      path: node.frontmatter.path,
+      path: appendTrailingSlash(node.frontmatter.path),
       component: BLOG_POST_TEMPLATE_PATH,
       context: {},
     });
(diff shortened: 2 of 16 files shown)


src/pages/imprint.tsx
```diff
@@ -34,6 +34,7 @@ const ImprintPage: React.FC = () => {
       <SEO
         title="Impressum | Satellytes"
         description="Pflichtangaben nach § 5 Telemediengesetz/Impressum"
+        noIndex={true}
```
georgiee, line 37: 👍

georgiee (commented):
Thanks for your detailed description! The only thing I'm missing are references to the search console (screenshots etc?) I can't verify the information you gave because I'm not able to find the information in the console.

## Canonical
Regarding the canonical url I think we can and should use the first plugin you mentioned. The problem with the blog might be the way we create it:

```
path: "/blog/angular-inject-the-injector"
```

I think adding a slash here will fix this.

## Biggest Learning: Always trailing slashes
The biggest learning from your research and the fixes is to go with a single approach and I like the trailing slash. It's also the default of Gatsby. Let's go with that and fix any issues we see on the way.

We could create redirect or similar for content without the trailing slashes but we could also try not to generate any content without a slash in the beginning. If we want to be sure I think a 301 is absolutely fine for that case — given a canonical url that is always of the correct format. We should never generate different canonical urls.

I don't understand this part:
> Users will probably link to our pages without the trailing slash, as this gets removed in the address bar by default.

Why should that be the default? We should eliminate anything on our side that creates them in the first place. I see once I navigate that the links are not updated anymore. Can we prevent that too?

I have checked the different code parts and it looks good, maybe we could clarify the pending issues before merging though, so it's only a comment I provide for now 🙏

feedm3:
> Regarding the canonical url I think we can and should use the first plugin you mentioned. The problem with the blog might be the way we create it:
> 
> ```
> path: "/blog/angular-inject-the-injector"
> ```

Nice find! I will fix it during the day and commit it here.

> Why should that be the default? We should eliminate anything on our side that creates them in the first place.

I doubled checked it, you are right. The URL in the browser is now always with a trailing slash. I think I just have used the wrong deployment URL 🙈 So seems like all URLs should be correct for now and in the future.

feedm3:
Done. Sitemap now has all links with traling slashes:
- https://satellytescommain-featimproveseo.gtsb.io/sitemap/sitemap-0.xml

Canonical is also set with trailing slashes:
<details>
<summary>Screenshot of a blog post page</summary>
grafik
</details>
<details>
<summary>Screenshot of the home page</summary>
grafik
</details>



If there is a page with a path without trailing slash, the build prints a warning:

Screenshot 2021-07-12 at 15 01 40

gatsby-config.js
```diff
@@ -116,6 +116,12 @@ module.exports = {
     },
     'gatsby-plugin-styled-components',
     'gatsby-plugin-typescript',
+    {
+      resolve: 'gatsby-plugin-canonical-urls',
```
georgiee, line 120: 👍

gatsby-node.js
```diff
@@ -214,3 +214,7 @@ exports.onCreateWebpackConfig = function (_ref) {
     });
   }
 };
+
+const appendTrailingSlash = (path) => {
```
georgiee, line 218: nice 👍

georgiee (approved):
Let's go with that, thank you for the additional fixes 👌

PR #104, opened 2021-07-16
feat: add shortSummary to overview when it is available

hofmann-felix:
The short summary of the blogpost on the overview page is now taken from the "shortSummary" tag of the blogpost markdown. This gives the author the possibility to describe his article precisely.

### DIFF
--- a/blog-posts/howto-blog-post.md
+++ b/blog-posts/howto-blog-post.md
@@ -5,6 +5,7 @@ title: "How to do a blog post"
 featuredImage: images/space.jpg
 author: "Georgios Kaleadis & Fabian Dietenberger"
 authorSummary: "CTO and Senior Developer at Satellytes"
+teaserText: "Learn how to properly write a blog post – in terms of formatting – so that everything is rendered properly & our blog looks super shiny."
 attribution:
   creator: upklyak
   source: https://www.freepik.com/free-vector/space-landscape-night_13748451.htm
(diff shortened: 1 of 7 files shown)


feedm3:
I think we should remove the excerpt, as in most cases it doesn't really read well. Having a separate variable lets us define a proper intro/preview text.

For the `shortSummary`: It looks like the primary usecase was the preview text when you share the link. This text may be too short for the preview in our blog post overview, so we may need a separate variable for the blog post.

My proposal:
- `seoMetaText`: contains the text that is currently used as `shortSummary``
- `previewText`: contains the preview of the blog post. I think `preview` fits a bit better then `summary`, as it should more promote the blog post then summarize it

georgiee (changes requested):
Thanks for making this more explicit.

What about `teaserText` instead of `previewText` ?
And I like `seoMetaText` 👍

We then have the following properties to fill. Those fields are mandatory for us.
+ teaserText
+ seoMetaText

We won't use the excerpt as proposed by @feedm3  ? I like the idea. We _could_ use it as a fallback but on the other side I really want the text to be defined anyway.

georgiee:
Please come back to this PR so we can move forward 🙏

hofmann-felix:
The `shortSummary` tag has now been split into `seoMetaText` and `teaserText`. Currently, the `teaserText` tag is not mandatory and the excerpt is used as a fallback. 

I have already adapted the "How to do a blog post" article and added a short explanation about the `teaserText` tag. For the other articles, the teaser text has to be written.

Here is the URL with the current changes: https://build-243a9fb8-d963-4318-a0e2-889c11102b38.gtsb.io/blog/

blog-posts/howto-blog-post.md
```diff
@@ -5,6 +5,7 @@ title: "How to do a blog post"
 featuredImage: images/space.jpg
 author: "Georgios Kaleadis & Fabian Dietenberger"
 authorSummary: "CTO and Senior Developer at Satellytes"
+teaserText: "Learn how to properly write a blog post – in terms of formatting – so that everything is rendered properly & our blog looks super shiny."
```
georgiee, line 8: FYI You don't need quotes for the content in yaml and I mix this up also regularly. You might have copied another yaml line for me here because it's good in the other files 😅

georgiee (approved):
Thanks for the changes. Works. I added a `teaserText` locally to another blog post which is then rendered with the updated text while the seo text is not changed. Let's merge this nice update. Thank you 👍

Screen Shot 2021-09-01 at 09 49 16
Screen Shot 2021-09-01 at 09 46 53

PR #106, opened 2021-07-19
feat: i18n for career and office-page

WiebkeFr:
Pr contains website with multilingual content.

Every page should be available in english. If the content is also available in german, the language can be switched.
To test: [https://satellytescommain-featgatsbypluginreacti18next.gtsb.io/](https://satellytescommain-featgatsbypluginreacti18next.gtsb.io/)

Technical notes:
- the navigation, headlines and other small text blocks are translated using [react-i18next](https://react.i18next.com/)
- url routes of german and english pages work with [gatsby-plugin-react-i18next](https://www.gatsbyjs.com/plugins/gatsby-plugin-react-i18next/)
- for each markdown-file there is an english and german version. Which one is loaded depends on the language provided by [gatsby-plugin-react-i18next](https://www.gatsbyjs.com/plugins/gatsby-plugin-react-i18next/)
- links in seo for google [https://developers.google.com/search/docs/advanced/crawling/localized-versions?hl=de#html](https://developers.google.com/search/docs/advanced/crawling/localized-versions?hl=de#html)


georgiee:
Looking forward seeing progressing through this feature 🥳

feedm3:
Wow, again awesome work 💪 Didn't expect to have everything translated that fast, very impressive 👍 

**Some general findings**
- filter out from sitemap: `/de/data-privacy/` and `/de/imprint/`

**Findings with Screenshots**
<details>
<summary>Headers: alternate URL is wrong</summary>
Screenshot 2021-07-26 at 16 15 51
</details>
<details>
<summary>Headers: MetaTags are not translated (can be tested at https://metatags.io)</summary>
Screenshot 2021-07-26 at 16 24 08
</details>
<details>
<summary>Translations are missing on the blog post</summary>
Screenshot 2021-07-26 at 18 15 04
</details>
<details>
<summary>Blog post still contains german text</summary>
Screenshot 2021-07-26 at 18 15 37

</details>
<details>
<summary>The `pageContext` from the HomePage has the wrong `langKey` in it</summary>
Screenshot 2021-07-26 at 18 15 37

</details>

gatsby-config.js
```diff
@@ -154,5 +154,70 @@ module.exports = {
...
+            matchPath: '/:lang?/career/:id/',
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/:lang?/career/',
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/:lang/office/',
+            getLanguageFromPath: true,
```
georgiee: Why is there a `:lang?` compared to the other `:lang` entries ? Is it optional one time and the other time not?

gatsby-config.js
```diff
@@ -154,5 +154,70 @@ module.exports = {
...
+          },
+          {
+            matchPath: '/:lang?/clients/',
+            getLanguageFromPath: true,
+            excludeLanguages: ['en'],
+          },
+          {
+            matchPath: '/:lang?/services/',
+            getLanguageFromPath: true,
+            excludeLanguages: ['en'],
```
georgiee: why is the language excluded here but not for any other entry ?

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
 
-  const jobsXmlResponse = await fetch(PERSONIO_JOBS_URL);
-  const jobsXml = await jobsXmlResponse.text();
-  const jobsParse = xmlParser.parse(jobsXml);
-  const positions = jobsParse['workzag-jobs'].position;
+  const positionsGroupedByLanguage = LANGUAGES.map(async (langKey) => {
+    const PERSONIO_JOBS_URL_LANG = PERSONIO_JOBS_URL.concat(
+      `?language=${langKey}`,
+    );
+    const jobsXmlResponse = await fetch(PERSONIO_JOBS_URL_LANG);
```
georgiee, line 75: so of course no batch API endpoint given by the very limited Personio API ? Glad that you managed to query for the data anyway.

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
 const createCareerPages = async ({ actions }) => {
```
georgiee, line 68: This is now a monster function with ~150 lines. Wouldn't it be much more readable to place the entire function in a separate file? Just to keep the file readable or is this fine for everyone else?

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
+          position.jobDescriptions.jobDescription,
+        )
+          ? position.jobDescriptions.jobDescription
+          : [position.jobDescriptions.jobDescription];
+
+        position.jobDescriptions.jobDescription =
+          position.jobDescriptions.jobDescription
+            .map((description) => {
+              if (description.name.trim() === 'Slug') {
+                position.satellytesPath = `/career/${description.value.trim()}/`;
```
georgiee: I don't know what `satellytesPath` is but I wonder why the slug doesn't play any role here and instead you extract the description?

src/components/career-form/career-components.tsx
```diff
@@ -150,16 +149,17 @@ export const Actions = ({
   isSubmitting,
   fieldErrors,
 }: ActionsProps) => {
+  const { t } = useTranslation();
   if (error) {
     return (
       <>
         <FormError error={error} />
-        <p>
```
georgiee, line 157: Is `Trans` a `p` by default or do we loose semantic information here by dropping the p ?

src/components/career-form/career-file-upload.tsx
```diff
@@ -76,7 +78,7 @@ export const FileUpload = ({
           name,
           {
             type: 'manual',
-            message: 'Maximal number of documents is reached.',
+            message: t<string>('career.error.max-number'),
```
georgiee, line 148: I checked the error messages first in the preview and everything was translated properly👌

src/components/career-form/career-textfields.tsx
```diff
@@ -80,18 +86,20 @@ export const CareerTextFields = ({
           inputRef={register('salary_expectations')}
           error={errors.salary_expectations}
           name="salary_expectations"
-          label="Gehaltsvorstellung"
+          label={t('career.salary_expectations')}
```
georgiee: I see why you chose to use snake case here (`salary_expectations`) instead of kebap case `salary-expectations` like before. 

The translation doesn't relate to the internal variable name. You should be able to refactor the name without touching the translation. That's why I think the translation keys should have the same naming schema so you should rename the key to `career.salary-expectations`

src/components/career-form/career-textfields.tsx
```diff
@@ -80,18 +86,20 @@ export const CareerTextFields = ({
...
         <InputField
           required={true}
-          inputRef={register('message', { required: 'Dein Anschreiben fehlt' })}
+          inputRef={register('message', {
+            required: t<string>('career.error.message'),
+          })}
           error={errors.message}
           name="message"
-          label="Anschreiben"
+          label={t('career.cover-letter')}
```
georgiee, line 102: i18n is super hard and there are different ways. What I really dislike with i18n next is that they don't require any default value while it's perfectly possible with `i18next.t('key', 'default value to show');` (see https://www.i18next.com/translation-function/essentials).

Without the default translation you remove quite some context and I have  a super hard time to read and understand the component because there is no written text left. 

Would it be technically possible to add default translations (English) everywhere we use i18n-next? I think the `Trans` components would make problems. But have you tried it? It's not crucial for this PR though and I would be happy to discuss this topic afterwards.

src/components/form/contact.tsx
```diff
@@ -74,67 +76,72 @@ export const ContactForm: React.FC = () => {
...
-            label="Name"
+            label={t('contact.name')}
           />
         </GridItem>
         <GridItem xs={12} md={6}>
           <InputField
             required={true}
             inputRef={register('email', {
-              required: 'Ihre E-Mail fehlt',
+              required: t<string>('career.error.email'),
```
georgiee: Interesting "i18n spaghetti issue". We have isolated the contact form but we are reusing i18n ids. I think that's wrong. You should not to that although the labels are the same. Ideally you name it differently and you provide duplicate content on your i18n backend (being a json file? I have not spotted it yet)

gatsby-config.js
```diff
@@ -154,5 +154,70 @@ module.exports = {
...
+          {
+            matchPath: '/:lang?/career/:id/',
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/:lang?/career/',
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/:lang/office/',
```
feedm3: Why is `:lang` and not `:?lang` like on the other pages`

gatsby-config.js
```diff
@@ -154,5 +154,70 @@ module.exports = {
...
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/:lang?/clients/',
+            getLanguageFromPath: true,
+            excludeLanguages: ['en'],
```
feedm3: Why do we need to exclude english here?

gatsby-config.js
```diff
@@ -154,5 +154,70 @@ module.exports = {
...
+          },
+          {
+            matchPath: '/blog/:title/',
+            languages: ['en'],
+          },
+        ],
+      },
+    },
+    {
+      resolve: 'gatsby-plugin-i18n',
```
feedm3: Do we still need this plugin? Shouldn't `gatsby-plugin-react-i18next` do all translation work?

gatsby-node.js
```diff
@@ -3,6 +3,7 @@ const slugify = require('slugify');
 const path = require('path');
```
feedm3, line 7: This file is now definitely too large :D But let's leave it for now and split it up after the merge.

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
 const createCareerPages = async ({ actions }) => {
   const { createPage } = actions;
 
-  const jobsXmlResponse = await fetch(PERSONIO_JOBS_URL);
-  const jobsXml = await jobsXmlResponse.text();
-  const jobsParse = xmlParser.parse(jobsXml);
-  const positions = jobsParse['workzag-jobs'].position;
+  const positionsGroupedByLanguage = LANGUAGES.map(async (langKey) => {
```
feedm3: 1. I'm not a fan of abbreviations. Code must be optimized to be readable for humans, not for machines. With abbreviations I always have to translate the variable name to the "real name", which is cognitive overhead.
My variable name recommendation: `languageKey`
2. I hope I'm not wrong, but `positionsGroupedByLanguage` is also missleading. What it actually is, is `fetchPositionsByLanguage`.

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
-      : [position.jobDescriptions.jobDescription];
-
-    position.jobDescriptions.jobDescription =
-      position.jobDescriptions.jobDescription.map((description) => {
-        return {
-          name: description.name.trim(),
-          value: description.value.trim(),
-        };
-      });
+  await Promise.all(positionsGroupedByLanguage).then(async (values) => {
```
feedm3: This line is a bit complicated because the actual async fetch is hidden in the `Promise.all`. Because 2 things happen here, I would split it up:

```js
const positionsByLanguages = await Promise.all(fetchPositionsByLanguage);
positionsByLanguages.then(async (positionsByLanguage) => {});
```

Edit: Maybe my code recommendation is not 100% correct. I'm wondering about the `for` loop in the next line...

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
+    for (let i = 0; i < values.length; i++) {
```
feedm3: `i` should be `index`

src/components/header/header.tsx
```diff
@@ -136,15 +171,45 @@ const Header: React.FC<HeaderProps> = (props) => {
...
+                  onClick={(e) => {
+                    e.preventDefault();
+                    changeLanguage(lng);
+                  }}
+                  selected={language === lng}
+                >
+                  {lng}
+                </LanguageSwitch>
+              );
+            })}
```
georgiee: what's happening here. We have two times the iteration over `languages` but sometimes it's the `StyledLanguageLink` and the other time the `LanguageSwitch` ?

It's very hard to read too, a single component for the switch would be nicer to read here.

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
+      const positions = values[i].positions;
+      const langKey = values[i].langKey;
+
+      // position is valid if it has a job description and slug
+      const isValidPosition = (position) => {
+        return (
+          position.jobDescriptions.jobDescription &&
```
feedm3: This check is only about the `jobDescription`, this is why I would extract the value first and then check it.

```js
const { jobDescription } = position.jobDescriptions
return (...)
```

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
+        position.jobDescriptions.jobDescription = Array.isArray(
+          position.jobDescriptions.jobDescription,
+        )
+          ? position.jobDescriptions.jobDescription
+          : [position.jobDescriptions.jobDescription];
+
+        position.jobDescriptions.jobDescription =
+          position.jobDescriptions.jobDescription
+            .map((description) => {
+              if (description.name.trim() === 'Slug') {
```
feedm3: Like `PERSONIO_SHORT_DESCRIPTION_NAME`, the `Slug` should be put into a variable.

gatsby-node.js
```diff
@@ -62,75 +64,146 @@ exports.createPages = async (createPagesArgs) => {
...
+         */
+        const fileName = `social-card---career-${slugify(position.name, {
+          lower: true,
+        })}.jpg`;
+        const outputFile = path.join('public', fileName);
+        const publicUrl = `${siteMetadata.siteUrl}/${fileName}`;
+
+        generateCard({ title: position.name }, outputFile);
+
+        const getTranslations = (id) => {
```
feedm3: I think we can savely rename this to `getTranslation` and only return one object, as we will not add any other language then german and english.

src/components/career-form/career-components.tsx
```diff
@@ -17,6 +17,7 @@ import { rgba } from 'polished';
 import { RightArrowIcon } from '../icons/buttons-icons/right-arrow';
 import { FieldErrors } from 'react-hook-form';
 import { Headline } from './career-form';
+import { Trans, useTranslation } from 'react-i18next';
```
feedm3: Sometimes you import those functions from `react-i18next` and sometimes from `gatsby-plugin-react-i18next`. Is there a reason behind this?

src/components/client-list/client-list.tsx
```diff
@@ -120,21 +121,30 @@ const ClientListLinkButton = styled(LinkButton)`
...
+  const { t } = useTranslation();
+  const { language } = useI18next();
+  const localizedName = 'en' === language ? nameEN : name;
   return (
     <ClientListEntryLi>
       <ClientListEntryLink to={path}>
-        <StyledTitle title={`Show client: ${name}`}>{name}</StyledTitle>
-        <StyledTimestamp>Seit {formatDate(start, 'MMMM y')}</StyledTimestamp>
-        <StyledArrow title={`Show client: ${name}`}>
+        <StyledTitle title={`Show client: ${localizedName}`}>
```
feedm3: I think we need to put the whole string into a translation with a variable. 

This should look like
```js
t('some.key', { clientName: localizedName })
```

src/components/career-form/career-form.tsx
```diff
@@ -158,9 +160,11 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
           >
             <>
               {(!selectedFiles || selectedFiles.length === 0) && <Upload />}
-              <div>
-                Drop files to upload or <span>browse</span>
-              </div>
+              <Trans i18nKey="career.action.upload">
```
feedm3, line 231: 1. Why not the `t()` function here (and on all other places where this component is used)?
2. I cannot find `career.action.upload` anywhere in the locales

src/components/career-form/career-form.tsx
```diff
@@ -188,18 +187,20 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
                 <CheckboxLabel htmlFor="privacy-policy">
                   {privacyChecked && <CheckboxMark />}
-                  <PolicyText>
-                    Hiermit bestätige ich, dass ich die{' '}
-                    <TextLink to={PRIVACY_POLICY} className={'policy-link'}>
-                      Datenschutzerklärung
-                    </TextLink>{' '}
-                    zur Kenntnis genommen habe. <Sup aria-hidden={true}>*</Sup>
-                  </PolicyText>
+                  <Trans i18nKey={'career.privacy-policy'}>
```
feedm3, line 263: Same like above: Why not the `t()` function?

src/components/header/header.tsx
```diff
@@ -136,15 +171,45 @@ const Header: React.FC<HeaderProps> = (props) => {
...
+                  key={lng}
+                  to={language === lng ? '' : props.translation || ''}
+                  language={lng}
+                  selected={language === lng}
+                >
+                  {lng}
+                </StyledLanguageLink>
+              );
+            })
+          : props.showLanguageSwitch &&
```
feedm3: I don't get the logic here :D 

First, you check if there are `languages`. If true, you iterate over them. If false, you check for `showLanguageSwitch`. If true, you again map over `languages`.

I guess the this can be simplified to:
```js
(props.showLanguageSwitch && props.translation) ? ... : ...
```

src/components/layout/layout.tsx
```diff
@@ -127,6 +129,8 @@ const Layout = (props: LayoutProps): JSX.Element => {
         siteTitleUrl={props.siteTitleUrl}
         $lightTheme={isLight}
         transparent={props.transparentHeader || Boolean(props.heroImage)}
+        showLanguageSwitch={props.showLanguageSwitch}
```
feedm3: Do we even need this? Or should we just always show translations if they are available? And if we don't have translations, we don't show the button

src/components/header/header.tsx
```diff
@@ -43,6 +44,37 @@ const StyledHeader = styled.header<{
   }
 `;
 
+const LanguageSwitch = styled.button<{ selected?: boolean }>`
```
feedm3: When do we need a `LanguageSwitch` and when a `StyledLanguageLink`? 

I checked the header on the HomePage, and there the `LanguageSwitch` is implemented. Those are buttons. But as they are links, it should be an `a` Tag. Otherwise it's not accessible*.

*I'm not 100% sure how to implement language switches perfectly from an a11y perspective. Let's check some popular pages how they have implemented it.

src/components/links/links.tsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
 import styled from 'styled-components';
-import { GatsbyLinkProps, Link as GatsbyLink } from 'gatsby';
+import { Link as GatsbyLink } from 'gatsby-plugin-react-i18next';
```
feedm3, line 3: What is this Link doing? Aren't all links already "hardcoded" when they reach this `Link` component?

src/components/util/format-date.ts
```diff
@@ -1,8 +1,13 @@
...
+import { de, enGB } from 'date-fns/locale';
 
-export const formatDate = (date: string, dateFormat?: string): string => {
+export const formatDate = (
+  date: string,
+  dateFormat?: string,
+  locale?: string,
+): string => {
   const parsedDate = parseISO(date);
+  const loc = locale == 'de' ? de : enGB;
```
feedm3: No abbreviations please.

src/pages/404.tsx
```diff
@@ -28,3 +29,17 @@ const NotFoundPage: React.FC = () => (
 );
 
 export default NotFoundPage;
+
+export const query = graphql`
+  query ($language: String!) {
```
feedm3, line 40: 1. All Queries should have a name (this is for all page Queries in the PR). Here it could be `NotFoundPageQuery`
2. If you use the locales Query on multiple places, you can create a GraphQl Fragment to not have to write it again and again
3. `nodes` instead of `edges -> node`

src/pages/blog.tsx
```diff
@@ -10,106 +10,135 @@ import { PageTitle, Text } from '../components/typography/typography';
...
+          <BlogPageTitle>{t('navigation.blog')}</BlogPageTitle>
         </GridItem>
         <GridItem md={8}>
-          <BlogPageSubTitle>
-            Our developers and designers talk about the latest trends and
-            know-how around tech, especially about our latest learnings and
-            insights on creating good software for the web.
-          </BlogPageSubTitle>
+          <BlogPageSubTitle>{t('blog.info')}</BlogPageSubTitle>
+          {language != 'en' && (
```
feedm3: The checks in the JSX - as well as the content of the checks - should be kept to a minimum, as they are quite hard to read.

In this case, you can outsource the actual content:

```
const BlogPostOverview = ({ data }) => {}
const OnlyEnglishInfo = () => {}

const BlogPage = ({ data }: BlogPageProps) => {
  return (
    <GridItem>
      <BlogPageTitle>{t('navigation.blog')}</BlogPageTitle>
    </GridItem>
    {language == 'en' ? 
       <BlogPostOverview data={data} /> : 
       <OnlyEnglishInfo />
  }
}
```

feedm3 (approved):

feedm3 (approved):
Clicked on "Approve" by accident. After the changes and recommendations are resolved I will approve again.

feedm3 (changes requested):
I spent too much time on this PR, clicked again on "Approve". I feel very stupid now :D

gatsby-config.js
```diff
@@ -154,5 +164,44 @@ module.exports = {
...
+            getLanguageFromPath: true,
+          },
+          {
+            matchPath: '/blog/:title/',
+            languages: ['en'],
+          },
+        ],
+      },
+    },
   ],
```
georgiee: here comes the i18n magic 🥳

src/components/header/header.tsx
```diff
@@ -136,18 +170,69 @@ const Header: React.FC<HeaderProps> = (props) => {
...
+            >
+              {lng}
+            </StyledLanguageSwitch>
+          );
+        })}
+      </div>
+    );
+  }
+};
+
```
georgiee: The language switch deserves its own component. This will make the header component much more readable again.

src/components/navigation/navigation.tsx
```diff
@@ -239,69 +241,69 @@ const Navigation: React.FC = (props) => {
                     to="/imprint"
                     $isSelected={activePath.includes('/imprint')}
                   >
-                    Impressum
+                    {t('navigation.imprint')}
```
georgiee, line 244: I have seen this yesterday on the blog pages (some missing translation context?) but I think it's fixed on the latest version ? Can you confirm?
Screen Shot 2021-07-26 at 18 07 29

src/components/seo.tsx
```diff
@@ -103,6 +120,23 @@ const SEO: React.FC<SeoProps> = ({
...
+            language !== lang && translation
+              ? prependAndAppendTrailingSlash(translation)
+              : pathname;
+          return (
+            <link
+              rel="alternate"
+              href={href.concat(pathname)}
+              hrefLang={lang}
+              key={lang}
+            />
```
georgiee: so this seems to add an alternate entry to refer to the translated version. Imagine you replace that entire code with a function named `listLocalizedVersions()` like so

```suggestion
      {listLocalizedVersions()}
```

That would make it so much more readable because I know what the piece of code is intended to do. Inside the function implementation I could then add more information regarding the alternate links because this is not common knowledge. I looked it up here https://developers.google.com/search/docs/advanced/crawling/localized-versions and only then I understood what's happening here. This cost me at least 5min to understand. You can help your future self by separating the code into a separate file, give the function a name and provide some comments about the implementation.

This tip is valid for most of the lengthy files I have seen in this PR, maybe you could invest some refactoring. Either within this PR or afterwards (to come to an end because the PR got too big already).

src/components/util/format-date.ts
```diff
@@ -1,8 +1,15 @@
...
+export const formatDate = (
+  date: string,
+  dateFormat?: string,
+  locale?: string,
+): string => {
   const parsedDate = parseISO(date);
+  const currentLocale = locale == 'de' ? de : enGB;
 
-  return format(parsedDate, dateFormat || 'do MMMM y', { locale: de });
+  return format(parsedDate, dateFormat || 'do MMMM y', {
```
georgiee, line 12: 👍

src/locales/de/translations.json
```diff
@@ -0,0 +1,107 @@
+{
```
georgiee, line 1: I see that this is how translation works usually with react (and the web). I usually prefer the extraction approach, where a parser finds all translation content and then creates a file automatically. That way you never forget any translation or you carry around obsolete ones.

Do you know how this is managed here? Can you prevent the translations from getting outdated (some "coverage" test?) What happens if you forget something?

The systems that can extract the content can usually extract the current status and you would process the content through a specialized translation backend. Something like poeditor.com where you would manage the translations and generate the technical translation files only as an export/artefact. I really like to work that way but I see where we are coming from here. I only wanted to put this into context of other technical possibilities. Thanks for putting that file together and providing a first translation for all of our content 🙌🥳

georgiee: FYI
If I import the file it's perfectly recognized by poeditor. The hierarchy is reflected as the context so you always know what you are looking at. 
Screen Shot 2021-07-27 at 16 39 58


The keys from the json file are called the `terms` and you can name a default language so you always know what the key is supposed to tell in any other language. 
Screen Shot 2021-07-27 at 16 38 32


They also have a machine learning API:
Screen Shot 2021-07-27 at 16 39 11

I'm a big fan of poeditor for the translation purposes, solves many problems especially if you have to deal with many languages. We used that at our insurance client a lot.

src/locales/en/translations.json
```diff
@@ -0,0 +1,107 @@
+{
```
georgiee, line 1: The file is located at `en/translations.json` but it's the only file. I would remove the folder and use the the name `translation.en.json` or so. We don't need artificial nesting if not necessary.

src/pages/blog.tsx
```diff
@@ -10,106 +10,139 @@ import { PageTitle, Text } from '../components/typography/typography';
...
+          <BlogPageTitle>{t('navigation.blog')}</BlogPageTitle>
         </GridItem>
         <GridItem md={8}>
-          <BlogPageSubTitle>
-            Our developers and designers talk about the latest trends and
-            know-how around tech, especially about our latest learnings and
-            insights on creating good software for the web.
-          </BlogPageSubTitle>
+          <BlogPageSubTitle>{t('blog.info')}</BlogPageSubTitle>
+          {language != 'en' && <OnlyEnglishInfo onClick={onClick} />}
```
georgiee, line 75: I like the fact we decided for a single language for the blog. Makes things much easier and it solved very good technically here on the blog overview page 🙌

src/pages/data-privacy.tsx
```diff
@@ -42,3 +40,23 @@ const DataPrivacyPage: React.FC = () => {
 };
 
 export default DataPrivacyPage;
+
+export const DataPrivacyPageQuery = graphql`
+  query ($language: String!) {
+    locales: allLocale(filter: { language: { eq: $language } }) {
```
georgiee, line 48: Looks so easy how you can use multiple content in the end. Thanks for putting everything in the right place 👍

src/components/header/header.tsx
```diff
@@ -136,18 +170,71 @@ const Header: React.FC<HeaderProps> = (props) => {
...
```
georgiee: Der Language Switch verdient eine eigene Komponente. Macht die Header Komponente viel lesbarer.
(Ich schreibe diesen Kommentar das dritte Mal weil das bei jedem Update dieser Datei rausgeworfen wird 😄)

src/templates/career-details.tsx
```diff
@@ -94,11 +99,17 @@ const CareerPage = ({ pageContext }: CareerPageProps): JSX.Element => {
...
+      <Layout
+        siteTitleUrl="/career/"
+        transparentHeader={true}
+        translation={pageContext.translation}
+        showLanguageSwitch={Boolean(pageContext.translation)}
+      >
         <SEO
           imageUrl={pageContext.socialCardImage}
           title={`Karriere - ${pageContext.position.name} | Satellytes`}
           description={`Bewirb dich jetzt als ${pageContext.position.name}! Schaue Dir unsere anderen offenen Stelle an. Wir freuen uns auf Deine Bewerbung.`}
```
georgiee: this seems to be fixed german content. Is this correct?

georgiee (approved):
That looks great! Thank you for putting so much effort into this. This is like the full i18n package. I can't think of anything more 🥳

I have noted some general issues, there is no big blocker but some things you should improve mostly related to refactoring which is nice to approach after this is merged because this PR got way too big! I understand how you came to this point but we should really, really prevent this in the future. It's much better to deliver a smaller piece. Just for your information: This is for our website so I'm fine with that size. In bigger projects and bigger teams this PR would be close to be rejected and you would have to split the changes into smaller PRs for sure.

Regarding the big size: I spend 30-60min in total for the review across two days as I wanted to check at least every single file. I didn't bother to run it locally or to check the actual text because this was so exhausting. I also don't want to overwhelm you with my a too detailed feedback, otherwise this PR stays open for much longer which is also frustrating for you.

Next time smaller PR if possible but for now I'm super happy that you deliver and solve i18n within a single PR🤘 Let's merge once you have fixed the other issues from Fabian and reviewed my refactoring comments. I will approve the changes already, but I trust you to give me some feedback before moving forward. Thank you!

gatsby-config.js
WiebkeFr: It used the exception when I only translated the career und career-detail-page. I deleted it 👍

gatsby-node.js
WiebkeFr: It is the only way to get the slug from Personio (to put it in the "Kurzbescheibung"). And satellytesPath is the pathname of the career-detail-page.

src/locales/en/translations.json
WiebkeFr: the structure seems to be set to that: [https://www.gatsbyjs.com/plugins/gatsby-plugin-react-i18next/#youll-also-need-to-add-language-json-resources-to-the-project](https://www.gatsbyjs.com/plugins/gatsby-plugin-react-i18next/#youll-also-need-to-add-language-json-resources-to-the-project)

src/components/career-form/career-components.tsx
WiebkeFr: Yes except a different tag is given

gatsby-config.js
WiebkeFr: This was a mistake. I changed it to `:lang?`.

src/templates/career-details.tsx
WiebkeFr: No. I translated it 👍

src/components/career-form/career-components.tsx
georgiee: that's what I meant. `p` transports a specific semantic "being a paragraph" but I think there is not downsides, not even from a11y side. I was just asking as we should not do the same with headlines for example because this would have an impact on different levels 🙏

src/locales/en/translations.json
georgiee: thanks, makes sense then 👍

src/components/header/header.tsx
WiebkeFr: Now they are both implemented as a link and in a single component 👍

src/locales/de/translations.json
WiebkeFr: Thank you! That looks very useful!!

gatsby-node.js
WiebkeFr: I changed the name from values to positionsByLanguage

gatsby-node.js
WiebkeFr: I would leave it for now and refactor in a new pr..

gatsby-node.js
```diff
@@ -15,6 +15,7 @@ const CAREER_DETAILS_TEMPLATE_PATH = path.resolve(
 const { siteMetadata } = require('./gatsby-config');
 const PERSONIO_JOBS_URL = 'https://satellytes.jobs.personio.de/xml';
 const PERSONIO_SHORT_DESCRIPTION_NAME = 'Kurzbeschreibung';
+const SLUG = 'Slug';
```
feedm3: It's the name of the Personio slug variable, so it should be named like this.

`PERSONIO_SLUG`

gatsby-node.js
```diff
@@ -75,136 +76,137 @@ const createCareerPages = async ({ actions }) => {
...
-      const getSlug = (position) => {
-        if (position.satellytesPath) return position.satellytesPath;
-        const slug = position.jobDescriptions.jobDescription.find(
-          (description) =>
-            description.name.trim() === 'Slug' && description.value.trim(),
-        );
-        return `/career/${slug.value.trim()}/`;
-      };
+        const getSlug = (position) => {
+          if (position.satellytesPath) return position.satellytesPath;
```
feedm3: Single line `if` are hard to read. They also should follow with an empty line so that they can be seen at the first sight.

```
if (position.satellytesPath) return position.satellytesPath;
const slug = position.jobDescriptions.jobDescription.find()
```

vs.

```

if (position.satellytesPath) {
  return position.satellytesPath;
}

const slug = position.jobDescriptions.jobDescription.find()
```

gatsby-node.js
```diff
@@ -75,136 +76,137 @@ const createCareerPages = async ({ actions }) => {
...
+          const slug = position.jobDescriptions.jobDescription.find(
+            (description) =>
+              description.name.trim() === SLUG && description.value.trim(),
+          );
```
feedm3, line 119: if there is no slug, a warning needs to be printed and the page shouldn't be generated

gatsby-node.js
```diff
@@ -75,136 +76,137 @@ const createCareerPages = async ({ actions }) => {
...
+          position.jobDescriptions.jobDescription = Array.isArray(
+            position.jobDescriptions.jobDescription,
+          )
+            ? position.jobDescriptions.jobDescription
+            : [position.jobDescriptions.jobDescription];
+
+          position.jobDescriptions.jobDescription =
+            position.jobDescriptions.jobDescription
+              .map((description) => {
+                if (description.name.trim() === SLUG) {
```
feedm3: Why do we check here for the `SLUG` and in `getSlug()`. Can we reuse the `getSlug()`function here?

package.json
```diff
@@ -73,8 +73,8 @@
     "react-dom": "17.0.2",
     "react-dropzone": "^11.3.4",
     "react-helmet": "6.1.0",
-    "react-hook-form": "^7.11.1",
     "react-i18next": "^11.11.3",
+    "react-hook-form": "^7.12.0",
```
feedm3: Can we remove `react-i18next`? There shouldn't be any import of it anymore, right? Which means we can delete the dependency.

src/components/client-list/client-list.test.tsx
```diff
@@ -14,6 +14,25 @@ const testClients = [
   },
 ];
 
+jest.mock('gatsby-plugin-react-i18next', () => ({
```
feedm3, line 17: It's best practice to put all functions inside the corresponding jest lifecycle method. In this case, the mocking should be put into `beforeAll()` or `beforeEach`.

https://jestjs.io/docs/setup-teardown

src/components/header/header.test.tsx
```diff
@@ -5,6 +5,20 @@ import { render } from '@testing-library/react';
 import { ThemeProvider } from 'styled-components';
 import { theme } from '../layout/theme';
 
+jest.mock('gatsby-plugin-react-i18next', () => ({
```
feedm3, line 8: Same as comment above. Please put into lifecycle method to avoid side effects.

src/components/header/header.tsx
```diff
@@ -171,35 +139,10 @@ const Header: React.FC<HeaderProps> = (props) => {
...
-                    changeLanguage(lng);
-                  }}
-                  selected={language === lng}
-                >
-                  {lng}
-                </LanguageSwitch>
-              );
-            })}
+      <Wrapper>
+        {(props.translation || props.showLanguageSwitch) && (
```
feedm3, line 148: If there are no translations, what will the `LanguageSwitch` show?

src/components/header/header.tsx
```diff
@@ -209,10 +152,15 @@ const Header: React.FC<HeaderProps> = (props) => {
...
           {!isNavigationVisible ? <BurgerMenu /> : <CloseBurgerMenuIcon />}
         </SiteMenu>
-      </div>
+      </Wrapper>
       <NavigationFlyout visible={isNavigationVisible} />
     </StyledHeader>
   );
 };
 
+const Wrapper = styled.div`
```
feedm3: Please put above the component that uses the `Wrapper`.

src/components/language-switch/language-switch.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ${({ selected }) => (selected ? `color: #668CFF;` : `color: #FFFFFF;`)}
+`;
+
+export const LanguageSwitch = ({ translation }) => {
+  const { languages, language, originalPath } = useI18next();
+
+  const getTo = (languageOfLink) => {
+    if (language === languageOfLink) return '';
+    if (translation) return translation;
+    else return originalPath;
```
feedm3: Single line `else` is even harder to read than the `if`, because there is no indication to which ìf`this `else` belongs.

src/components/language-switch/language-switch.tsx
```diff
@@ -0,0 +1,47 @@
+import { useI18next } from 'gatsby-plugin-react-i18next';
+import React from 'react';
+import styled from 'styled-components';
+import { Link as LanguageLink } from 'gatsby-plugin-react-i18next/dist/Link';
+
+const StyledLanguageLink = styled(LanguageLink)<{ selected?: boolean }>`
```
feedm3: I'm not so happy about the usage of `LanugaLink`. We should use our common `Link` component everywhere to a central place to handle links. With this component here we create an exception - we now have 2 places where links are rendered.

Links need to be treated all in the same way, regardless of where they appear. If we have 2 places where we create links, we always need to keep them in sync. 

Is it somehow possible to use our own `Link` component here?

src/components/seo.tsx
```diff
@@ -51,14 +52,39 @@ const SEO: React.FC<SeoProps> = ({
   let pathname;
   if (typeof window !== 'undefined') {
```
feedm3: We cannot use this check in the render function. This will lead to a different page that gets rendered on SSR and in the browser (important initial render). We also need to make sure the set the headers independently of the browser, to have them in place on the very first request.

The translations need to be passed into the SEO component directly from Gatsby. This array should then be used to render the correct headers.

src/pages/404.tsx
feedm3: The name of the query should also be put in the GraphQL query itself:

```
query NotFoundPageQuery ($language: String!) {
```

feedm3 (changes requested):

src/components/header/header.tsx
WiebkeFr: default is `showLanguageSwitch = true` (eg. blog has no language switch)
if a translation is provided, it is chosen as new pathname.

package.json
WiebkeFr: Nein `gatsby-plugin-react-i18next` basiert darauf. Löschen konnte ich aber `gatsby-plugin-18n`

georgiee:
FYI: I resolved the package.json conflict through the github UI

georgiee:
@feedm3 what's left before merging? If nothing is left we can merge otherwise I can help to fix.

georgiee:
I have pushed a fix for the personio quirk adding `<p>` to any field including our slug workaround field. I use the regex approach to strip the html taken from https://stackoverflow.com/a/822464 fully aware of the shortcomings of that approach but I think it's super fine for our limited use case here. I don't need to pull in any html parser for that matter.

I didn't want to blindly change the code, that's why I added a helper method plus spec to be sure that it's doing what we are expecting without affecting any other code 🙏 I hope the test will run correctly because we didn't have any tests yet outside of our web sources.


before/after
Screen Shot 2021-08-03 at 16 09 45
Screen Shot 2021-08-03 at 16 28 20

feedm3 (approved):

PR #111, opened 2021-07-29
feat: transition of burger menu

WiebkeFr:
pr contains the hover-transition of the burger menu (button to open flyout menu)
to test: [https://satellytescommain-feattransitionburgermenu.gtsb.io/](https://satellytescommain-feattransitionburgermenu.gtsb.io/)


feedm3 (changes requested):
Nice! Can you make the hover area a bit larger? Currently, the mouse pointer needs to be exactly on the bar. The hover area should be the same as the click area.

feedm3 (approved):

PR #115, opened 2021-08-04
fix: improve size of links to make them more accessible

feedm3:
Lighthouse score was complaining about the size of our legal links. For a good reason, they are too small and hard to use on mobile. I made them a bit larger.


georgiee (approved):
I checked before and after. Looks good. Thanks!
Screen Shot 2021-08-04 at 17 52 48

PR #117, opened 2021-08-05
i18n updates hot fixes

### LINKED ISSUE #116: i18n Update Issues
1. Success Button is not translated
Screen Shot 2021-08-05 at 08 18 19

Fix: Use the correct id which is not `career.sent` but `career.action.sent`

Question: How to prevent that in the future. As suggested before a translation backend like poeditor could help but for that the label must appear in the json in the first place. Is there an extraction available?

2. Submitted Name = First Name + First Name, is it a bug?
We received a few applications (with the used channel being our website) where the first name is given as the first and last name. Initially I thought it's a bug in our recently updated (i18n refactoring) form but my testing wasn't able to confirm that problem. On the other the hand the past three applications had their first name set for both fields: first & last name. Coincidence? 

Fix: Unknown. Can't reproduce.

Screen Shot 2021-08-05 at 08 32 49

3. Dropdown color (Firefox)
The dropdown renders white text onto a white background. I enforce a white/black layout until we can clarify why this happens and if we should even keep the custom dropdown. It seems that it's not accepted in chrome anyway?

Screen Shot 2021-08-05 at 08 18 01
Screen Shot 2021-08-05 at 08 24 09

Fix: white/black layout 

4. Translations
Wrong key `career.actions.alternative` instead of `career.action.alternative`.
General a lot of problem inside the markdown files (duplicate sentences) and hilarious sentences. At some places I just removed the part of I quickly tried a another wording. But I didn't won't to spend much time on translating that content right now.

georgiee:
close #116
See the linked issue for details, we have quite some issues with the auto-translated content and some minor technical things I was able to fix too.

### DIFF
--- a/data/pages/clients-de.md
+++ b/data/pages/clients-de.md
@@ -3,7 +3,7 @@ title: Kunden
 language: de
 ---
 
-Da diese meist streng geheim sind, schätzen unsere Auftraggeber Diskretion. Sie verstehen sicher, dass wir deshalb an
+Da diese Strategien meist streng geheim sind, schätzen unsere Auftraggeber Diskretion. Sie verstehen sicher, dass wir deshalb an
 dieser Stelle keine konkreten Namen nennen möchten, sondern uns auf generische Branchenbezeichnungen beschränken. Unsere
 vier größten Kunden sind:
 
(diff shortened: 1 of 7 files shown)


georgiee:
I carefully self-reviewed the changes and use my admin rights to merge to bring the changes online. I hope that's okay 🤞

PR #122, opened 2021-08-16
feat: added language-switch to footer

WiebkeFr:
pr contains:
-  language-switch added to footer 
- color change of footer

to test:
[https://satellytescommain-featlanguageswitchinfooter.gtsb.io/](https://satellytescommain-featlanguageswitchinfooter.gtsb.io/)

### DIFF
--- a/src/components/icons/social/github.tsx
+++ b/src/components/icons/social/github.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 interface GithubProps {
   opacity?: string;
 }
-const GithubIcon: React.FC = (props) => {
+const IconGithub: React.FC<GithubProps> = (props) => {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
@@ -18,7 +18,7 @@ const GithubIcon: React.FC = (props) => {
   );
 };
 
-export const GithubIconWrapper = styled(GithubIcon)<GithubProps>`
+export const GithubIconWrapper = styled(IconGithub)<GithubProps>`
   transition: fill 0.2s ease-in;
   fill: #ffffff;
 
--- a/src/components/icons/social/github.tsx
+++ b/src/components/icons/social/github.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
 
+interface GithubProps {
+  opacity?: string;
+}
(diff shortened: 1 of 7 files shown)


src/components/icons/social/github.tsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
 
+interface GithubProps {
```
feedm3, line 4: Thats only for the `GithubIconWrapper` right? Then the naming should be `GithubIconWrapperProps`. The interface should also be right above the implementation that you can see with one sight what the props are.

src/components/icons/social/github.tsx
```diff
@@ -15,11 +18,12 @@ const GithubIcon: React.FC = (props) => {
...
 
-export const GithubIconWrapper = styled(GithubIcon)`
+export const GithubIconWrapper = styled(GithubIcon)<GithubProps>`
   transition: fill 0.2s ease-in;
-  fill: #202840;
+  fill: #ffffff;
 
   &:hover {
     fill: #ffffff;
+    opacity: 1;
```
feedm3, line 27: props are not used here?

src/components/icons/social/linkedin.tsx
```diff
@@ -22,8 +23,9 @@ export const IconLinkedIn: React.FC<LinkedinProps> = (props) => (
 
 export const LinkedinWrapper = styled(IconLinkedIn)<LinkedinProps>`
   transition: fill 0.2s ease-in;
-  fill: ${(props) => (props.color ? props.color : '#202840')};
+  fill: ${(props) => (props.color ? props.color : '#FFFFFF')};
   &:hover {
     fill: ${(props) => (props.color ? props.color : '#ffffff')};
+    opacity: 1;
```
feedm3, line 29: also `opacity` prop is not used here

feedm3 (changes requested):

src/components/icons/social/github.tsx
WiebkeFr: It is not used in `GithubIconWrapper` but in the component `IconGithub` itself with `{...props}` 
(just like in `IconLinkedIn`)
Shall I write it extra with `opacity={props.opacity}`?

feedm3 (approved):

PR #124, opened 2021-08-16
feat: process accordion for career page

WiebkeFr:
pr contains: process accordion for career page

technical notes:
- Reach UI for accordion: [https://reach.tech/accordion/](https://reach.tech/accordion/)

to test:
[https://satellytescommain-featprocessaccordionforcareerp.gtsb.io/career/](https://satellytescommain-featprocessaccordionforcareerp.gtsb.io/career/)


src/components/process-accordion/process-accordion.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <StyledAccordionButton>
+          <ProcessCaption>
+            {t('career.process.application.title')}
+          </ProcessCaption>
+        </StyledAccordionButton>
+        <StyledAccordionPanel>
+          <LaunchingRocket />
+          {t('career.process.application.text')}
+        </StyledAccordionPanel>
+      </StyledAccordionItem>
```
georgiee, line 104: I would have expected something like this:

```
<OurAccordionItem title={} illustration={LaunchingRocket}>
 {/** content **/}
</Accordion>
```

Where `OurAccordionItem` is our custom component bundling everything that is repeated. This is some code golfing idea here but some obvious one I think. Because it would also solve the issue with the toggle icons being inline in css here ( see other comment)

src/components/process-accordion/process-accordion.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  &:last-child {
+    border-bottom: 1px solid rgba(255, 255, 255, 0.1);
+  }
+
+  &[data-state='open'] {
+    position: relative;
+    &::after {
+      content: url("data:image/svg+xml;utf8,<svg width='12' height='8' viewBox='0 0 12 8' fill='none' xmlns='http://www.w3.org/2000/svg'>
```
georgiee, line 68: I don't like those inline icons. Let's refactor so we can use icon components instead.

georgiee (changes requested):
Very nice! I missed this PR.
1. Maybe we should move the process to the top (before the opening)?
2. I have some refactoring ideas

georgiee:
technically this is covered by the new component https://github.com/satellytes/satellytes.com/issues/178
and in addition I have created an issue to cover the career page changes we might need including the additional content of describing the application process which is new and covered by this PR as well.

https://github.com/satellytes/satellytes.com/issues/204

PR #129, opened 2021-08-25
feat: added style for anchor links 

WiebkeFr:
pr contains (example) style for anchor links in markdown files.

to test: [https://satellytescommain-featmarkdownanchorlinks.gtsb.io/clients/](https://satellytescommain-featmarkdownanchorlinks.gtsb.io/clients/)

Bildschirmfoto 2021-08-25 um 15 26 37


georgiee (dismissed):
I think it would be much more cleaner to provide a `AnchoredLink` component that I can use explicitly. On the other side I see and I understand that it's not that easy to `tag` some markdown content to switch to another component. This will work but I fear that it might have effects on things we don't expect because you add this anchor prefix icon virtually to every link we create. That's too much especially with the fact that we only wanted to add the anchor to the few headlines on the clients page.

I'm unsure how to deal with that though. I would like to block this feature until clarified.

georgiee:
I've pushed a small refactoring to provide a distinct component which makes use of an icon instead of pushing everything in the existing component and by using inline svg. That just smelled here.

On the other hand I think there is indeed no way to make this such a global change because of the way we replace components in markdown documents.

I think this is now good enough to be merged 👍

src/components/markdown/custom-components.tsx
```diff
@@ -54,11 +56,42 @@ export const SimpleLink = styled.a`
...
+  
+  &:hover {
+    text-decoration: none;
+    ${AnchorIconStyled} {
+      visibility: visible;
+    }
+  }
+}
+`;
+
```
georgiee, line 94: this is a little bit more code then the previous inline svg solution but it fits pretty good into the existing way of having icon components and explicit components per use case.

src/components/markdown/custom-components.tsx
```diff
@@ -118,6 +151,12 @@ const customSatellytesComponents = {
     return <DetailsStyled {...props}>{props.children}</DetailsStyled>;
   },
   a(props) {
+    if (props.href && props.href?.startsWith('#')) {
+      return (
+        <AnchoredLinkStyled {...props}>{props.children}</AnchoredLinkStyled>
+      );
+    }
```
georgiee, line 158: before this condition was buried in styled components now it's a a much clearer place 🙏

feedm3:
Nice! There is still one design issue:

- The text color and the margin (top 160, bottom 24 on desktop) is not correct

I tried to rush it and fix it but wasn't successful within 5 minutes. Will come back to the issue and do the fix the next days.

Screenshot 2021-09-16 at 08 54 42

georgiee:
This is outdated unfortunately. We lost the feature of linking to headlines in the new design apparently ? 
I don't like that the headlines are colored like links but I like the github like anchor icon appearing to the left. Mayeb we can request that in the design at some point? Where would we want to use it ? In the blog only?

PR #130, opened 2021-08-25
feat: copy of code block in blog post

WiebkeFr:
pr contains icon and logic to copy code-blocks of blog posts to the clipboard.

to test: [https://satellytescommain-featcopycodeblock.gtsb.io/blog/github-billing-dashboard/](https://satellytescommain-featcopycodeblock.gtsb.io/blog/github-billing-dashboard/)


package.json
```diff
@@ -33,6 +33,7 @@
     "babel-plugin-styled-components": "^1.13.2",
     "canvas": "^2.8.0",
     "cheerio": "1.0.0-rc.10",
+    "copy-to-clipboard": "^3.3.1",
```
georgiee, line 36: I checked the npm page and it looks like this is a widely used tool and actively maintained 👍
https://www.npmjs.com/package/copy-to-clipboard

src/components/markdown/code-block.tsx
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+  }
+`;
+
+export const CodeBlock = (props: CodeBlockProps) => {
+  const copyCodeBlock = (event) => {
+    const codeNode = event.currentTarget.nextSibling;
+    const { innerText } = codeNode;
+    copy(innerText);
```
georgiee: that's pretty indirect, can't we explicitly query for the  `pre` tag ? This currently relies on a fixed DOM order of the icon followed by the pre tag.

Screen Shot 2021-08-31 at 12 01 17

georgiee (dismissed):
That works 🙌 Great convenience. I'm missing some (visual) feedback when I click the icon. I always think I have missed the click without it even though it works. We can deliver this later though.

There is one issue regarding the even and DOM order which I would like to fix before merge though. Thank you.

georgiee:
FYI github itself has a nice UI confirmation we could adapt

https://user-images.githubusercontent.com/1701755/131630924-7ffaf326-2620-4db3-bb58-0a7008422869.mp4

georgiee:
I wanted to solve the nextSibling issue by being more explicit. We already receive the children content so we could access it for the copy action instead of referring to the actual DOM.

I use a "only text" function cherry picked from a random repo (https://github.com/fernandopasik/react-children-utilities/blob/master/src/lib/onlyText.ts). Works pretty good. While being at it I wanted to drop in a checkmark that appears on copy and fadeouts. I tried to use `onAnimationEnd` on the component but styled components seems to interfere because I don't get any callbacks for my animation. I stop here because the important refactoring I wanted is present now. Maybe @feedm3 you know some insights with `onAnimationEnd` in react? Plan was to add a state flag when copied, this will add the checkmark which will animate once (hidde, show, hidden). Once the cycle is completed (animation end) I would reset the state. Or maybe just go with https://reactcommunity.org/react-transition-group/ ? Although I don't know how they interact with styled components.

Long story short. Ready to be reviewed. We can either merge then or provide some other small UI feedback.

feedm3 (approved):

feedm3:
PR looks good, I will merge it and create a new issue with the successful copy animation. When trying out the feature, this was also something I missed.

PR #142, opened 2021-09-27
fix: use semantically correct html for the language switch

feedm3:
I tried to fix our broken Google result ("Ende" page because of `en`/`de` language switch) with a more expressive HTML syntax. I'm not 100% sure if it works, but optimistic:
- added a `nav` around the whole switch
- put the links into a unordered list

**Screenshot of the bug** 
Screenshot 2021-09-27 at 08 45 50

- also, having `Satellytes` as RSS feed title is enough

Screenshot 2021-09-27 at 08 47 44

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -152,7 +152,7 @@ module.exports = {
               }
             `,
             output: RSS_FEED_URL,
-            title: 'Satellytes Blog Feed',
+            title: 'Satellytes',
             // optional configuration to insert feed reference in pages:
             // if `string` is used, it will be used to create RegExp and then test if pathname of
             // current page satisfied this regular expression;
--- a/src/components/language-switch/language-switch.tsx
+++ b/src/components/language-switch/language-switch.tsx
@@ -3,11 +3,6 @@ import React from 'react';
 import styled from 'styled-components';
 import { Link } from '../links/links';
 
-const UnstyledListItem = styled.li`
-  list-style-type: none;
-  display: inline;
-`;
-
 const StyledLanguageLink = styled(Link)<{ selected?: boolean }>`
   margin: 1px 0 1px 12px;
   font-weight: bold;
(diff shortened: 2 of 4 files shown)


gatsby-config.js
```diff
@@ -152,7 +152,7 @@ module.exports = {
               }
             `,
             output: RSS_FEED_URL,
-            title: 'Satellytes Blog Feed',
```
georgiee, line 155: fully agree 👍

src/components/language-switch/language-switch.tsx
```diff
@@ -18,26 +23,30 @@ export const LanguageSwitch = ({
...
-            }
-            language={languageOfLink}
-            selected={language === languageOfLink}
-          >
-            {languageOfLink}
-          </StyledLanguageLink>
-        );
-      })}
-    </div>
+    <nav aria-label={languageSwitchLabel} className={className}>
```
georgiee, line 25: It's a black box decision by google to pick this part so we don't know if this excludes anything though. 
But it's semantically correct and might help to mitigate the issue. Thanks for the change.

src/components/language-switch/language-switch.tsx
```diff
@@ -3,8 +3,13 @@ import React from 'react';
...
 import { Link } from '../links/links';
 
+const UnstyledListItem = styled.li`
+  list-style-type: none;
+  display: inline;
+`;
+
 const StyledLanguageLink = styled(Link)<{ selected?: boolean }>`
-  margin: 1px 0px 1px 12px;
+  margin: 1px 0 1px 12px;
```
georgiee, line 7: auto change or why did you do this? 0px is valid (as well it's the one without any unit).

src/components/language-switch/language-switch.tsx
```diff
@@ -18,26 +23,30 @@ export const LanguageSwitch = ({
...
-            language={languageOfLink}
-            selected={language === languageOfLink}
-          >
-            {languageOfLink}
-          </StyledLanguageLink>
-        );
-      })}
-    </div>
+    <nav aria-label={languageSwitchLabel} className={className}>
+      <ul>
```
georgiee: The nav element doesn't need the list. You could have swapped the div with the nav instead 🙏

src/components/language-switch/language-switch.tsx
```diff
@@ -18,26 +23,30 @@ export const LanguageSwitch = ({
...
+              <StyledLanguageLink
+                to={
+                  language !== languageOfLink && translation
+                    ? translation
+                    : originalPath
+                }
+                language={languageOfLink}
+                selected={language === languageOfLink}
+              >
+                {languageOfLink}
```
georgiee: The problem could be the `de` and `en`. I think we should provide at least a title with the actual language name. This will also help accessibility to spell the language instead of having some cryptic characters?

Something n`< a title="English" .. >`. Can you add this ?

georgiee (approved):
This is good. If you want to test it feel free to merge.
I have on request, to add the language titles to the actual links for accessibility reasons. Either add it to this PR or let's do it in a following one.

src/components/language-switch/language-switch.tsx
feedm3: I wasn't sure about the list. Will remove it.

src/components/language-switch/language-switch.tsx
feedm3: IntelliJ always shows this as warning, that's why I remove it when I see it 😄 

Screenshot 2021-09-27 at 12 08 20

src/components/language-switch/language-switch.tsx
feedm3: Yep, but in a separate PR

PR #143, opened 2021-09-27
chore: add title to languages

feedm3:
Language select now with appropriate title:

Screenshot 2021-09-27 at 12 31 14

### DIFF
--- a/src/locales/de/translations.json
+++ b/src/locales/de/translations.json
@@ -114,6 +114,8 @@
   "navigation": {
     "language-aria": "Sprachauswahl",
     "main-navigation-aria": "Hauptnavigation",
+    "en": "English",
+    "de": "Deutsch",
     "menu": "Menu",
     "imprint": "Impressum",
     "data-privacy": "Datenschutz",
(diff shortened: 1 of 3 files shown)


src/locales/de/translations.json
```diff
@@ -114,6 +114,8 @@
   "navigation": {
     "language-aria": "Sprachauswahl",
     "main-navigation-aria": "Hauptnavigation",
+    "en": "Englisch",
```
georgiee: I know this is the straight forward solution. Maintain a list of the languages for each available language. But I think we need to list the target language in it's own language. There is no point of translating `English` into the source language (here `Englisch`) because the user would not recognize the translated word anyway.

Similar to wikipedia where the language switch is always written in the target language:
Screen Shot 2021-09-27 at 12 55 05

```suggestion
    "en": "English",
```

src/locales/en/translations.json
```diff
@@ -114,6 +114,8 @@
   "navigation": {
     "language-aria": "Language Select",
     "main-navigation-aria": "Main Navigation",
+    "en": "English",
+    "de": "German",
```
georgiee:
```suggestion
    "de": "Deutsch",
```

georgiee (approved):
Thanks for the changes. I have added two suggestions for your convenience if you agree 🙏

src/locales/de/translations.json
feedm3: Nice catch, I changed it.

PR #149, opened 2021-10-18
Daniel interview

abdolg:
feat(blog): daniel interview


fischeversenker (approved):
Reads and looks great 🙌 
I really like Daniel's picture 📷 ✨

The one thing that's different about this post: So far all the blog posts have been in English while this one is in German. 
As far as I understand we don't plan to make the Blog translatable so we might need to think about translating this into English 🤔

abdolg:
> Reads and looks great 🙌 I really like Daniel's picture 📷 ✨
> 
> The one thing that's different about this post: So far all the blog posts have been in English while this one is in German. As far as I understand we don't plan to make the Blog translatable so we might need to think about translating this into English 🤔

I spoke to Georgios and he said that the interview in German is ok for now. For the future, he could imagine that a company would translate the texts. We can talk about this topic when Georgios is back.

PR #159, opened 2021-11-02
feat: add new header

### LINKED ISSUE #152: Component: Header
### Designs
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3655%3A25475

### Things to consider
- Can we reuse the existing Header component?
- Can we reuse the existing Aurora effect?
- The three icons in the top right corner don't need to have any functionality yet. The theme switcher might even be removed for now.

### Expected outcome
- We have a component called "Header" that is responsive and complies to the design.
- The company logo in the top left links to the home page.

hofmann-felix:
**Preview**
- https://satellytescommain21751-featheaderredesign.gtsb.io/

**What's included?**
- new language selection dropdown
- new icon colors
- new background & border color behavior 

**Screenshots**
[Transparent header with white icons and dark background](https://satellytescommain21751-featheaderredesign.gtsb.io/)
Bildschirmfoto 2021-11-09 um 08 28 12
[Transparent header with white icons and dark background (scrolled)](https://satellytescommain21751-featheaderredesign.gtsb.io/)
Bildschirmfoto 2021-11-09 um 08 29 27
[Transparent header with hero ](https://satellytescommain21751-featheaderredesign.gtsb.io/blog/interview-daniel-eissing/)
Bildschirmfoto 2021-11-09 um 09 00 15
[Header with hero (scrolled)](https://satellytescommain21751-featheaderredesign.gtsb.io/blog/interview-daniel-eissing/)
Bildschirmfoto 2021-11-09 um 09 01 05
[Header with white background and separation line](https://satellytescommain21751-featheaderredesign.gtsb.io/blog/)
Bildschirmfoto 2021-11-09 um 09 01 50

Corresponding ticket: https://github.com/satellytes/satellytes.com/issues/152

### DIFF
--- a/src/@types/styled-components.d.ts
+++ b/src/@types/styled-components.d.ts
@@ -38,6 +38,7 @@ declare module 'styled-components' {
         body: string;
         card: string;
         footer: string;
+        headerTransperent: string;
         leadbox: string;
       };
     };
--- a/src/@types/styled-components.d.ts
+++ b/src/@types/styled-components.d.ts
@@ -22,6 +22,7 @@ declare module 'styled-components' {
         secondary: string;
         header: string;
         headerLight: string;
+        headerHover: string;
         breadcrumb: string;
         link: {
           default: string;
--- a/src/@types/styled-components.d.ts
+++ b/src/@types/styled-components.d.ts
@@ -39,7 +39,7 @@ declare module 'styled-components' {
         body: string;
         card: string;
         footer: string;
-        headerTransperent: string;
+        headerTransparent: string;
         leadbox: string;
       };
     };
--- a/src/components/header/header.stories.tsx
+++ b/src/components/header/header.stories.tsx
@@ -15,10 +15,19 @@ const Template: ComponentStory<typeof Header> = (args) => <Header {...args} />;
 export const Regular = Template.bind({});
 Regular.args = {
   siteTitle: 'Satellytes',
+  $lightTheme: false,
+  showLanguageSwitch: true,
 };
 
-export const WithLanguageSwitch = Template.bind({});
-WithLanguageSwitch.args = {
+export const WithLightTheme = Template.bind({});
+WithLightTheme.args = {
   ...Regular.args,
-  showLanguageSwitch: true,
+  $lightTheme: true,
+};
+
+export const WithLightThemeTransparent = Template.bind({});
+WithLightThemeTransparent.args = {
+  ...Regular.args,
+  $lightTheme: true,
+  transparent: true,
 };
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -155,7 +155,10 @@ const Header: React.FC<HeaderProps> = (props) => {
       </SiteTitle>
       <Wrapper>
         {(props.translation || props.showLanguageSwitch) && (
-          <LanguageSwitch translation={props.translation} />
+          <LanguageSwitch
+            translation={props.translation}
+            $lightTheme={Boolean(!isHeaderTransparent && props.$lightTheme)}
+          />
         )}
         <SiteMenu
           aria-label="Open menu"
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -27,7 +27,10 @@ const StyledHeader = styled.header<{
       : props.$lightTheme
       ? props.theme.palette.background.bodyLight
       : props.theme.palette.background.body};
-
+  background: ${(props) =>
+    props.$transparent &&
+    props.$lightTheme &&
+    props.theme.palette.background.headerTransperent};
   height: ${HEADER_HEIGHT};
   display: flex;
   justify-content: space-between;
@@ -35,7 +38,9 @@ const StyledHeader = styled.header<{
   padding: 0 24px;
 
   border-bottom: ${(props) =>
-    props.$lightTheme ? '1px solid rgba(32, 40, 64, 0.05)' : 'none'};
+    props.$lightTheme && !props.$transparent
+      ? '1px solid rgba(32, 40, 64, 0.05)'
+      : 'none'};
 `;
 
 const HeaderSwoosh = styled(Swoosh)`
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -59,10 +59,17 @@ const SiteTitle = styled(Link)<{ $lightTheme: boolean }>`
   font-size: 20px;
   font-weight: bold;
   text-decoration: none;
+  transition: color 0.2s;
+
   color: ${(props) =>
     props.$lightTheme
       ? props.theme.palette.text.headerLight
       : props.theme.palette.text.header};
+
+  &:hover {
+    color: ${(props) =>
+      !props.$lightTheme && props.theme.palette.text.headerHover};
+  }
 `;
 
 const SiteMenu = styled.button<{ $lightTheme: boolean }>`
@@ -96,6 +103,13 @@ const SiteMenu = styled.button<{ $lightTheme: boolean }>`
       props.$lightTheme
         ? props.theme.palette.text.headerLight
         : props.theme.palette.text.header};
+  }
+
+  &:hover {
+    .bar {
+      background-color: ${(props) =>
+        !props.$lightTheme && props.theme.palette.text.headerHover};
+    }
   }
 `;
 
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -83,7 +83,7 @@ const SiteMenu = styled.button<{ $lightTheme: boolean }>`
   .bar {
     background-color: ${(props) =>
       props.$lightTheme
-        ? props.theme.palette.text.headerLight
+        ? props.theme.palette.text.default
         : props.theme.palette.text.header};
   }
 
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -24,7 +24,7 @@ const StyledHeader = styled.header<{
   background: ${(props) =>
     props.$transparent
       ? props.$lightTheme
-        ? props.theme.palette.background.headerTransperent
+        ? props.theme.palette.background.headerTransparent
         : 'none'
       : props.$lightTheme
       ? props.theme.palette.background.bodyLight
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -19,7 +19,7 @@ const StyledHeader = styled.header<{
   top: 0;
   width: 100%;
   z-index: 100;
-  transition: background-color 0.2s;
+  transition: background 0.2s;
 
   background-color: ${(props) =>
     props.$transparent
@@ -32,16 +32,10 @@ const StyledHeader = styled.header<{
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 0 16px;
+  padding: 0 24px;
 
   border-bottom: ${(props) =>
-    props.$lightTheme
-      ? '1px solid rgba(32, 40, 64, 0.05)'
-      : '1px solid rgba(255, 255, 255, 0.1)'};
-
-  ${up('md')} {
-    padding: 0 24px;
-  }
+    props.$lightTheme ? '1px solid rgba(32, 40, 64, 0.05)' : 'none'};
 `;
 
 const HeaderSwoosh = styled(Swoosh)`
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -21,16 +21,15 @@ const StyledHeader = styled.header<{
   z-index: 100;
   transition: background 0.2s;
 
-  background-color: ${(props) =>
+  background: ${(props) =>
     props.$transparent
-      ? 'none'
+      ? props.$lightTheme
+        ? props.theme.palette.background.headerTransperent
+        : 'none'
       : props.$lightTheme
       ? props.theme.palette.background.bodyLight
       : props.theme.palette.background.body};
-  background: ${(props) =>
-    props.$transparent &&
-    props.$lightTheme &&
-    props.theme.palette.background.headerTransperent};
+
   height: ${HEADER_HEIGHT};
   display: flex;
   justify-content: space-between;
--- a/src/components/language-switch/language-switch.tsx
+++ b/src/components/language-switch/language-switch.tsx
@@ -7,22 +7,34 @@ interface LanguageSwitchProps {
   translation: any;
   className?: string;
   $lightTheme?: boolean;
+  fromNavigation?: boolean;
 }
-const StyledSelection = styled.select<{ $lightTheme: boolean }>`
+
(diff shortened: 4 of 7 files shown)


src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,91 @@
...
+      : props.theme.palette.text.header};
+  &:hover {
+    color: ${(props) =>
+      props.fromNavigation
+        ? props.theme.palette.text.header
+        : !props.$lightTheme &&
+          !props.fromNavigation &&
+          props.theme.palette.text.headerHover};
+  }
+`;
```
georgiee: I know that's styled component but it's a mess to read and understand 😬
Do we have better patterns to handle a "complex" data structure like this? Like "calculated" theme values? You could use local methods in the tsx file for that purpose. This will make it much, much more readable.

This is a suggestion which you can implement within this PR or afterwards. But I would really like to see this addressed. Even if we did it like that in the past, I think we should find a better pattern. I hope you agree.

Maybe @fischeversenker or @feedm3 have some opinion in that too?

```
/**
Idea: calculate the theme colors outside of the style definition. Do not use ternary operator.
If possible comment if it's not clear what's happening too.
**/
const getTextColor(props) => {
   if(props.fromNavigation || props.$lightTheme) {
      return props.theme.palette.text.default
   }else {
      return props.theme.palette.text.header
   }
}


const getTextHoverColor(props) => {
   // ....
}


const StyledNav = styled.nav
// ....
color: ${(props) => getTextColor(props)
&:hover {
    color: ${(props) => getTextHoverColor(props)
```

src/components/header/header.tsx
```diff
@@ -99,6 +104,13 @@ const SiteMenu = styled.button<{ $lightTheme: boolean }>`
         ? props.theme.palette.text.headerLight
         : props.theme.palette.text.header};
   }
+
+  &:hover {
+    .bar {
+      background-color: ${(props) =>
+        !props.$lightTheme && props.theme.palette.text.headerHover};
+    }
+  }
```
georgiee, line 113: that makes such a difference for the menu icon animation 💫

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,91 @@
...
+
+const StyledSelection = styled.select<{
```
georgiee: I like the select/dropdown but it shows only `de` and `en` while we have no the space to write `Deutsch` and `English` and such (which was a problem before).

Can you change the labels in the dropdown? I think we should have the full names laying around somewhere already? 
Screen Shot 2021-11-08 at 10 33 22

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,91 @@
...
-  text-transform: uppercase;
+`;
 
-  ${({ selected }) => (selected ? `color: #668CFF;` : `color: #FFFFFF;`)}
+export const StyledChevron = styled(Chevron)<{
+  $lightTheme?: boolean;
+  fromNavigation?: boolean;
+}>`
+  margin-bottom: 2px;
+  /*margin-right: -6px makes the Chevron clickable*/
```
georgiee, line 71: that's because you pull it into the bounding box of the select. I now understand but the comment doesn't help. I also see that the magic number of 6px equals the width of the chevron. It works, and it's okay for now. 

I would have tried to use a `<label>` which can include the icon, some (visual hidden) text for accessibility reasons. Then the label can refer to the select through `label for` or such? Does this work here? It's a very small detail and I only want to give general feedback about those small things. No need to fix right now.

Screen Shot 2021-11-08 at 10 35 23

src/components/navigation/navigation.tsx
```diff
@@ -268,6 +268,7 @@ const Navigation: React.FC<NavigationProps> = ({
                 <LanguageSwitchWrapper
                   translation={translation}
                   className={'language-switch'}
+                  fromNavigation
```
georgiee, line 271: kind of awkward to have the language selection in the navigation only when it's triggered by the navigation and not when I scroll down (look at the footer, where the language select is appearin/disappearing when I click). Is this the intention of `fromNavigation` ? (I couldn't find anything else related that would match this behavior). Thanks for your clarification.

https://user-images.githubusercontent.com/1701755/140721057-fab95474-e3ad-437f-8f4b-c173cceec8ae.mp4

src/@types/styled-components.d.ts
```diff
@@ -37,6 +38,7 @@ declare module 'styled-components' {
         body: string;
         card: string;
         footer: string;
+        headerTransperent: string;
```
georgiee: typo
```suggestion
        headerTransparent: string;
```

georgiee (changes requested):
Thanks for the changes.
I would help tremendously if you drop some screenshots of the changes.
A note like `new icon colors` + the link to the issue helps but it doesn't tell me where to look at and makes the review pretty cumbersome. Do you think you can help me with some more details in the PR description?

Besides that please see my comments to improve the code quality (more readable styled components), accessibility (full language names in the dropdon) and consistency (navigation in the footer vs toggled by the header)

Thank you for you efforts so far 🌟 🙌

src/components/language-switch/language-switch.tsx
hofmann-felix: I agree with you, I didn't like that part during development either, but I couldn't think of a better solution either. A local method is definitely better 👍

feedm3: Agree. Make sure to use types that it's not guessing which props are available ;-)

src/components/language-switch/language-switch.tsx
hofmann-felix: Do you know if there is a pure HTML/CSS solution for displaying the abbreviation (e.g. "EN") when collapsed and the full text (e.g. "English") when expanded?

I talked briefly with Fabian about this and in our opinion, there is only the possibility to build a custom dropdown component to achieve this behavior.

src/components/navigation/navigation.tsx
hofmann-felix: `fromNavigation` only ensures that the LanguageSwitch gets the right color. The distinction of whether the NavigationSwitch is displayed or not is made in the actual navigation component (`showLanguageSwitch`) and depends on whether the navigation is displayed as a footer or as a flyout. 
I agree with you that the behavior is a bit unexpected. I can talk to Eric and ask him if it would make sense to always show the language switch in the navigation.

src/components/language-switch/language-switch.tsx
fischeversenker: This is not possible with the native `select` element. I'm pretty sure this would require a custom solution. I'd say that this is out of scope of this PR. I suggest we iterate on the LanguageSwitcher at a later point and merge this with the simple native `de/en` dropdown.

georgiee: out of scope, I agree. I will remove this from my list of expectations then.

src/components/navigation/navigation.tsx
georgiee: If we don't find a solution for this I'm happy to tackle this later so you can move forward with the PR 👍

georgiee:
I see you pushed some updates. Once the PR is ready or updated please notify us. Then I can update my review status. Thanks!

src/components/header/header.stories.tsx
```diff
@@ -15,10 +15,19 @@ const Template: ComponentStory<typeof Header> = (args) => <Header {...args} />;
...
   ...Regular.args,
-  showLanguageSwitch: true,
+  $lightTheme: true,
+};
+
+export const WithLightThemeTransparent = Template.bind({});
+WithLightThemeTransparent.args = {
+  ...Regular.args,
+  $lightTheme: true,
+  transparent: true,
```
fischeversenker, line 32: In the PR where @georgiee [introduced the story for the Navigation component](https://github.com/satellytes/satellytes.com/pull/180) he questioned the "many stories for simple properties" approach. He proposed to have a single story with a lot of controls for simple thins like "with or without language switch". I like that idea and wanted to suggest to merge all these stories into one... but it's really not easy to decide which property deserves its own story and which doesn't. I'm fine to leave this as it is since these three stories now show three very different appearances of the same component. When we switch to MDX in #185 we might be able to come up with a more intuitive way of defining stories. No todo here... just raising awareness for "what is worth writing a story?".

Another thing: While playing around with this I noticed that the `transparent` property behaves strangely. It can't be turned off. Maybe this is a storybook bug but this could also hint at some fishy property-logic in the component. It seems like `props.transparent` is only taken into account during the initial mounting of the component but later changes are ignored. This is nothing that you introduced so no need to fix it in this PR. Maybe you want to have a look anyways?

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,115 @@
...
+  color: ${(props) => getTextColor(props)};
+  &:hover {
+    color: ${(props) => getTextHoverColor(props)};
+  }
+`;
+
+const StyledSelection = styled.select<{
+  $lightTheme?: boolean;
+  fromNavigation?: boolean;
+}>`
```
fischeversenker: This styled component does not make use of any of the props so this generic type is dead code. I suggest to remove it:
```suggestion
const StyledSelection = styled.select`
```

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,115 @@
...
+  padding-left: 10px;
   cursor: pointer;
-  text-transform: uppercase;
+`;
 
-  ${({ selected }) => (selected ? `color: #668CFF;` : `color: #FFFFFF;`)}
+export const StyledChevron = styled(Chevron)<{
+  $lightTheme?: boolean;
+  fromNavigation?: boolean;
+}>`
```
fischeversenker: The same here. No need for the generic type as far as I can tell:
```suggestion
export const StyledChevron = styled(Chevron)`
```

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,115 @@
...
+        fromNavigation={fromNavigation}
+      />
+      <StyledSelection
+        onChange={(event) => {
+          changeLanguage(event.target.value);
+        }}
+        value={language}
+        $lightTheme={$lightTheme}
+        fromNavigation={fromNavigation}
+      >
```
fischeversenker: If you agree to remove the generic types some of these props would also need to be removed. Less code means less bugs so I still suggest to do so.
```suggestion
      <StyledChevron />
      <StyledSelection
        onChange={(event) => {
          changeLanguage(event.target.value);
        }}
        value={language}
      >
```

fischeversenker (changes requested):
Thanks for the updates 🙌 

I left a few comments. Especially the ones regarding "dead code" are ones that I'd like to see resolved before we merge this. Please take a look :pray:

fischeversenker (approved):
Thanks for removing these unused types :+1:
The one leftover open discussion is for you to decide. It's perfectly fine to leave it like it is if you want.

Another thing that I noticed: We might want to remove the `position: fixed` from this component and make it a property of a `HeaderWrapper` inside the `Layout` component to make it easier to write stories. This could be part of another PR, though. No todo right now.

This is good to go now in my opinion.

src/components/header/header.stories.tsx
hofmann-felix: Thanks for your review 👍
I have the same problems with the transparent property in the storybook. I did a bit of research and tried a few things (e.g. define custom argTypes), however, it still didn't work.
That's why I would also suggest leaving the three stories as they are, so you can see all styles. In general the approach "fewer stories with many controls" sounds good. I will include it in other PRs

fischeversenker: This is not related to Storybook. I had a look and figured out it's because of the way the initial state of `isHeaderTransparent` is set up. It takes the initial value of `props.transparent` during the mounting and from then on ignores any value changes of `props.transparent`. Let's keep it as it is and maybe work on this in the future. It's not going to be a problem on the actual website.

georgiee: > it seems like props.transparent is only taken into account during the initial mounting of the component but later changes are ignored. This is nothing that you introduced so no need to fix it in this PR. Maybe you want to have a look anyways?

I have checked the header component, it's missing an `useEffect` to update the property. It's counter intuitive but the default value for useState is not updated on future renderings. That's `useEffect` for, where you can tell what is going to change and then trigger the state update.

With the following code in header it will work. Please provide a fix after merging this PR 👍

```
const transparent = Boolean(props.transparent);

useEffect(() => {
    setIsHeaderTransparent(transparent)
  }, [transparent])
```

src/components/header/header.tsx
```diff
@@ -32,16 +34,12 @@ const StyledHeader = styled.header<{
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 0 16px;
+  padding: 0 24px;
 
   border-bottom: ${(props) =>
-    props.$lightTheme
+    props.$lightTheme && !props.$transparent
```
georgiee, line 40: I don't understand the naming convention here with the leading dollar sign. It was there before for $lightTheme can someone please help explaining ?

src/components/language-switch/language-switch.tsx
```diff
@@ -1,44 +1,104 @@
...
+  fromNavigation?: boolean;
+  $lightTheme?: boolean;
+  theme: DefaultTheme;
+}) => {
+  if (fromNavigation) {
+    return theme.palette.text.header;
+  } else if (!$lightTheme) {
+    return theme.palette.text.headerHover;
+  }
+};
```
georgiee, line 43: thanks for making this more readable 👍

src/@types/styled-components.d.ts
```diff
@@ -38,6 +39,7 @@ declare module 'styled-components' {
         body: string;
         card: string;
         footer: string;
+        headerTransparent: string;
```
georgiee, line 42: We have a nested hierarchy in this theme file. In the future I think it makes sense to resolve camel case into the hierarchy. Instead of `headerTransparent` I would like to see `header:{ transparent: string }` for example.

georgiee (approved):
Hello,
I think it's good enough so we have a story and a slightly updated style. Thanks for your efforts 👍 I have also checked the gatsby page itself and the header still looks good including the updates.

Please consider working on the following aspects in following PRs & discussions:

+ Fix the props initialization with `useEffect` and friends as reported
+ I would like to see the naming of props with leading dollar signs `$lightTheme` clarified or updated
+ The point from Felix is good: I would like to remove the position aspect from the component and instead providing a support Layout helper component like `HeaderPositioningLayout` to contain the fixed position and such (and no need for a story then)
+ I didn't see any discussion related to `isNavigationVisible` which is not covered in the story. Any plans on that?

Anyway I would like to move forward and merge this to have a first story in place. Let's merge! Approved🌟

PR #160, opened 2021-11-04
chore: introduce storybook

### LINKED ISSUE #156: Make it easier to develop single components
I suggest to add Storybook (or some other UI comp tool) to make it easier to develop components. This will not only help during the current migration but also serve as living documentation for future developers.
There is a gatbsy plugin that hopefully makes this really easy. Let's investigate 🙌 

### Resources
Documentation: https://www.gatsbyjs.com/docs/how-to/testing/visual-testing-with-storybook/

### Expected outcome
A UI where I can view, test, and develop components.

fischeversenker:
This PR introduces `storybook` to more easily develop and inspect the single components that make up our website. I included an example "story" for the `Header`. I hope that we can add stories for other components during the ongoing redesign.

### How to review
```sh
# install new dependencies
yarn

# run storybook
yarn storybook
```

closes #156


package.json
```diff
@@ -119,6 +130,7 @@
     ]
   },
   "resolutions": {
-    "cheerio": "1.0.0-rc.10"
+    "cheerio": "1.0.0-rc.10",
+    "webpack": "^5"
```
georgiee, line 134: what's the conflict to override the webpack dependencies?

package.json
```diff
@@ -109,6 +119,7 @@
     "jsdom": "^18.0.0",
     "lint-staged": "11.2.4",
     "prettier": "2.4.1",
+    "storybook-addon-gatsby": "^0.0.2",
```
georgiee, line 122: I found the [GH repo] https://github.com/prismicio-community/storybook-addon-gatsby but I wasn't much smarter then from the readme. By looking at the code I see it's about patching the webpack configuration.

> This modifies the Webpack configuration to simulate Gatsby's internal Webpack configuration. It is derived from Gatsby's Visual Testing with Storybook guide and a collective effort of Gatsby users.

It's a contribution from the [prismic](https://prismic.io/docs) folks which have quite some knowledge on gatsby so I trust this to be good enough for our small case 🤓🙌

package.json
```diff
@@ -86,6 +88,13 @@
     "what-input": "^5.2.10"
   },
   "devDependencies": {
+    "@babel/core": "~7.15.8",
+    "@storybook/addon-actions": "^6.3.12",
+    "@storybook/addon-essentials": "^6.3.12",
+    "@storybook/addon-links": "^6.3.12",
+    "@storybook/builder-webpack5": "^6.3.12",
+    "@storybook/manager-webpack5": "^6.3.12",
+    "@storybook/react": "^6.3.12",
```
georgiee, line 97: It's such a shame that there is no other format for `package.json` like `package.yaml` or so,
so we can get comments (json will never support comments). In order to fully appreciate your PR I need to look up all the different packages while a single comment directly at the package would explain so much more. 

The plugins look like the defaults one, nothing special besides the webpack 5 thing required for gatsby/webpack ?

package.json
```diff
@@ -24,7 +24,9 @@
     "test:lint": "eslint \"**/*.{tsx,ts}\"",
     "test:format": "prettier \"**/*.{tsx,ts,md}\" --check",
     "postinstall": "husky install",
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "storybook": "start-storybook -p 6006 -s ./",
```
georgiee, line 28: simple & beautiful. Thank you for the addition.

src/components/header/header.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default {
+  component: Header,
+  title: 'Components/Header',
+  parameters: {
+    layout: 'fullscreen',
+  },
+} as ComponentMeta<typeof Header>;
+
+const Template: ComponentStory<typeof Header> = (args) => <Header {...args} />;
```
georgiee, line 13: the header is a great starter👍

georgiee (approved):
Thank you @fischeversenker for the great addition. I think this really helps to make the current set of componnets more accessible for existing and new developers. Would be nice, if we can deploy them too in the future, but it's absolutely secondary.

I also love how a few changes make it possible to use storybook 🥳 Let's merge this an then grow the content from here on.

package.json
fischeversenker: Storybook by default currently ships with a dependency to webpack 4. This led to the two major versions 4 and 5 being installed. For some reason yarn decided to hoist webpack 4 as the default one which obviously led to build errors in Gatsby since it relies on webpack 5. With the opt-in into the webpack 5 builder for storybook and this resolution I was able to deal with this.

package.json
fischeversenker: This addon was recommended in the [official Gatsby documentation](https://www.gatsbyjs.com/docs/how-to/testing/visual-testing-with-storybook/#using-an-addon). I noticed that the addon mentions Gatbsy^3 as a peer dependency which made me skeptical at first, but it works fine so I decided to stick with it.

package.json
fischeversenker: For adding `storybook` initially, I ran [`npx sb init`](https://storybook.js.org/docs/react/get-started/install) which took care of all the heavy lifting. It figured out that this was a React project and added most of these dependencies, added the scripts in the `packge.json` and created an initial set of configurations. Really nice DX 🙌

package.json
fischeversenker: I'm forwarding this "Thank you" to the great `npx sb init` command 🙏

fischeversenker:
Thanks for the review @georgiee 🙏 
I'll merge this, now 🚂 
Every time we touch/add a component in the future we can then create a Storybook story or improve an existing one 🙌

PR #163, opened 2021-11-08
feat: add breadcrumb

### LINKED ISSUE #153: Component: Breadcrumbs
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3655%3A25498

### Things to consider
- Currently there is no Breadcrumb component. Is there still something that we can reuse?
- There is a plugin that we might want to use: https://www.gatsbyjs.com/plugins/gatsby-plugin-breadcrumb/

### Expected outcome
A component called "Breadcrumbs" that complies to the design and allows to properly navigate up the hierarchy.

hofmann-felix:
**URL**
https://satellytescommain21751-featbreadcrumb.gtsb.io/

**What's included?**
- breadcrumb navigation in blog, blogpost and job advertisement
- Uses https://www.gatsbyjs.com/plugins/gatsby-plugin-breadcrumb/

**Screenshots**
[Breadcrumb in blogpost](https://satellytescommain21751-featbreadcrumb.gtsb.io/blog/howto-blog-post/)
Bildschirmfoto 2021-11-09 um 09 41 36
[Breadcrumb in job advertisement](https://satellytescommain21751-featbreadcrumb.gtsb.io/career/senior-backend-engineer/) (background will be white in the new design)
Bildschirmfoto 2021-11-09 um 09 42 09

Resolves: https://github.com/satellytes/satellytes.com/issues/153


src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+    text-overflow: ellipsis;
+  }
+`;
+interface BreadcrumbProps {
+  breadcrumbEntries: BreadcrumbType[];
+  customLabel?: string;
+  hasHero: boolean;
+}
+
+export interface BreadcrumbType {
```
feedm3: Name is to generic and should be more specific: `BreadcrumbEntry`

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+  pathname: string;
+  crumbLabel: string;
```
feedm3: Shouldn't `label` be enough?

package.json
```diff
@@ -41,6 +41,7 @@
     "gatsby": "4.1.0",
     "gatsby-background-image": "^1.5.3",
     "gatsby-image": "^3.11.0",
+    "gatsby-plugin-breadcrumb": "^12.2.0",
```
feedm3: I have mixed feelings about this plugin:
- it only has 51 stars on Github, so not many people seem to use it
- it adds black-box behavior to our build. I don't understand where the breadcrumbs come from or how they are generated. Breadcrumbs seem to be a very easy concept: An array of links that is shown as list. Wouldn't it be simpler to just manually add a `breadcrumbs` array to the `pageContext` of each page?

feedm3 (changes requested):

package.json
fischeversenker: Not sure about the 51 stars and how much that says about the quality of the plugin. On gatsbyjs.com it has 12.9k downloads (not sure whether this is overall or yearly or monthly or hourly...). It seems to be actively maintained and apparently works well for us, so why not use it?

Generally I do very much like your second point, asking whether we can do the same without using such an external black-box dependency. Looking at the code from that plugin I'm not sure if it's actually that easy, though. If I understand you correctly you suggest to basically hardcode the "array of links" to each page... I'd like to avoid that if we can instead have such a plugin to take care of it. What are the downsides?

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+import styled from 'styled-components';
+import { HEADER_HEIGHT } from '../header/header';
+import { Link } from '../links/links';
+import { useI18next } from 'gatsby-plugin-react-i18next';
+import { BreadcrumbText } from '../typography/typography';
+import { theme } from '../layout/theme';
+
+const BreadcrumbContainer = styled.ol<{ hasHero: boolean }>`
+  margin: calc(${HEADER_HEIGHT} + 16px) 24px 0;
+  margin-top: ${(props) => props.hasHero && '16px'};
```
fischeversenker: I feel like these are values that the breadcrumb component does not need to know. Can't we have the breadcrumb just as a simple UI element that is being placed somewhere on the page? The spacing from the top is something that a parent component (`Layout`?) should take care of.

src/components/breadcrumb/breadcrumb.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  component: Breadcrumb,
+  title: 'Components/Breadcrumb',
+  decorators: [
+    (Story) => (
+      <>
+        <GlobalStyle $lightTheme={true} />
+        <Story />
+      </>
+    ),
+  ],
```
fischeversenker: Is this needed? The general storybook setup already pretty much adds the same decorator. Do we need that explicit `$lightTheme` input? We might as well add it to the global decorator in `.storybook/preview.js` since we are currently working only on the light theme.

src/components/breadcrumb/breadcrumb.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+WithHero.decorators = [
+  (Story) => (
+    <>
+      {CustomHeader(true)}
+      <img style={{ width: '100%' }} alt={''} src={AngularIMG} />
+      <Story />
+    </>
+  ),
+];
```
fischeversenker: All these stories are needed because the Breadcrumb knows too much. If we treat the Breadcrumb component as a simple "dumb" component that knows nothing about where it's being used then we can drastically reduce the complexity in these stories. I think we should aim for as-simple-as-possible components that we can then use to assemble a nice UI. As written in my other comment I strongly recommend to remove the header-height and hero knowledge/dependencies from this component.

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {breadcrumbEntries.map((breadcrumbEntry, index) => {
+        if (index === 0 && !isDefaultLanguage) {
+          return;
+        }
+        const crumbLabel: string =
+          (index === breadcrumbEntries.length - 1 && customLabel) ||
+          t(`${breadcrumbEntry.crumbLabel}.breadcrumb`);
+        return (
+          <BreadcrumbListItem key={crumbLabel}>
+            <Link to={breadcrumbEntry.pathname} language={defaultLanguage}>
```
fischeversenker: Is the `defaultLanguage` on purpose?
```suggestion
            <Link to={breadcrumbEntry.pathname} language={language}>
```

fischeversenker (changes requested):
Thanks for this new component, @hofmann-felix!

I'd like to understand if it needs to know that much about its surroundings, though. In my imagination this component was a lot simpler, without any knowledge of the header height or whether there is a hero. Does it really need to be this complex?

.storybook/preview.js
```diff
@@ -6,7 +6,7 @@ import { GlobalStyle } from '../src/components/layout/global-style';
 addDecorator((storyFn) => (
   <>
     <ThemeProvider theme={theme}>
-      <GlobalStyle />
+      <GlobalStyle $lightTheme />
```
fischeversenker: This still led to the dark theme when I tried it:
image

We need to explicitly set this to `true` to make it work.
```suggestion
      <GlobalStyle $lightTheme={true} />
```

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,64 @@
...
+      content: '›';
+      margin-left: 7px;
+      display: inline-flex;
+      flex: 0 0 auto;
+    }
+  }
+  :last-child {
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
```
fischeversenker, line 30: This is a very nice touch. Thanks for thinking about this 🙏 
image

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,64 @@
...
+  customLabel,
+}: BreadcrumbProps): JSX.Element => {
+  const { t } = useTranslation();
+
+  return (
+    <BreadcrumbContainer>
+      {breadcrumbEntries.map((breadcrumbEntry, index) => {
+        const crumbLabel: string =
+          (index === breadcrumbEntries.length - 1 && customLabel) ||
+          t(`${breadcrumbEntry.label}.breadcrumb`);
```
fischeversenker: I would suggest to pass in strings that are already translated instead of implicitly doing the translation here in the component. I was surprised to find the `.breadcrumb` added to the strings given in Storybook. I wouldn't expect a property called `label` to be used as the i18n key.
Is there a benefit of doing it this way?

fischeversenker (commented):
Thanks for the updates, @hofmann-felix. The component looks a lot cleaner now :sparkles:
Also I approve of removing the breadcrumb-plugin. Hardcoding the breadcrumbs looks perfectly reasonable for now. Thanks for taking care of this.

The `lightTheme` attribute needs some more love and I feel like there is no benefit of doing the translation inside the component so we might want to think about moving it to outside of the component. Please have a look at my comments.

src/components/breadcrumb/breadcrumb.tsx
hofmann-felix: Thanks for your feedback 🙏
I have now removed the translation logic and it will directly give the translated string to the breadcrumb component. I am not sure if the solution is ok, especially the part in `create-carrer-details.js`.

gatsby/create-pages/create-carrer-details.js
```diff
@@ -176,6 +184,7 @@ const createCareerDetails = async ({ actions }) => {
               socialCardImage: publicUrl,
               language: languageKey,
               translation: getTranslation(position.id),
+              breadcrumb: breadcrumb,
```
fischeversenker: There is no need to generate the breadcrumbs here in this file and pass them to the template. We can move all this into the template and use `t(...)` to receive the translations. See my other comment in the template for details.

gatsby/create-pages/create-carrer-details.js
```diff
@@ -192,6 +201,7 @@ const createCareerDetails = async ({ actions }) => {
           context: {
             positions: jobPositions,
             language: languageKey,
+            breadcrumb: breadcrumb,
```
fischeversenker: The `/career/` page doesn't have a breadcrumb so we can remove this.

src/components/breadcrumb/breadcrumb.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+);
+
+export const Regular = Template.bind({});
+Regular.args = {
+  breadcrumbEntries: [
+    { pathname: '/', label: 'Main' },
+    { pathname: '/blog', label: 'Blog' },
+    { pathname: '/blog/howto-blog-post/', label: 'How to do a blog post' },
+  ],
+};
```
fischeversenker, line 21: This get's nicer and nicer with every change. Very clean story now :sparkles:
We don't even need the `parameters: { layout: 'fullscreen' }` in my opinion... but it's also fine to leave it.

src/pages/blog.tsx
```diff
@@ -58,14 +58,18 @@ const BlogPage = ({ data, location }: BlogPageProps) => {
   const { t } = useTranslation();
   const blogPosts = data.allMarkdownRemark.nodes;
   const { language, changeLanguage } = useI18next();
+  const breadcrumb = [
+    { pathname: '/', label: 'Satellytes' },
+    { pathname: '/blog', label: 'Blog' },
```
fischeversenker: To be completely future-proof we could make use of the translations here:
```suggestion
    { pathname: '/blog', label: t('navigation.blog') },
```

I do realize that the `SEO` element on line 73 receives this same string hardcoded but I consider it a best practice to make everything translatable (except maybe "Satellytes"). Maybe we will add a Mongolian version in the future and they have a special word for "Blog"?

src/templates/blog-post.tsx
```diff
@@ -91,8 +95,13 @@ const BlogHeader = ({ readingTime, frontmatter }) => {
 const BlogArticleTemplate: React.FC<BlogArticleTemplateProps> = ({
   data,
   location,
+  pageContext,
 }) => {
   const markdown = data.markdownRemark;
+  const breadcrumb = [
+    ...pageContext.breadcrumb,
+    { pathname: location.pathname, label: markdown.frontmatter.title },
+  ];
```
fischeversenker: What's the benefit of having parts of the breadcrumb in the page context? I suggest to move the breadcrumb "calculation" completely to the template and remove it from the `create-pages` script. I also recommend to use `t(...)` wherever possible (in this case for "Blog"), see other comment for my reasoning.
```suggestion
  const breadcrumb = [
    { pathname: '/', label: 'Satellytes' },
    { pathname: '/blog', label: t('navigation.blog') },
    { pathname: location.pathname, label: markdown.frontmatter.title },
  ];
```

(Disclaimer: `t` is not yet available in this context so if you accept this suggestion please make sure to also get it with `useTranslation()`.)

src/templates/career-details.tsx
```diff
@@ -83,6 +88,13 @@ const CareerPage: React.FC<CareerPageProps> = (props): JSX.Element => {
   const { t } = useTranslation();
 
   const ref = React.useRef<HTMLDivElement | null>(null);
+  const breadcrumb = [
+    ...pageContext.breadcrumb,
+    {
+      pathname: pageContext.i18n.originalPath,
+      label: pageContext.position.name,
+    },
+  ];
```
fischeversenker: The same goes for this template. I don't see the benefit of having parts of the breadcrumb in the page context. I suggest to remove the `BREADCRUMB_*` bit from the `create-pages` script and instead hardcode them here - including the usage of `t(...)`:
```suggestion
  const breadcrumb = [
    { pathname: '/', label: 'Satellytes' },
    { pathname: '/career', label: t('navigation.career') },
    {
      pathname: pageContext.i18n.originalPath,
      label: pageContext.position.name,
    },
  ];
```

fischeversenker (commented):
Thanks again for the changes. The story looks really neat and tidy now. I like it 🤩

Your gut feeling about the breadcrumb translations in the career pages is right. That's not what I imagined. Let's try to improve this. I left a bunch of comments regarding this. Was there any reason why you added them to the `create-pages` scripts instead of adding them to the templates directly? I suggest to leave the `create-pages` untouched and instead move everything into the templates. And to also make use of the `t(...)` function to have everything translated. Please have a look at my comments.

src/templates/blog-post.tsx
hofmann-felix: I wrote the breadcrumbs in the page context because I wanted to keep the structure where all the data comes from the create-pages files. 
But I agree with you and I think that the advantage of using `t()` is much bigger and that in other places in the code the data doesn't always come from the create-pages files either. I have now adapted the code according to your review and it looks better than before. Thanks for your feedback 😊

src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Link } from '../links/links';
+import { BreadcrumbText } from '../typography/typography';
+import { theme } from '../layout/theme';
+
+const BreadcrumbContainer = styled.ol`
+  margin: 16px 24px 0;
```
fischeversenker: I'm not convinced we should keep this "outer spacing" in this component. There are different opinions about this and there is no right or wrong. Personally I feel like it makes everything easier when we stick to the rule of "components don't define any outside spacing". See [this rather long article about component spacing](https://medium.com/fed-or-dead/handling-spacing-in-a-ui-component-library-70f3b22ec89) for an explanation of why this is a difficult topic.

In this particular case I suggest we get rid of at least the horizontal spacing. Personally I'd like to get rid of the margin completely but I'm happy to hear some more opinions on this.

fischeversenker (approved):
Thanks a lot for your patience and the continuous updates, @hofmann-felix 🙏

I noticed that there are no hover effects on the breadcrumb links but it looks like this is something that the `Link` component is lacking and thus not actually related to this PR.

The only thing that is left that I suggest to change is the "outer spacing" of the Breadcrumb component. See my comment for details. This does not block this PR anymore so I'm finally approving.

fischeversenker (approved):
Thanks for the latest changes. This is a very clean and reusable component now 🎉

package.json
feedm3: Better late than never: The plugin does a lot of things we don't need (like click tracking to get the breadcrumb path). What I don't like is when a plugin adds something to the page context. This makes it later very hard to understand where those page context variables come from, as you just see them popping up in the templates. Also, this plugin doesn't work for all pages but just for some pages we have (as the career pages are added via the Personio API). 

In general, I prefer to have all link logic either in `gatsby-node`, where you can see how links are generated. Or directly from an API. This way you have a better overview of how those components get data.

feedm3:
This pull request looks very good now. Thanks both Felix for implementation! 👯

feedm3 (approved):

Issue #154, opened 2021-10-26
Component: Navigation/Footer


fischeversenker:
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3655%3A25488

### Things to consider
- Ca we reuse the existing `Navigation`?
- The menu is expected to be [animated](https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=122%3A1191) when "opened" via the hamburger button. I suggest to add this feature at a later point.

### Expected outcome
A component called `Navigation` that complies to the design.

georgiee:
I will examine the existing navigation. Ideally it's already shared across the different places so I can provide a story and necessary refactorings if any.

georgiee:
FYI This will need the introduction of the i18next plugin for storybook to provide our translations:
https://storybook.js.org/addons/storybook-react-i18next

We need to align the follow configuration with the guide from storybook: (like a different namespace etc)
https://github.com/microapps/gatsby-plugin-react-i18next/blob/e507e23157c3a6d66bfa6f7e0acd08f5f656ff67/src/plugin/wrapPageElement.tsx#L128-L138

georgiee:
I renamed this issue to `Navigation/Footer` as we have a combined component (kind of). We can now use this place to coordinate our efforts. I will move forward and try to solve the edge case when people open the navigation being at the bottom already.

georgiee:
we have worked on that let's plan through #254 where to continue.

PR #180, opened 2021-11-11
chore: add story for the navigation component

georgiee:
Relates to #154 

---

## Description
This change introduces a story for `Navigation` by using the existing component which is shared probably between the footer and the header. The main complexity coming with this PR is the introduction of `i18next` inside the storybook so we have access to our translations. This was kind of tricky as the official [README](https://github.com/stevensacks/storybook-react-i18next) of the storybook plugin uses `addResources` to load the translations while I have figured out that we need `addResourceBundle` to do so — that was super hard because I assumed to read best practices. I'm not sure why this is so and I opened a [PR](https://github.com/stevensacks/storybook-react-i18next/pull/3) to propose a change and to help some other lost souls. 

## Config Issues
As we are using a gatsby plugin which pre-configures a few things I had to lookup some defaults to make sure both worlds can handle our translation files equally. Here the [gatsby plugin defaults](https://github.com/microapps/gatsby-plugin-react-i18next/blob/e507e23157c3a6d66bfa6f7e0acd08f5f656ff67/src/plugin/wrapPageElement.tsx#L128-L138) and the [i18next](https://www.i18next.com/overview/configuration-options) option list to understand what I'm doing.

The default namespace from i18next is actually `translation` so I omit everything related to that namespace, while the gatsby plugin provides it explicitly. The only addition worth to mentions is the `interpolation.escapeValue` property which I copy over from our gatsby config to prevent any surprises in the future.

## Preview
Here a preview image:
Screen Shot 2021-11-11 at 14 45 28-fullpage

We can also change the language:
https://user-images.githubusercontent.com/1701755/141312714-d2f75076-f022-4be7-b38f-037da836e646.mp4

## Discussion
In the header (initially by @fischeversenker) I have noticed the pattern to create a variant per flag. I think this creates an unnecessary split as storybook already gives us the proper toggles. Why is that and is it okay to have only variant like I did for this story ?

Screen Shot 2021-11-11 at 15 13 53

## Next
Once this PR is landed we can work on the actual design changes of the component.


fischeversenker (approved):
Thanks for this! Especially for integrating everything related to `i18n`.

> is it okay to have only variant like I did for this story ?

Absolutely. For the Leadbox it makes sense to have different stories for "with mail" or "with link" but for most simple boolean properties (like the one on the Header) I agree that we don't need a separate story.

georgiee:
Thanks, I'm going to merge this once green 🙏

PR #181, opened 2021-11-11
feat: add teaser component

### LINKED ISSUE #173: Component: Teaser
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3732%3A26923

### Things to consider
- Can we reuse the TextBlock and the Icon component for this?

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/173

### Whats's included?
- new Teaser component with a corresponding story
- different teaser styles from the [design](https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3732%3A26923) 

### Notes
- ~~the style for the illustrations (e.g. hover effect, width, ...) is not implemented, because there is no generic icon/illustration component yet.~~
- run `yarn storybook` to see the changes

### Preview 
_Regular:_
Bildschirmfoto 2021-11-15 um 10 38 36
_With illustration:_
Bildschirmfoto 2021-11-15 um 10 38 28
_With image:_
Bildschirmfoto 2021-11-15 um 10 38 49


src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+  margin-top: ${(props) => (props.hasImage ? '8px' : '16px')};
+
+  ${(props) =>
+    props.hideOverflow &&
+    css`
+      overflow: hidden;
+      text-overflow: ellipsis;
+      display: -webkit-box;
+      -webkit-line-clamp: 3;
+      -webkit-box-orient: vertical;
```
georgiee: there is one simple rule with CSS: Never, ever use prefixed properties. Either a post processor (which is available by default in gatsby with postCSS I think?) will fix your statement and make it compatible with browsers that only support prefixed content or you can't use it.

Browser support looks good https://caniuse.com/?search=line-clamp and I like the line based clamping. Please check and confirm if this is properly processed with PostCSS if you remove the prefix.

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+  children,
+}) => {
+  const largeTitle = !image && !icon;
+  return (
+    <TeaserContainer>
+      <Link to={linkTo}>
+        <Illustration>{icon}</Illustration>
+        {image && (
+          <ImageContainer>
+            <TeaserImage className="teaser-img" alt={''} image={image} />
```
georgiee: Ideally this should be wrapped by styled components. You don't use class names with styled components usually. Is this possible here or why did you pick this strategy ?

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const ToplineContainer = styled.div`
+  display: flex;
+  margin-top: 16px;
+  width: 100%;
+  justify-content: space-between;
+`;
+
+const StyledTeaserTitleLarge = styled(TeaserTitleLarge)<{
+  hideOverflow?: boolean;
```
georgiee: Is this something we want to be configurable or is it better to enforce an overflow in any case for the beginning? I would remove this variation for the beginning

src/components/teasers/teaser.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+        },
+        {
+          srcSet:
+            '/static/f09962349f7b6084a9508ccfe1993c25/f5103/space.webp 150w,\n/static/f09962349f7b6084a9508ccfe1993c25/dc425/space.webp 300w,\n/static/f09962349f7b6084a9508ccfe1993c25/7802b/space.webp 600w,\n/static/f09962349f7b6084a9508ccfe1993c25/2dc4e/space.webp 1200w',
+          type: 'image/webp',
+          sizes: '(min-width: 600px) 600px, 100vw',
+        },
+      ],
+    },
+    width: 600,
```
georgiee: This structure looks like you are piping through the parameter set of gatsby image? I would rather see to pass in the actual image component instead. Makes it much more lightweight. Similar hwo you pass the icon. This makes this component also much more flexible as it removes the gatsby dependency. 

```

const MyImage = <GatsbyImage.../>

<Teaser
  icon={ <SpaceShuttle /> }
  image={ <MyImage image={source}/> }
  linkTo="/blog"
  title="Leadbox Title"
>
  Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Morbi leo risus, porta ac consectetur ac, vestibulum at eros. Maecenas sed diam eget risus varius blandit sit amet non magna.
</Teaser>

```

No need to mimic the GatsbyImage Image API. What do you think ?

.storybook/preview.js
```diff
@@ -6,7 +6,7 @@ import { GlobalStyle } from '../src/components/layout/global-style';
 addDecorator((storyFn) => (
   <>
     <ThemeProvider theme={theme}>
-      <GlobalStyle />
+      <GlobalStyle $lightTheme={true} />
```
georgiee: How to access the dark theme then? There is a background toggle in storybook (dark & light), is there a way of toggling our theme in general? Because we will need to check our components on both backgrounds for sure.

What design is more correct right now to begin with? Light? Than this is fine, otherwise we need to rollback.

src/components/teasers/teaser.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export default {
+  component: Teaser,
+  title: 'Components/Teaser',
+  parameters: {},
+  argTypes: {
+    icon: {
+      options: Object.keys(icons),
+      mapping: icons,
+    },
```
georgiee: oh boy, I didn't know about that. I searched for a mapping like these and instead I did some weird manual mapping in my latest PR. You should have warned me and catch this on my PR 😅 I will go ahead and update it now I know how to map values properly. Thank you!

https://github.com/satellytes/satellytes.com/blob/4b842f651a6ac62c31558a2eb90f93c2802fcd3a/src/components/callout/callout.stories.tsx#L34-L50

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const Illustration = styled.div`
+  margin-bottom: 24px;
+`;
+
+interface TeaserProps {
+  title: string;
+  topline?: string;
+  timestamp?: string;
+  hideOverflow?: boolean;
```
georgiee: Does it make sense to always limit the text? I think so yes. I think we should remove this option then. We should be bold to make things simpler.

src/@types/styled-components.d.ts
```diff
@@ -22,6 +22,8 @@ declare module 'styled-components' {
         secondary: string;
         header: string;
         headerLight: string;
+        topline: string;
+        timestamp: string;
```
georgiee, line 24: I don't like the wording `Timestamp`. It's the human readable date and actually not limited to date. Timestamp suggests that I have to pass in a ISO date string or such while it's a simple string to display besides the title with dimmed colors. What about `dateFormatted`, `dateInfo` or such? (I have no explicit recommendation here though). Try out what sounds best for you from the perspective being a user of this component)

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+          <ImageContainer>
+            <TeaserImage className="teaser-img" alt={''} image={image} />
+          </ImageContainer>
+        )}
+        {(topline || timestamp) && (
+          <ToplineContainer>
+            <Topline>{topline}</Topline>
+            <Timestamp>{timestamp}</Timestamp>
+          </ToplineContainer>
+        )}
```
georgiee: This will render if topline or timestamp is given. What happens if one is given? You will render empty nodes and I think you could prevent this here too

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        {largeTitle ? (
```
georgiee: What is `largeTitle` ? When the title is rendered big like here on the right side? I think there is a UX rule hidden, that if there is not content above the title render the title big. It's not a user decision, isn't it? Have you asked the design team about that?

Basically: There is no large headline with an image or icon

Screen Shot 2021-11-15 at 14 48 13

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+    css`
+      overflow: hidden;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+    `}
+`;
+
+const TeaserText = styled(Text)<{ hasImage: boolean; hideOverflow?: boolean }>`
+  margin-bottom: 22px;
+  margin-top: ${(props) => (props.hasImage ? '8px' : '16px')};
```
georgiee: the margin should be applied from the image, that way you can remove the supporting prop `hasImage`, can't you ?

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,149 @@
...
+}) => {
+  const largeTitle = !image && !icon;
+  return (
+    <TeaserContainer>
+      <Link to={linkTo}>
+        <Illustration>{icon}</Illustration>
+        {image && (
+          <ImageContainer>
+            <TeaserImage className="teaser-img" alt={''} image={image} />
+          </ImageContainer>
```
georgiee: I think you can never combine icon and image. The easiest way to prevent this is to offer a single prop where you can pass in either an icon or an image? I already commented on the image prop being not ideal (it's a gatsby specific data structure, while it could be a whole image component).

So one prop named `cover` or such could bring in the illustration or a gatsby image for example. Those are brief ideas while writing the PR and I'm happy to discuss this. Don't take my opinion as the final truth to implement 🤓 Thank you!

georgiee (changes requested):
Nice, thank you for the great addition 🙌 That looks promising.
I commented a few bits. 

1. The image data structure is blocking, I would like to see you pass in an image.
2. I know there is no text available, but such a teaser could use a description. How is it usually used and where. You can discuss this with us in Slack in before hand but such a description would help to clarify the context where people should use this component.

.storybook/preview.js
fischeversenker: We agreed to focus on the light theme for the beginning. To make things easer I suggested to globally turn on the light theme here in the `preview.js`. We can think about this and figure out how to make use of the Storybook-internal Theme switcher in the future. Let's first get the light theme looking splendid :sparkles:

src/components/teasers/teaser.tsx
hofmann-felix: I asked Eric and he said that the headline can wrap. So we can remove the property.

src/components/teasers/teaser.tsx
hofmann-felix: Yes, I've asked the design team and you are right, the rule is: No icon and no image -> large headline. I have also defined the constant this way (`const largeTitle = !image && !icon`), probably it makes sense if I remove the constant and write `!image && !icon` directly into the code, so you can see it  immediately 🤔

src/components/teasers/teaser.tsx
hofmann-felix: The image has an influence on the margin between headline and text, that's why I used this property.
Bildschirmfoto 2021-11-16 um 10 02 50
Bildschirmfoto 2021-11-16 um 10 02 58

src/components/teasers/teaser.tsx
hofmann-felix: With only one property the component would definitely look cleaner 👍
However, some spacing changes depending on whether an icon or an image is used. Is there a best practice to find out in the component which type it is?

src/components/teasers/teaser.stories.tsx
hofmann-felix: Since the icon is also passed directly as a component, it makes sense to do the same for the image. However, I have a few problems with the integration into the storybook: 
`source` should be of type IGatsbyImageData. How would you then add a mock image to the story?

hofmann-felix:
@georgiee 
Thanks for your review 👍 I have added a description based on the one from your Callout component.

src/components/teasers/teaser.tsx
fischeversenker: I can very well imagine that this is a mistake in the designs. I don't see a reason to have different spacings depending on whether the media element is an illustration or an image 🤷 
Maybe you can ask Eric to double check and confirm the spacings?

src/components/teasers/teaser.tsx
georgiee: thanks!

src/components/teasers/teaser.stories.tsx
georgiee: I don't think I can answer this ad-hoc and in text. Maybe we should clarify this in chat. I though you were already able to display an image, can't you get your hands on that reference? Or do you have problems exposing it as a control?

src/components/teasers/teaser.tsx
georgiee: local named variables like `largeTitle` help explaining the purpose. Feel free to do it either way. No need for code golfing (being short as possible) in any case 👍 Thanks for tackling this.

src/components/teasers/teaser.tsx
georgiee: that's sounds very wrong and must be an error in design. I see no reason for this to be different? What do you think?

src/components/teasers/teaser.tsx
georgiee: Full agree, this relates to my other comment about the margin I think?

georgiee:
Once ready, please tell to review again explicitly. Did not look like yet, as some issues are pending or in clarification? If not, I'm happy to check again. Thanks.

src/components/teasers/teaser.tsx
hofmann-felix: As discussed, I assumed that this was a mistake in the design and set the margin to 16px.

src/components/teasers/teaser.stories.tsx
hofmann-felix: I asked Fabian and he said that it is ok to just use a normal HTML image in the storybook

src/components/illustration/illustration.tsx
```diff
@@ -4,6 +4,7 @@ import styled, { css } from 'styled-components';
 
 export enum IllustrationSize {
   NORMAL = 'normal', // 72px
+  MEDIUM = 'medium', // 80px
```
hofmann-felix, line 7: @georgiee I noticed that illustrations with 80px often appear in the design. Therefore I have added another size, I hope that's okay

georgiee: Alright, good for now. Can you make a comment in figma because the 8px difference feels like a design inconsistency rather than a decision. We might want to revert this change then.

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,103 @@
...
+  &:hover {
+    color: ${theme.palette.text.topline};
+
+    p {
+      color: ${theme.palette.text.topline};
+    }
+
+    img {
+      transform: scale(1.05);
+    }
```
georgiee, line 31: Is this effect agreed? Feels some awkward that the image zooms in this case. But let's keep it and I would like to come back to this.

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const TeaserContainer = styled.div`
+  cursor: pointer;
+  overflow: hidden;
+
+  img {
+    transition: transform 0.2s;
+  }
+
+  &:hover {
```
georgiee, line 22: This is spaghetti architecture but it works and I'm thankful 😁
It's spaghetti because this hover definition bleeds into many different things: p, img, general color, svg/path. This cries for problems and confusion in the future and I would rather see the objects itself handle hover.

Most probably with some prop `hover` that you can pass into the receiving components (however this would work as you only receive the instance). I have no clear guide so this is some refactor for afterwards.

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,103 @@
...
+ */
+export const Teaser: React.FC<TeaserProps> = ({
+  topline,
+  title,
+  dateFormatted,
+  cover,
+  linkTo,
+  children,
+}) => {
+  const hasToplineContainer = Boolean(topline || dateFormatted);
```
georgiee, line 79: 👍

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Teasers are used to link to another page and give the user a brief overview of the content of that page.
+ * This requires a headline, the path to the page (linkTo) and a short text, which is entered as a child.
+ * In addition, an illustration or an image, a formatted date and a topline can be displayed.
+ */
+export const Teaser: React.FC<TeaserProps> = ({
+  topline,
+  title,
+  dateFormatted,
```
georgiee, line 74: thansk for the change of the wording.

src/components/teasers/teaser.tsx
```diff
@@ -0,0 +1,103 @@
...
+            {dateFormatted && <Timestamp>{dateFormatted}</Timestamp>}
+          </ToplineContainer>
+        )}
+        {cover ? (
+          <StyledTeaserTitle hasTopline={hasToplineContainer}>
+            {title}
+          </StyledTeaserTitle>
+        ) : (
+          <TeaserTitleLarge>{title}</TeaserTitleLarge>
+        )}
```
georgiee, line 97: I would have though that you can handle the larger title within the same component just by passing in a prop `large` instead you use two titles `StyledTeaserTitle` & `TeaserTitleLarge` which makes reading this component rather complex.

src/components/teasers/teaser.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { ILLUSTRATION_NAMES } from '../illustration/illustration-set';
+
+export default {
+  component: Teaser,
+  title: 'Components/Teaser',
+} as ComponentMeta<typeof Teaser>;
+
+const Template: ComponentStory<typeof Teaser> = (args) => <Teaser {...args} />;
+
+export const Regular = Template.bind({});
```
georgiee, line 16: The teasers looks unnatural wide in the previews because they naturally spread. I would have expected a story where you showcase the natural habitat: A grid that holds multiple instances of a teaser. Something like the special showcase story like I did for the illustrations to show all illustrations.

src/components/teasers/teaser.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+const SampleImage = styled.img`
+  width: 100%;
+`;
+
+export const WithImage = Template.bind({});
+WithImage.args = {
+  ...Regular.args,
+  topline: 'Topline',
+  dateFormatted: '15th November 2021',
+  cover: <SampleImage src={ExampleImage} alt="" />,
```
georgiee, line 56: This prop looks bad in the preview. Can you hide it? It's a great idea to just pass in an img instead of a GatsbyImage. I didn't think of that when you asked 😅

Screen Shot 2021-11-18 at 17 11 52

src/components/teasers/teaser.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Illustration key={name} show={name} size={IllustrationSize.MEDIUM} />
+    );
+  });
+}
+
+export const WithIllustration = Template.bind({});
+WithIllustration.args = {
+  ...Regular.args,
+  cover: (
+    <Illustration show={'space_shuttle_043'} size={IllustrationSize.MEDIUM} />
```
georgiee, line 37: That feels okay. Either pass in an image or an illustration. Now seeing this I can imagine two variants:

```
<Teaser illustration='keyword'/>
<Teaser cover={<img/> }/>
```

where both props are mutually exclusive or one wins over the other. Something for a later refactoring if peopel agree.

georgiee (approved):
This will be one of our most used components throughout the page and I'm glad that we have a first version now 🙌 I have commented on a few different aspects. I don't see massive blockers and because it's a long running PR already I would suggest it's okay to merge now and that we address the remaining problems through smaller PRs after merging this one.

Maybe you can list the pending changes/questions/issues you encounter in the ticket for the teaser so we can handle them afterwards? That would help!

src/components/teasers/teaser.tsx
hofmann-felix: > If it contains an image the image is zoomed in on hover and back out on mouseOut

That's the description in the design. Maybe we need to adjust the values (time and scale factor)

Issue #177, opened 2021-11-08
Component: Quote ContentBlock


fischeversenker:
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3701%3A23643

georgiee:
I think we need a cite source optionally too have such a quote content block.
See this reference https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote

I see myself using it once we have it and I think the design team aligned the design more on the status quo that's why it's missing. I'm going to propose it.

![Screen Shot 2021-11-25 at 17 00 01](https://user-images.githubusercontent.com/1701755/143472550-10628b32-17ca-45ac-8524-3697a11c7203.png)


Good for now though👍

PR #182, opened 2021-11-12
feat: add checkbox component

### LINKED ISSUE #165: Component: Checkbox
### Design
- Without label:
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3705%3A23748

- With label:
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3705%3A24031

### Things to consider

### Expected outcome
A checkbox component that can optionally receive a label.

milljoniaer:
image

### New things:
- new checkbox component
- new storybook story for that component


src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <path
+              d="M1.2002 7.59995L4.15404 10.7999L10.8002 1.19995"
+              stroke="#202840"
+              strokeWidth="1.5"
+              strokeLinecap="round"
+            />
+          </Icon>
+        </StyledCheckbox>
+      </HiddenCheckboxLabel>
+      <LabelText>{props.label}</LabelText>
```
fischeversenker: The label is currently not clickable because it is outside of the `label` element. Please move it back inside the `HiddenCheckboxLabel` to improve the UX.

src/components/checkbox/checkbox.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import Checkbox from './checkbox';
+
+export default {
+  component: Checkbox,
+  title: 'Components/Checkbox',
+  parameters: {
+    layout: 'fullscreen',
+  },
```
fischeversenker: This leads to no padding from Storybook. I suggest to only use this for full-screen elements that span the entire screen width. For smaller inline elements like this I suggest we don't use the `fullscreen` layout.
```suggestion
```

src/components/checkbox/checkbox.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+} as ComponentMeta<typeof Checkbox>;
+
+const Template: ComponentStory<typeof Checkbox> = (args) => (
+  <Checkbox {...args} />
+);
+export const Regular = Template.bind({});
```
fischeversenker: Rather nitpicky, but important: Please add a new line between these two declarations.
```suggestion
);

export const Regular = Template.bind({});
```

src/components/checkbox/checkbox.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+  label: 'Satellytes',
+};
+Regular.parameters = {
+  backgrounds: {
+    values: [
+      { name: 'black', value: '#000' },
+      { name: 'white (figma)', value: '#F2F2F2;' },
+    ],
+  },
+};
```
fischeversenker: Thanks for figuring this out. Is it specific to this component, though? Do we maybe want these to be our global storybook background options? I would say so, yes.
Feel free to add them to the exported `parameters` in `.storybook/preview.js` to make them global :raised_hands:

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+    background: #f7f8fa;
+    border-radius: 4px;
+
+    ${Icon} {
+      visibility: ${checked ? 'visible' : 'hidden'};
+    }
+  `;
+
+  const CheckboxContainer = styled.div`
+    display: flex;
```
fischeversenker: `flex` turns this component into a "block element" which means that it will (almost) always be placed on a new line and will be followed by a new line. To make this more flexible we might want to turn this into `inline-flex` which would allow to have a checkbox "inline".
I'm not sure if this is an actual requirement but when using this component I would expect it to be an "inline element".

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+    align-items: center;
+  `;
+
+  const HiddenCheckboxLabel = styled.label`
+    display: inline-block;
+  `;
+
+  const LabelText = styled(Text)`
```
fischeversenker: The `Text` component has a `margin-bottom` of 16 pixels which does not quite fit here. I would also not recommend to nest a `p` element (`<p>` is a block element, `Text` is `styled.p`) inside a `label` element (`<span>` is an inline element). We should almost never nest block elements inside inline elements. Here a brief introduction into the "block"/"inline" topic: https://www.w3schools.com/html/html_blocks.asp

I suggest to create a new typography type for this special case that does not have any outer spacing as well as uses an inline element (`<span>` in this case) as its base.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+    overflow: hidden;
+    position: absolute;
+    visibility: hidden;
+  `;
+
+  const Icon = styled.svg`
+    display: inline-block;
+    fill: none;
+    margin-top: 5px;
+    margin-left: 5px;
```
fischeversenker: These margins seem to push the checkbox to the bottom and thus break the alignment with the label:
image

Can you finetune this so the checkbox and the label are well aligned?

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Text } from '../typography/typography';
+
+const Checkbox = (props) => {
+  const [checked, setChecked] = useState(true);
+
+  const handleChange = () => {
+    setChecked(!checked);
+    props.onChange(checked);
```
fischeversenker: This will throw an exception if `props.onChange` is not defined. You can make this more failsafe by using ["optional chaining"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining):
```suggestion
    props.onChange?.(checked);
```

fischeversenker (changes requested):
Thanks for this new component :sparkles:

It does not quite match the design specifications and there are a few improvements that I'd like to suggest. Please have a look at my comments.

georgiee:
Thank you for this PR and your efforts! And also thank you Felix for the review of this draft PR!

Forms & Form Elements are a little bit more complex. We also might need some common interface or such for them to work the same. Forms are only used in two places. I consider them a lower priority.

Maybe you feel better when you can deliver components more easily.  For example the quote component (https://github.com/satellytes/satellytes.com/issues/177). Might be more rewarding as I can see that this review could take a while to make it ready to use.

Any objections or would this be a good addition? In parallel this PR could be handled nonetheless but I fear it's going to take a while.

---

1. Besides this, I have also seen that you manually linked the related issue. You can simply type `#` and then search for the issue to link or put in the number of the issue. If it's in the text like `Related to # 123 (without the space)` somewhere in your text it will link automatically.
2. You seem to commit with great effort. I see good messages like `feat: introduce checkbox component` but I would like to point out that your PR is later ideally merged as a single commit (you can select that option). Which means you can happily commit multiple `chore: some information about what you did' or simply `chore: update` if you are tinkering around. If you are more comfortable with git you can also amend commits which is "updating the last commit". Ask us in the channel if you need some git assistance/knowledge 🙌

src/components/checkbox/checkbox.stories.tsx
fischeversenker: We agree that we are focusing on the light theme for now so there is no need to have these settings just yet. We will add them when we tackle the dark theme.

src/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+        >
+          <path
+            d="M7.2002 13.6L10.154 16.7999L16.8002 7.19995"
+            stroke="#202840"
+            strokeWidth="1.5"
+            strokeLinecap="round"
+          />
+        </svg>
+      </StyledCheckbox>
+      <CheckboxLabelText>{props.label}</CheckboxLabelText>
```
fischeversenker: If this component get's used without a label we still render this `CheckboxLabelText` element which leads to remaining whitespace. I suggest to render this element conditionally with something like the below code:
```suggestion
      {props.label && <CheckboxLabelText>{props.label}</CheckboxLabelText>}
```

src/components/checkbox/checkbox.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import Checkbox from './checkbox';
+
+export default {
+  component: Checkbox,
+  title: 'Components/Checkbox',
```
fischeversenker, line 7: We can add an "[Action](https://storybook.js.org/docs/react/essentials/actions)" here to showcase the `onChange` property. With the below addition, Storybook will automatically print something in the Actions tab of the component by attaching a special Action callback to the `onChange` property.
```suggestion
  title: 'Components/Checkbox',
  argTypes: {
    onChange: { action: 'onChange' },
  },
```

fischeversenker (commented):
Thanks for the updates, @milljoniaer :pray:

I left two more comments. Please have a look before we merge this.

Also thanks a lot for raising your concerns, @georgiee.
I do agree that form elements are rather complex and it seems like with React this is not different. I just now read about [Controlled components](https://reactjs.org/docs/forms.html#controlled-components) but I don't see how this applies to a checkbox. @milljoniaer and I had a collaborative session where we took care of finetuning the layout and I now believe this is a solid checkbox component. What else would you expect from a component like this?

milljoniaer:
I have added the requested changes by @fischeversenker. Please verify if they fit. 

Additionally I changed the code a little bit in, since I saw that onChanged returned the inverted boolean of checked in the Action. Is the react state-handling too slow?
But with the workaround in line 9 it works. Is that ok?

fischeversenker (approved):
> Additionally I changed the code a little bit in, since I saw that onChanged returned the inverted boolean of checked in the Action. Is the react state-handling too slow?

Yes, kind of. The react state-handling is "too slow" as in it's "asynchronous". So after the `setChecked()` function is done executing the state value is not yet updated. This has to do with React keeping track of all the changes that happen inside a component to know when to re-render the component. We will talk more about (a)synchronicity soon. For now, consider an asynchronous operation as something that happens in the background.

In my opinion the workaround that you came up with is fine. Another way of doing this would be to use the `useEffect()` hook which allows you to do something whenever the component's state changes. Take a look at the documentation about this "hook" [here](https://reactjs.org/docs/hooks-effect.html).

Since we will work on the `onChange` handler in the near future I suggest we keep your workaround and merge this as it is 🚀

Issue #157, opened 2021-10-26
Component: Icon Illustrations


fischeversenker:
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3659%3A25045

### Things to consider
- Which of the existing Icons is still of any use?
- In Figma there seem to be mostly unique names but there are already duplicates (like "satellite"). How do we deal with this?

### Expected outcome
A new component that can be used to show any of the given illustration icons.

georgiee:
Thank you @hofmann-felix for pointing me to this issue.
I think the best way to make those illustrations accessible are by exposing them through a central export/place so people now what's available. I'm thinking of a central curated place named `illustrations` (those are not icons). That place will give each svg a proper name and everyone will know where to get the illustrations.

```
// illustrations.ts: our central illustration curation

export ShootingStar from '../../illustrations/shooting-star.svg';
export ConstellationBigDipper from '../../illustrations/constellation-big-dipper.svg';
```

and later you can easily access them:
```
import { ShootingStar } from './llustrations';
// then use in jsx <ShootingStar/>


// or maybe bad for tree shaking:
import * as illustrations from './llustrations';

// jsx usage: <illustrations.ShootingStar/>
```

georgiee:
Working on this. I had to export all files from Figma manually, then I used the OSX Automator to clean up the file names and after that I removed the explicit sizes from all svg files with sed:

```
find . -name "*.svg" -type f -exec sed -i  "" 's/ width="144" height="144"//g' {} \;
```

Now I have a folder with SVG which I export and that I can process with Storybook.

PR #187, opened 2021-11-15
feat: add callout component

### LINKED ISSUE #176: Component: TextHighlight ContentBlock
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3713%3A18327

georgiee:
Close #176

---

# Description
This will provide a callout component to highlight text. I struggled with two parts:

+ How to load svgs in Gatsby & Storybook
+ How does the design look on mobile or in edge cases
+ Illustrations are in Figma only

## SVGR
For the loading I chose to pick `gatsby-plugin-svgr` and then manually configure `svgr` for storybook by modifying the webpack configuration. Feels awkward to configure thinks twice, but that's the price of having to different environments (gatsby & storybook).

## Design Issues
Regarding the design I did some best guesses and in parallel I asked in Figma for comments. I got them after I finished this comments so I would keep the current "best guess" and discuss it with you slightly. Basically Eric suggested to remove the icon on small screens after I pointed out that the layout won't work on mobile as it wastes too much of screen estate. I created a solution where I can keep the icon by floating it in the text. I favor this solution. What's your oppinion?

## Illustrations
Illustrations are located in Figma and we are expected to store them piece by piece. I would like to do this for all in one step. I provided the necessary SVGR infrastructure with this PR and placed two examples in `src/illustrations`. Should we create some kind of barrel file to expose all of them through a single `Illustrations` "enum"? Because that way people won't have to import svg files everywhere and know the full list of available illustrations. What do you think?

# Technical Remarks
1. I had to provide a complex argument parsing in order to provide the dropdown. I basically offer strings to select an illustrations and then I swap them for a svg component.  I provided a documentation for this property so people understand that it's not a string to pass in. 

2. I wrote a few bits about the component as JSDoc. I though it's picked up by Storybook and then displayed, but it's not working. Anyone has a clue why ?

3. I wrote no tests. I think that's over the edge right now. We are not working on business critical components and such and should focus on delivering fast right now.  I hope we are aligned on this.

# Preview Image
Here some random previews with the different variants and the story itself.

Screen Shot 2021-11-15 at 11 51 15

Screen Shot 2021-11-15 at 10 52 41

Screen Shot 2021-11-15 at 10 52 59

Screen Shot 2021-11-15 at 11 52 18

Screen Shot 2021-11-15 at 10 52 55


hofmann-felix:
In general, I really like the callout component and I'm looking forward to see it in a "real" blogpost ✨

Regarding the design issues: 
I like your solution with the icon floating in the text and I think it is better than not showing the icon

Regarding the illustrations: 
We need the illustrations in many places (some of them have already been included in other PR's). That's why I like the suggestion with the illustration "enum" because the integration becomes very easy. There is also a [ticket](https://github.com/satellytes/satellytes.com/issues/157) for it. I think it would be useful to implement this illustration component as soon as possible, so we can use it in the new components right away.

hofmann-felix (approved):

package.json
```diff
@@ -56,6 +57,7 @@
     "gatsby-plugin-smoothscroll": "^1.2.0",
     "gatsby-plugin-styled-components": "^5.1.0",
     "gatsby-plugin-typescript": "4.1.2",
+    "gatsby-plugin-svgr": "^3.0.0-beta.0",
```
fischeversenker, line 60: ~~Where is this configured? Is it automatically picked up by Gatsby? How?~~
Apparently I did not see the change in `gatsby-config.js` :joy:

src/components/callout/callout.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+} as ComponentMeta<typeof Callout>;
+
+const Template: ComponentStory<typeof Callout> = (args) => {
+  return <Callout {...args}>{args.children}</Callout>;
+};
+
+export const Regular = Template.bind({});
+Regular.args = {
+  illustration: <ConstellationBigDipper />,
```
fischeversenker: For some reason this does not work. It shows no illustration here without explicitly selecting one. It does work for the `WithLongText` story but not for `Regular`. I can't find any difference... maybe you can figure this out? Of course I don't consider this a blocker... although I'd love to understand what's going on here :thinking:

src/components/callout/callout.tsx
```diff
@@ -0,0 +1,59 @@
...
+    height: 72px;
+    flex: 0 0 72px;
+    order: 1;
+  }
+`;
+
+/**
+ - Use an avatar for attributing actions or content to specific users.
+ - The user's name should always be present when using Avatar – either printed beside the avatar or in a tooltip.
+ **/
```
fischeversenker: Is this related to this component? I don't understand...

src/components/callout/callout.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+`;
+
+const CallOutText = styled.div`
+  ${up('sm')} {
+    flex: 1 1 auto;
+  }
+`;
+const CallOutIcon = styled.div`
+  float: left;
```
fischeversenker, line 31: I got nostalgic for a second... 🤩

src/components/callout/callout.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  argTypes: {
+    illustration: {
+      options: Object.keys(EXAMPLE_ILLUSTRATIONS),
+      mapping: EXAMPLE_ILLUSTRATIONS,
+      description:
+        'This is an example list of icons, there is not default set. You have to provide a svg react component/',
+      control: {
+        type: 'radio', // Type 'select' is automatically inferred when 'options' is defined
+        labels: {
+          // 'labels' maps option values to string labels
```
fischeversenker: For a splendidly clean code base I suggest we are very consistent with _everything_. In this case I'd suggest not to use mixed types of comments (at the end of the line for `type` and as a block for `labels`). Also the placement of the `labels`  comment seems slightly off. Why is it inside the object? I guess one could argue both ways... when in doubt I suggest to not use any comments at all. Most of this information is readily available in the documentation.

```suggestion
        // Type 'select' is automatically inferred when 'options' is defined
        type: 'radio',
        // 'labels' maps option values to string labels
        labels: {
```

src/components/callout/callout.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  parameters: {
+    componentSubtitle:
+      'Display additional information detached from your surrounding text content',
+  },
+  argTypes: {
+    illustration: {
+      options: Object.keys(EXAMPLE_ILLUSTRATIONS),
+      mapping: EXAMPLE_ILLUSTRATIONS,
+      description:
+        'This is an example list of icons, there is not default set. You have to provide a svg react component/',
```
fischeversenker: Not sure about this `not` and this `/`...
```suggestion
        'This is an example list of icons, there is no default set. You have to provide a svg react component.',
```

fischeversenker (approved):
Thanks for this :+1:

Regarding Design: I like the floating icon on small screens. I do think, however, that slightly more space would be good to let it "breathe".

Regarding illustrations: A barrel file that exports all available illustrations sounds great 👌 

Regarding the technical questions:
1. Thanks for the info. I left a comment regarding the comments that you left. We basically have the same structure for the Leadbox comp btw.
2. I don't know what's going on. I had a look at the documentation and concluded we were missing the `addon-docs` addon but apparently that's part of the `addon-essentials` addon. I don't know why this doesn't work :-( 
3. Perfectly fine with me. We never explicitly talked about this but I guess we all silently agreed that there is right now no need for tests. Even though I saw that some of the "old" components do have tests (e.g. `client-list`).

All in all I'm approving since it looks good and feels like a neat little unit that we can start using in the various pages.

georgiee:
I'm going to merge the current version. I keep the mobile version and we will clarify the design later.
Once it's merged we can try to use it in our existing blog articles, I think replacing the blockquote would work
but we have to think about how to pass in the illustration name and such. I can imagine we can implement a solution as we rely on the powerful `rehype` for parsing.

src/components/callout/callout.stories.tsx
georgiee: oh boy, I'm sorry for those stupid text errors. Next time I will check them more carefully. Thank you.

src/components/callout/callout.stories.tsx
georgiee: yes you are right. that is a leftover from where ever I copied it. I'm slowly coming back into "be careful" mode, especially with our now long-term components in storybook. 🙏 I will fix this, both comments are now obsolete.

src/components/callout/callout.tsx
georgiee: I tinkered around with the comments, and I copied this piece of text from some working solution. I lost my actual comment. This text should be recognized by Storybook as the description but it's not. I have now placed the correct description in there again and ignore the fact it's not displayed for now.

src/components/callout/callout.stories.tsx
georgiee: I have updated the args definition. It's now working properly since I switched to a proper args mapping but I still have big problems selecting the default value from the list. The documentation is no help and I guess reading the sources will help. Anyway it's working now.

package.json
georgiee: phew 😅

src/components/callout/callout.tsx
georgiee: 🙏

src/components/callout/callout.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+  ShootingStar: <ShootingStar />,
+  ConstellationBigDipper: <ConstellationBigDipper />,
+};
+
+export default {
+  component: Callout,
+  title: 'Components/Callout',
+  parameters: {
+    componentSubtitle:
+      'Display additional information detached from your surrounding text content',
```
georgiee, line 18: That's the sub title displayed in the docs. The actual description should be taken from the actual comment in the component but it's not picked up.

Screen Shot 2021-11-15 at 17 02 21
:

georgiee:
I'm moving forward by mergign this. Thanks for the comments and approvals!
Next low hanging fruit is probably offering the illustrations export and then trying to include this component in the blog or pick another component suited for #189

PR #188, opened 2021-11-15
feat: add textfield

### LINKED ISSUE #164: Component: Textfield
### Design 
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3705%3A23850

milljoniaer:
### Description
This PR adds a Textfield Component according to the new Design.
For now there is no theme support (everything looks like the light theme).

There is an error State (`props.errorMessage === truely`) in which the border gets "danger" colors and an Error message appears (`props.errorMessage`)

---

### Screenshots

Normal
image

ErrorState: 
image


src/components/textfield/textfield.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import Textfield from './textfield';
+
+export default {
+  component: Textfield,
+  title: 'Components/Textfield',
+  parameters: {
+    layout: 'fullscreen',
+    background: 'white (figma)',
+  },
```
fischeversenker: Let's please only use the `fullscreen` layout for components that span the entire screen width. With the default layout Storybook adds some padding which makes it easier to review the components and nicer to look at.

Regarding the background: Instead of adding this to every component, please add `$lightTheme={true}` to the `<GlobalStyle />` in `.storybook/preview.js` as seen [here](https://github.com/satellytes/satellytes.com/pull/163/files#diff-98b614e1838b171ee71c04450a8f1a562753193fb7d4f53a4de8b9b5a7e980ee).

src/components/textfield/textfield.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+Regular.parameters = {
+  backgrounds: {
+    default: 'white (figma)',
+    values: [
+      { name: 'black', value: '#000' },
+      { name: 'white (figma)', value: '#F2F2F2;' },
+    ],
+  },
+};
```
fischeversenker: Let's differentiate between component-specific properties and global properties. These very much seem like global properties. See my similar comment in the [Checkbox PR](https://github.com/satellytes/satellytes.com/pull/182#discussion_r749369645).

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,46 @@
...
+  display: block;
+  margin-bottom: 4px;
+
+  &.danger {
+    font-weight: 700;
+    color: #ff0d35;
+  }
+`;
+
+const Textfield = (props) => {
```
fischeversenker: Would you mind adding an `Interface` for these properties? This makes it easier to read and also makes the compiler happy 😃

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import styled from 'styled-components';
+import { TextfieldLabel } from './../typography/typography';
+
+const StyledLabel = styled(TextfieldLabel)`
+  display: block;
+  margin-bottom: 4px;
+
+  &.danger {
```
fischeversenker: I suggest to name this class something with `error` to not cause any confusion. The property is called `error` and it makes everything easier if the class name reflects that.

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,46 @@
...
+    border-color: ${props.error ? '#FF0D35' : '#f7f8fa'};
+    background: #f7f8fa;
+  `;
+
+  return (
+    <label>
+      <StyledLabel>{props.label}</StyledLabel>
+      <StyledInputText />
+      <StyledLabel className="danger">
+        {props.error ? props.errorMessage : ''}
```
fischeversenker: Wild suggestion: we don't need an `error` property. We can simply assume that there is no error if the `errorMessage` is falsy (i.e. `"" (empty string)`, `null`, `undefined`, `0`, `false`, ...). If `errorMessage` is set to something truthy we can assume that this textfield has an error.

fischeversenker (commented):
This looks promising :tada:

I left a few comments, mostly regarding the redundant `error` property and the storybook theme-ing.

As @georgiee pointed out there is a lot to take into account when dealing with form fields. For text inputs we might want to forward the `onChange/onInput` events and probably the current `value`. What else would you like to see added, @georgiee?

georgiee:
@fischeversenker regarding the forms/inputs "architecture".

Please see the existing components:
+ [CareerForm](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-form.tsx)
+ [CareertextFields](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-textfields.tsx)
+ [CareerComponents](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-components.tsx)

In addition our framework of choice to handle inputs, validation & errors: [react-hook-form](https://react-hook-form.com/)

1. You will see how we have to forward the reference from `register()` to the input and spread it then `<input {...register("exampleRequired", { required: true })} />`. I don't know what's in there, probably it's accessing one of the many properties available on inputs (value, aria etc)
2. Similar to the textfields we usually build in Angular I moved forward to create [some assemblings](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-components.tsx#L35-L64) in the shape of a form group with inputs + space for errors.

This all being said, maybe it's just a matter of enabling the component to handle the `register()` pattern. There is really not much attached to the native inputs. See here

+ [Input](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-components.tsx#L52-L58)

+ [TextArea](https://github.com/satellytes/satellytes.com/blob/8333b0f1bda0528118f4414c4f566227262d76ff/src/components/career-form/career-components.tsx#L45-L49)

I have problems wrapping my head around this PR though, as I can't grab the full picture. 
+ Will checkboxes share the same pattern or even interface as the textfield? 
+ How do we prevent to implement error handling for any type of input? 
+ Do we need one component per input type or is it enough to handle textrea & inputs (with different types being set) ?
+ What is the big picture of components we need to put this component in context.

Many questions which give me an uncanny feeling of proceeding in isolation with the form controls. I think we need an integrated approach, where we handle forms in general and figure out how to implement the individual form parts.

fischeversenker:
Thanks for the feedback, @georgiee.

I do understand and agree now why you are hesitant when it comes to form elements.

It looks like `react-hook-form` does provide nice tools to work with "controlled components" and I suggest we use them.
There is a well explained approach to using `register` with controlled components [in the "Tips" section here](https://react-hook-form.com/api/useform/register), and there is also a [`useController()` hook](https://react-hook-form.com/api/usecontroller) that can be used to create reusable controlled components.

To me the current career form and its components look a lot more complicated then what I imagined them to be. The "assemblings" don't seem to follow the `react-hook-form` guide on how to deal with controlled components, maybe that would simplify them. I still think we should aim for small and reusable components, even if they are used in only a few places on the page.

Maybe it's not the right job for @milljoniaer right now, but I sense a great learning opportunity in this for all of us.

Regarding the missing full picture: To be honest, I wasn't aware that there needs to be one. I agree now that we need to develop these components in the context of a form. On the other hand we might take an iterative approach where we first get the layout done correctly and then start integrating `react-hook-form`.

I suggest we have a meeting about this topic and will set up one.

milljoniaer:
I have added the requested Changes by @fischeversenker. With the Interface I was not quite sure what to do, but I did some things Google helped me with :) , hope they fit.
Please have another look at the PR.

src/components/typography/typography.tsx
```diff
@@ -197,3 +197,10 @@ export const LeadboxText = styled.p`
   margin: 0 auto;
   max-width: 320px;
 `;
+
+export const TextfieldLabel = styled.span`
```
fischeversenker: As discussed in the meeting just now, I suggest to add one more level of abstraction to all of our typography configurations. In Figma this configuration is called `Text XS`:
image

Would you mind moving this as a local style to the component and using it there. We will start to refactor all the typography settings into a single file once we introduced a proper separation between old and new components.

So the suggestion would be, inside your component file:
```ts
const textXS = css`
  // font settings (except color)
`;

const StyledLabel = styled.span`
  ${textXS}
  // other properties
`;
```

Take a look at [this comment](https://github.com/styled-components/styled-components/issues/773#issuecomment-299667585) if something is unclear.

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,51 @@
...
+    color: #ff0d35;
+  }
+`;
+
+interface Textfield {
+  errorMessage: string;
+  label: string;
+}
+
+const Textfield = ({ label, errorMessage }) => {
```
fischeversenker: This is not quite what I meant with "introduce an interface". I was hoping for a interface for the properties of this component, like so:

```suggestion
interface TextfieldProps {
  errorMessage: string;
  label: string;
}

const Textfield = ({ label, errorMessage }: TextfieldProps) => {
```

This way we can benefit from proper types inside the component.
Before:
image

After:
image

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,51 @@
...
+  })`
+    width: 320px;
+    height: 48px;
+
+    padding-left: 16px;
+    padding-right: 16px;
+
+    font-size: 16px;
+    line-height: 150%;
+    letter-spacing: -0.01em;
```
fischeversenker: This is `Text R` from Figma, so please do the same here as suggested for `Text XS` for the label. Add a new (local) typography setting in this file
```ts
const textR = css`
  // ...
`;
```

and use this style in here, like so:
```diff
-    font-size: 16px;
-    line-height: 150%;
-    letter-spacing: -0.01em;
+    ${textR}
```

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,51 @@
...
+    background: #f7f8fa;
+  `;
+
+  return (
+    <label>
+      <StyledLabel>{label}</StyledLabel>
+      <StyledInputText />
+      {errorMessage && (
+        <StyledLabel className="error">{errorMessage}</StyledLabel>
+      )}
```
fischeversenker: This is almost right but the spacings are slightly off.
image

I suggest to **not re-use the `StyledLabel` for the error message**. Instead introduce a new `StyledErrorMessage` (or similar) that also uses the new `TextXS`. The input label has a `margin-bottom` of 4px, whereas the error message needs to have a `margin-top` of 4px. Also the font-weight and the color differ. These are enough reasons to create a new styled component for this. Mostly the margin though. Also - I've learned - it's rather unusual to use classes with styled components.

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,51 @@
...
+    padding-left: 16px;
+    padding-right: 16px;
+
+    font-size: 16px;
+    line-height: 150%;
+    letter-spacing: -0.01em;
+    color: #202840;
+
+    border: 2px solid;
+    border-color: ${errorMessage ? '#FF0D35' : '#f7f8fa'};
```
fischeversenker, line 53: Well done 👌

fischeversenker (changes requested):
Thanks for the updates. This is getting better and better :sparkles:

After the recent meeting that we had about the mess in the typography file I'd like to ask you to start using the new agreed approach in this component. I left a few comments that point out exactly what needs to be done.
I also left some other comments. Please have a look 🙏

milljoniaer:
I have pushed the requested Changes, please have another look at them.

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const StyledErrorMessage = styled.span`
+  display: block;
+  margin-top: 4px;
+
+  ${TextXS}
+  letter-spacing: -0.01em;
+  font-weight: 700;
+  color: #ff0d35;
+`;
```
fischeversenker: For the sake of consistency I suggest to keep the `${TextXS}` at the very top of these two components. If they both start with the following three lines then it's immediately visible that they basically are very similar.
```css
  ${TextXS}
  display: block;
  margin-...
```

According to Figma the `letter-spacing` is also part of the typography so please move them into `TextXS` and `TextR` to avoid duplication.

src/components/textfield/textfield.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const TextXS = css`
+  font-size: 12px;
+  line-height: 150%;
+`;
+
+const TextR = css`
+  font-size: 16px;
+  line-height: 150%;
+`;
```
fischeversenker: To differentiate between styled components and mere snippets of CSS I suggest to start these snippets in lower case, like regular variables. I know that in the past I might have suggested `TextR`, but reading this now I think it might get confusing in the future. So my suggestion would be to use `textXS` and `textR` instead of `TextXS` and `TextR`. This suggestion is backed by a comment in the styled-components repository: https://github.com/styled-components/styled-components/issues/773#issuecomment-299667585

I know that you used the "upper camel case" naming schema in the Headerblock PR already and I'm fine with both variants. This is obviously not a blocker and we can easily refactor this later. Maybe you have an opinion on this and prefer the one or the other. Feel free to keep the names as they are or rename them if you agree.

fischeversenker (approved):
Thanks for the recent changes :pray:

I added two more comments of things that I suggest we change to improve readability. They don't block this PR so I'm approving. Please have a look and if you agree change them before you merge 🟢

src/components/textfield/textfield.tsx
milljoniaer: Yes, I like the lower Case schema for those css snippets

PR #190, opened 2021-11-16
feat: provide an illustration component

### LINKED ISSUE #157: Component: Icon Illustrations

georgiee:
close #157 

---
## Description
I have manually saved all illustrations from Figma, processed the file content and names slightly to expose them easily with SVGR through a single file. Being there I decided to create an `<Illustration/>` component with a typed property to pick the one you want to display and not to fiddle around with components.

Once this worked I created a special story to showcase all available illustrations in an overview with a search field.

And because this proofs nothing I tried to use the Illustration component in the Callout component. I refactored it to accept a string instead of a component. The string has the same typing as the illustrations. This makes the Callout basically dependent on the Illustration component, which is what it is I think.

You can now use illustrations in Callouts the same way as with the Illustration component itself:

```
<Callout illustration="earth_010"/>
<Illustration show="earth_010"/>
```

I'm not sure about the `show` prop name, sounds more like a boolean.

## Preview
Screen Shot 2021-11-16 at 14 34 51-fullpage

Screen Shot 2021-11-16 at 14 34 35-fullpage

## API
1. There is a closed set of available illustrations. The prop `show` is typed accordingly. 
Screen Shot 2021-11-16 at 14 40 36
2. You can set three different sizes: 72px (normal, default), 144px (large) and inherit which will let the illustration fit its container (basically width 100%).

## Remarks
1. I tried to adapt the Leadbox, but the current css/html architecture is  not suited for a quick change, so let's update it in a separate PR.
2. The jsdoc extraction works for this component and displays my comment in the docs:
Screen Shot 2021-11-16 at 14 43 33

Also the props are partially extracted. It looks better outside of the docs but I think there are some nasty problems hidden since this extraction is a new feature.


src/components/illustration/illustration-files.tsx
```diff
@@ -0,0 +1,57 @@
+/**
```
georgiee: I deliberately decided to split this file from  the export `illustration-set.tsx` where I do the types & keys processing so we can easily update this file without worrying about actual code that does things.

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { ILLUSTRATION_NAMES } from './illustration-set';
+import { IllustrationPreview, PreviewGrid } from './illustration-preview';
+
+export default {
+  component: Illustration,
+  title: 'Components/Illustration',
+  parameters: {
+    componentSubtitle: 'Helps to show one of our illustrations',
+  },
+  /**
```
georgiee: I had argTypes defined here, but they really don't coexist well with the automatically extracted values. Plus they seem to bleed into any story, even if excluded. So weird. I removed it here  as the automatic extraction works okay for the component itself and I define only a single field for the overview story later.

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+   * one. When excluding the control by name there are still leftovers (bug?)
+   *
+   * Similar problems with the defaultValue. It's not clear, nor working properly.
+   * One deprecation warning is also given even though following the official docs.
+   * Reference: https://storybook.js.org/docs/react/api/argtypes
+   */
+} as ComponentMeta<typeof Illustration>;
+
+/**
+ * Special Story
```
georgiee, line 29: I have no clue if this is best practice. It felt like I'm doing something forbidden, but it's so useful. Why wouldn't I add this? Maybe there is a story type that is more isolated? I couldn't find any matching pattern in the docs though.

https://user-images.githubusercontent.com/1701755/141997289-c75d1218-ecd4-46ed-9bfe-6a8215666676.mp4

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+ */
+const Template: ComponentStory<typeof Illustration> = (args) => {
+  return <Illustration {...args} />;
+};
+
+export const Default = Template.bind({});
+
+Default.args = {
+  show: 'alien_002',
+};
```
georgiee: This the the actual component related story we already know. And by the way the value is set as the correct default in the generated list. Did not work before with manual argTypes.

Screen Shot 2021-11-16 at 14 51 38

src/components/illustration/illustration.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface IllustrationProps {
+  /**
+   * Pick an illustration by name.
+   */
+  show: IllustrationType;
+  /**
+   * Choose the size. Defaults to 72x72px, can be large (172x172px) or grow with the container.
+   */
+  size?: IllustrationSize;
```
georgiee: I wrote a few jsdocs comments here. And I swear, they were visible in the automatically generated table in `Docs` at some point (hence the motivation to write them).

Now it looks like this:
Screen Shot 2021-11-16 at 14 52 23

`Search` is wrong and from the overview component. While writing this comment I tried to swap the order and the table updated. So it's a hidden convention that the first story is taken for the table generation I guess. 
Screen Shot 2021-11-16 at 14 53 42

I push the new order. This will remove the overview as the first story which was kind of a good UX I think. Anyway.

src/components/callout/callout.stories.tsx
```diff
@@ -19,25 +12,22 @@ export default {
...
-        'This is an example list of icons, there is no default set. You have to provide a svg react component.',
-      defaultValue: 'ShootingStar',
+        'Pick one of the available illustrations. See the Illustration component for more details.',
       control: {
-        type: 'radio',
-        labels: {
-          None: 'None',
-          ShootingStar: 'Shooting Star',
-          ConstellationBigDipper: 'Constellation (Big Dipper)',
-        },
```
georgiee: I removed the manual mapping and I simpky provide the available illustrations which is then passed to the according parameter. 
Screen Shot 2021-11-16 at 14 57 15

src/components/callout/callout.tsx
```diff
@@ -29,15 +31,11 @@ const CallOutText = styled.div`
...
-  width: 48px;
-  height: 48px;
   padding-right: 0.5em;
 
   ${up('sm')} {
     padding-right: 0;
     padding-left: 0.5em;
-    width: 72px;
-    height: 72px;
     flex: 0 0 72px;
```
georgiee, line 39: i remove the explicit sizing to rely on the illustration being sized appropriately. I removed the smaller variant,
as we don't have such a size for illustrations. It still looked good with that size. I would like to take that easy route as the 48px was a blind guess anyway before.


Screen Shot 2021-11-16 at 15 00 12

src/components/callout/callout.stories.tsx
```diff
@@ -19,25 +12,22 @@ export default {
   },
   argTypes: {
     illustration: {
-      options: Object.keys(EXAMPLE_ILLUSTRATIONS),
-      mapping: EXAMPLE_ILLUSTRATIONS,
+      options: ILLUSTRATION_NAMES,
```
fischeversenker: Now there is no way to select `None`. Could you re-add that?

src/components/illustration/illustration-files.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * This file was manually created.
+ * Based on the svg files exported by hand from Figma.
+ *
+ * Created on 16.11.2021
+ */
```
fischeversenker: It seems very important to you to stress how much manual effort this was. Thanks a lot for taking care :pray:
However, I'm not sure how much value this comment provides... do you insist in keeping it?

src/components/illustration/illustration-files.tsx
```diff
@@ -0,0 +1,57 @@
...
+export { default as scientist_042 } from '../../illustrations/042-scientist.svg';
+export { default as space_shuttle_043 } from '../../illustrations/043-space shuttle.svg';
+export { default as astronaut_044 } from '../../illustrations/044-astronaut.svg';
+export { default as sputnik_045 } from '../../illustrations/045-sputnik.svg';
+export { default as stars_046 } from '../../illustrations/046-stars.svg';
+export { default as sun_047 } from '../../illustrations/047-sun.svg';
+export { default as telescope_048 } from '../../illustrations/048-telescope 1.svg';
+export { default as ufo_049 } from '../../illustrations/049-ufo.svg';
+export { default as uranus_050 } from '../../illustrations/050-uranus.svg';
+export { default as venus_051 } from '../../illustrations/051-venus.svg';
```
fischeversenker, line 51: I personally would prefer the schema `123_name`, e.g. would prefer `051_venus` over `venus_051`. What made you turn around the number and name?

src/components/illustration/illustration.stories.tsx
fischeversenker: If you want these `argTypes` only on the `Regular` story you could add them there instead of here in this "global" config.

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Special Story
+ *
+ * Overview of all illustrations.
+ * It wasn't straight forward to store this story in a separate file
+ * hence I keep it in this file even though it's not a classic story
+ * describing the component itself.
+ *
+ * Comes first to be the first story to be shown.
```
fischeversenker: I'm assuming this last line is outdated?

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  );
+};
+
+export const Overview = OverviewTemplate.bind({}) as any;
+Overview.argTypes = {
+  filterPattern: {
+    name: 'Search',
+    table: {
+      type: { summary: 'string' },
+    },
```
fischeversenker: Since we don't want this control to appear on the Docs page I'm wondering if we can remove this `table` property?

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+    control: {
+      type: 'text',
+    },
+  },
+};
+
+Overview.parameters = {
+  // We only want to expose our
```
fischeversenker: This sentence misses the

src/components/illustration/illustration.tsx
```diff
@@ -0,0 +1,57 @@
...
+  INHERIT = 'inherit', // 100% of the container
+}
+
+export interface IllustrationProps {
+  /**
+   * Pick an illustration by name.
+   */
+  show: IllustrationType;
+  /**
+   * Choose the size. Defaults to 72x72px, can be large (172x172px) or grow with the container.
```
fischeversenker: Large, according to the comments in `IllustrationSize` and the css, is `144x144px`. Would you please update this?

src/components/illustration/illustration.stories.tsx
fischeversenker: I like it! I agree that it feels slightly hacky but it works and provides great value so let's keep it 🙌

src/components/illustration/illustration.tsx
fischeversenker: I feel like "Regular" should be the first story for every component so I'm fine with the now-changed order. That undocumented convention regarding the order affecting the table generation is nice to know.

src/components/illustration/illustration.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * describing the component itself.
+ *
+ * Comes first to be the first story to be shown.
+ */
+interface OverviewProps {
+  filterPattern: string;
+}
+
+/**
+ * General Usage Stories with the common template.
```
fischeversenker: There is only one such story and the next template is introduced without a proper separator-comment so this comment is slightly misleading.

src/components/illustration/illustration-set.ts
```diff
@@ -0,0 +1,9 @@
+import * as LIST from './illustration-files';
+export default LIST;
+
+/**
+ * This file makes the given set of svg files available
+ * to the codebase.
+ */
+export type IllustrationType = keyof typeof LIST;
+export const ILLUSTRATION_NAMES = Object.keys(LIST) as IllustrationType[];
```
fischeversenker, line 10: Very useful type magic in here. Approved 👍

src/components/callout/callout.tsx
```diff
@@ -1,9 +1,11 @@
 import React, { ReactNode } from 'react';
 import styled from 'styled-components';
 import { up } from '../breakpoint/breakpoint';
+import { Illustration } from '../illustration/illustration';
+import { IllustrationType } from '../illustration/illustration-set';
 
-interface CalloutProps {
-  illustration?: React.ReactNode;
+export interface CalloutProps {
```
fischeversenker: Why did you decide to export this?

src/components/illustration/illustration.tsx
```diff
@@ -0,0 +1,57 @@
...
+   */
+  size?: IllustrationSize;
+}
+
+const IllustrationLayout = styled.div<Pick<IllustrationProps, 'size'>>`
```
fischeversenker, line 22: This usage of `Pick<>` is impressive... but seems like overkill here to be honest. What's the benefit of keeping this?

fischeversenker (commented):
Thanks for taking care of this. Especially the laborious extraction of the svgs from Figma :pray:

I left a bunch of comments and I'm also not particularly fond of the property name `show`... I think I understand the API-driven idea that led to `show` but it sounds wrong. I can't think of anything nicer right now (except `illustration` which obviously feels very redundant) so I'm fine with keeping `show` for now.

src/components/illustration/illustration.stories.tsx
georgiee: Thanks 🙏

src/components/illustration/illustration.tsx
georgiee: makes sense and people will come along I guess 😎

src/components/illustration/illustration-files.tsx
georgiee: No, we can remove it. I think it's important to note two things though:
+ When was this file created (I know git blame can tell, but why not being explicit here for readability?)
+ That this is a manual file and not generated, which people might assume?

Basically I fear updates in the future and that we don't know how we created this file in the first place. I'm happy to remove, but on the other side, nobody sees this file anyway in the best case so the additional information doesn't hurt, does it ?

src/components/callout/callout.stories.tsx
georgiee: good point I will provide this again then.

src/components/illustration/illustration-files.tsx
georgiee: Because you can't start variable names with numbers 🤓

src/components/callout/callout.stories.tsx
georgiee: fixed and I took the chance to reduce the variants as it doesn't help in this case anyway.

Screen Shot 2021-11-17 at 06 02 16

src/components/illustration/illustration-files.tsx
georgiee: Removed.

src/components/illustration/illustration.tsx
georgiee: I only use the property `show` in this definition. Would you expect me to use an isolated type with a single property name or just pass in the `IllustrationProps`?

src/components/callout/callout.tsx
georgiee: One desperate attempt to render the docs correctly. I though the extractor needs this to be exported (having some feeling from my experiences with the introspection utility in our customer project). I will revert this of course, sorry this slipped in.

src/components/illustration/illustration-set.ts
georgiee: Thanks!

What I'm still missing is a typed hash. Right now it's easy to use the typed props
but when UI want to access it somewhere else, like in the story in only have this big fat array of keys `ILLUSTRATION_NAMES` which I can access `ILLUSTRATION_NAMES[0]` but I would rather have something like the original export `ILLUSTRATION_HASH.blimp_001` (which should then return a string `blimp_001`) and such. I had a reduce in there but I wasn't able to quickly type it appropriately. So there is a challenge left 😁🤘

src/components/illustration/illustration.stories.tsx
georgiee: Updated. Thanks.

src/components/illustration/illustration.tsx
georgiee: Which is the correct size. I think the 72 bleed into my thoughts, thanks 😬

src/components/illustration/illustration.stories.tsx
georgiee: updated to this. thanks.

> We only want to expose our `filter` property on this special story

src/components/illustration/illustration.stories.tsx
georgiee: Oh, indeed. Sneaky debugging leftovers. Thank you again for your attentive eye.

src/components/illustration/illustration.tsx
```diff
@@ -48,7 +48,7 @@ export const Illustration = (props: IllustrationProps) => {
   const IllustrationSvg = IllustrationSet[props.show];
 
   return (
-    <IllustrationLayout size={props.size ?? IllustrationSize.NORMAL} {...props}>
```
georgiee: leftover from tinkering round with my component it the lead component context where I tried to style our Illustration component with styled components which forwards a `className`. We still might want to support styled components by adding a className to our components but that's a separate discussion.

georgiee:
Thank you @fischeversenker for your thorough review.

I have pushed a commit which handles all of the issues:
https://github.com/satellytes/satellytes.com/pull/190/commits/1a5b37dbb25f62da33a95731346a4367ce2864e0

Once checked, I hope this PR is ready to merge 🙌

src/components/illustration/illustration.tsx
georgiee: By the way, I didn't want to randomly use an impressive typescript tool. `Pick` and the counterpart `Omit` are some great utility types to know and use (wisely) in your daily work 💪

Reference: https://www.typescriptlang.org/docs/handbook/utility-types.html

src/components/illustration/illustration.stories.tsx
georgiee: Removed. Thanks.

src/components/illustration/illustration.stories.tsx
georgiee: I tried this and yes it works. I was curious then. When will the docs engine read it? And what I found is weird.
That parameter is ignored when it's placed on the overview story. I though it's related to the export order but changing this would not work either. It's only accepted on the "normal" story and on the default export. Maybe the `as any` of the overview story makes things worse but at some point I stopped testing this. I will check in future stories again how this behaves.

The more correct place is the default export I think, because that's describing the overall component we are dealing with, so I keep it.

src/components/illustration/illustration-files.tsx
fischeversenker: Alright. Good reason :joy:

src/components/illustration/illustration.tsx
fischeversenker: I would have expected you to just pass in `IllustrationProps`. I don't see the benefit of using `Pick` here. The props are getting destructured and you are only pulling out the `size` property anyways.
Another point: This styled component will receive all the properties of `IllustrationProps` so this generic type might even be considered wrong since it hides away other properties of `IllustrationProps` that are very well available in the scope of the styled component.
My main concern is that it will confuse more juniory developers and make them wonder why this is used here and nowhere else.

This is clearly not a blocker and I'm fine with keeping it so I'll leave it to you to decide.

fischeversenker (approved):
Thanks for the cleanup. From my point of view this is ready to be merged now, yes 🚢

src/components/illustration/illustration.tsx
georgiee: Regarding  this quote:

> might even be considered wrong since it hides away other properties of IllustrationProps that are very well available in the scope of the styled component.

I think we have a misunderstanding. I try to clarify this but thanks for unblocking this issue 👍

---

Here a quick snippet from the related code (simplified)

```
export const Illustration = (props: IllustrationProps) => {
  return (
    <IllustrationLayout size={props.size}> // <-- our styled component with the props in discussion
      <IllustrationSvg />
    </IllustrationLayout>
  );
};
```

You can see that the component `IllustrationLayout` has the following props definition (writte here for demonstration purposes only)

```
interface IllustrationLayout {
   size: IllustrationSize;
}
```

which is different (a actual subset) from the props coming in from the outside, being the full `IllustrationProps`:

```
export interface IllustrationProps {
  show: IllustrationType;
  size?: IllustrationSize;
}
```

That's why this pick make sense. If I would use `IllustrationProps` on the styled component definition (`const IllustrationLayout = styled.div<IllustrationProps>`) I would suggest that there are other properties like `show` which is wrong.

georgiee:
Thanks for the review. I will merge this now 🌟

src/components/illustration/illustration.tsx
fischeversenker: Got it. You are right. I was assuming (without double checking, my bad) that you just did `{...props}` on this component.

georgiee: I did before for the styled components experiment ☝️ Riddle solved 😁

See my explanation:
https://github.com/satellytes/satellytes.com/pull/190#discussion_r750896607

PR #191, opened 2021-11-18
feat: add pagination component 

### LINKED ISSUE #168: Component: Pagination Button
### Design
- prev
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3719%3A16569

- next
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3719%3A16570

### Things to consider

### Expected outcome
A new component that can be used as both Prev and Next button **or** two dedicated components (e.g. `PaginationButtonNext` and `PaginationButtonPrev`). This is up to the implementer.

### LINKED ISSUE #170: Component: Pagination
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3719%3A16602

### Things to consider
- What happens when a user clicks on the current page? The designs imply that there is going to be a dropdown.
- Do we know all pages in advance?
- How does that dropdown look like?

### Expected outcome
A new component to display the pagination bar. No particular business logic expected yet. Maybe an Input for the pages dropdown.

hofmann-felix:
Resolves https://github.com/satellytes/satellytes.com/issues/170
Storybook: https://deploy-preview-191--satellytes-website-storybook.netlify.app

## What's included?
- new Pagination component (is currently only used in the blog overview)

## Technical thoughts
Before implementing the component, I discussed a few technical details with Fabian. I would be happy to hear more opinions about it. 

**Should the pagination change the URL?**
- The advantage of this would be that not all blogpost teasers have to be loaded, but only those that are currently displayed. But this is rather irrelevant because the number of blogposts is small.
When a new blog post is added, all other blogposts move to another page. This may cause problems with the Google preview. That's why we decided that the pagination does not change the URL.

**Should the logic take place in the pagination component or in the parent?**
- We decided to keep the component as simple as possible and build the logic into the parent. 

## Preview
Bildschirmfoto 2021-11-18 um 17 22 33

### DIFF
--- a/src/components/pagination/pagination.tsx
+++ b/src/components/pagination/pagination.tsx
@@ -3,8 +3,8 @@ import styled from 'styled-components';
 import { PrevArrow } from '../icons/buttons-icons/prev-arrow';
 import { NextArrow } from '../icons/buttons-icons/next-arrow';
 import { theme } from '../layout/theme';
-import { PaginationDropdown } from '../typography/typography';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
+import { TextStyles } from '../typography/typography-v2';
 
 interface PaginationProps {
   onPreviousClick: () => any;
@@ -20,7 +20,8 @@ const PaginationContainer = styled.div`
   justify-content: space-between;
 `;
 
-const Dropdown = styled(PaginationDropdown)`
+const Dropdown = styled.select`
+  ${TextStyles.toplineR}
   background: none;
   border: none;
   color: ${theme.palette.text.link.default};
--- a/src/components/pagination/pagination.tsx
+++ b/src/components/pagination/pagination.tsx
@@ -4,6 +4,7 @@ import { PrevArrow } from '../icons/buttons-icons/prev-arrow';
 import { NextArrow } from '../icons/buttons-icons/next-arrow';
 import { theme } from '../layout/theme';
 import { PaginationDropdown } from '../typography/typography';
+import { useTranslation } from 'gatsby-plugin-react-i18next';
 
 interface PaginationProps {
   onPreviousClick: () => any;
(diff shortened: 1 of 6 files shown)


src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  ${(props) =>
+    props.inactive &&
+    css`
+      cursor: default;
+      opacity: 50%;
+    `}
+`;
+
+export const Pagination: React.FC<PaginationProps> = ({
```
feedm3: Using `React.FC<...>` is outdated and shouldn't be used anymore. It better to use the props interface directly and `JSX.Element` as return type.

```ts
export const Pagination = (props: PaginationProps): JSX.Element => {}
```

This blog post goes into detail about this topic: https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,91 @@
...
+`;
+
+const Dropdown = styled(PaginationDropdown)`
+  background: none;
+  border: none;
+  color: ${theme.palette.text.link.default};
+  cursor: pointer;
+`;
+
+const StyledButton = styled.button<{ inactive: boolean }>`
```
feedm3: For buttons that are not active the disabled flag should be used: https://www.w3schools.com/tags/att_disabled.asp

- The property should also be rendered in HTML, which makes it accessible for screen readers and other tools that detect "inactive" buttons. Best to do a double-check if this property is really set. 
- With this property, you can also use the native `:disabled` CSS selector instead of the prop.
- No extra checks in your `onClick` handler is necessary anymore, as the browser won't let you click disabled buttons in the first place.

feedm3 (changes requested):

src/components/pagination/pagination.tsx
hofmann-felix: Thanks, good hint 👍 The component has also become much clearer. I did a double-check and the flag works fine:
Bildschirmfoto 2021-11-23 um 08 33 20

feedm3 (approved):

PR #192, opened 2021-11-18
fix: add useEffect to update transparent prop (header)

hofmann-felix:
Solves problems that came up in this PR: https://github.com/satellytes/satellytes.com/pull/159#discussion_r752152192

### What's included?
- add a useEffect() hook in the header component to update style when `transparent` prop changes
- remove unnecessary stories from header because all design cases are available via storybook controls

### DIFF
--- a/src/components/header/header.stories.tsx
+++ b/src/components/header/header.stories.tsx
@@ -17,17 +17,5 @@ Regular.args = {
   siteTitle: 'Satellytes',
   $lightTheme: false,
   showLanguageSwitch: true,
-};
-
-export const WithLightTheme = Template.bind({});
-WithLightTheme.args = {
-  ...Regular.args,
-  $lightTheme: true,
-};
-
-export const WithLightThemeTransparent = Template.bind({});
-WithLightThemeTransparent.args = {
-  ...Regular.args,
-  $lightTheme: true,
-  transparent: true,
+  transparent: false,
 };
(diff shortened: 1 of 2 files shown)


src/components/header/header.tsx
```diff
@@ -135,6 +135,12 @@ const Header: React.FC<HeaderProps> = (props) => {
   );
   const [hoverWithTransition, setHoverWithTransition] = useState(true);
 
+  const transparent = Boolean(props.transparent);
+
+  useEffect(() => {
```
georgiee, line 140: thank you! Have you checked the docs to understand why this is working?
Here the excellent docs: https://reactjs.org/docs/hooks-effect.html and the comparison with `componentDidMount` & friends from the classic react world

src/components/header/header.stories.tsx
```diff
@@ -17,17 +17,5 @@ Regular.args = {
...
-WithLightTheme.args = {
-  ...Regular.args,
-  $lightTheme: true,
-};
-
-export const WithLightThemeTransparent = Template.bind({});
-WithLightThemeTransparent.args = {
-  ...Regular.args,
-  $lightTheme: true,
-  transparent: true,
```
georgiee, line 32: Much better now 👍

georgiee (approved):
That looks great and I can confirm it's working properly in storybook. Thank you.

Some random thoughts for the future:  You hijacked a `fix` PR to sneak in unrelated changes (which are good for themselves). That's okay here and keeps things simple (thank you for that!) but that's something we should not do in client projects as it hides a change in the changelog. Well we could merge a PR with two commits (one `fix` one `chore` or such of course) but we agreed on one PR equals one commit by squashing them.

PR #193, opened 2021-11-18
fix: add header object to theme

hofmann-felix:
Solves problems that came up in https://github.com/satellytes/satellytes.com/pull/159#discussion_r752155948

### What's included?
- In theme.tsx `header`, `headerLight` and `headerHover` is moved in to a new `header` object

### DIFF
--- a/src/@types/styled-components.d.ts
+++ b/src/@types/styled-components.d.ts
@@ -20,9 +20,11 @@ declare module 'styled-components' {
         defaultLight: string;
         defaultDark: string;
         secondary: string;
-        header: string;
-        headerLight: string;
-        headerHover: string;
+        header: {
+          default: string;
+          light: string;
+          hover: string;
+        };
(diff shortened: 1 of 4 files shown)


georgiee (approved):
I did a shallow review and only checked the gatsby preview without starting storybook locally. The changes look exactly like anticipated 🙌 Thanks for picking this up. How do you feel about it? Make much more sense like so?

The commit is actually of type `refactor` I would say as you are not fixing anything. Same fucntionality, different code. If you want to tell that this is about the header you can also use a `scope` in the commit format. Something like this.

`refactor(header): introduce header hierarchy in the website theme`

The scope usually helps to explain where this change belongs to and a changelog generator can then group changes by the scope. This is usually very useful in monrepos where multiple libraries (aka scopes) are used, but you can try it out in this repo if you like 👍

PR #194, opened 2021-11-18
refactor(header): move positioning to parent

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/pull/159#pullrequestreview-809768342
Preview: https://satellytescommain21751-choreheaderposition.gtsb.io/
Storybook: https://deploy-preview-194--satellytes-website-storybook.netlify.app

### What's included?
- move header's positioning attributes to parent (layout) to make it easier to write stories


src/components/layout/layout.tsx
```diff
@@ -31,6 +31,13 @@ const Main = styled.main`
   margin: 0 auto;
 `;
 
+const HeaderPositioningLayout = styled.div`
```
fischeversenker: I know that you suggested this name, @georgiee... I'm not sure about this `Layout` suffix, though. What's the reasoning? Why not the classical terms `Wrapper` or `Container`? Or `LayoutPositioning` instead of `PositioningLayout`? To me this is simply not a "Layout".

georgiee: It's a rough idea and I'm happy about new ideas. `HeaderContainer` would work for me. What about `HeaderStickyContainer`? wrapper & container are kind of overloaded and hides the nice idea of separating the sticky concern from the header itself.

fischeversenker: I very much like the idea of making it clear in the name that this is about the positioning. `HeaderStickyContainer` works well for me. To make it technically more correct it could also be `HeaderFixedContainer`. It won't make a lot of difference.

For now I suggest to go with `HeaderStickyContainer` 👍

feedm3 (approved):
Well done! 

One thing I noticed is that the component has a horizontal scroll bar in storybook. It also has a menu that cannot be closed once opened. Not high prio, but maybe you can try that somehow time-boxed.

Screenshot 2021-11-23 at 10 26 51

I already approve it as the functionality on the website is perfect 👍

PR #195, opened 2021-11-19
feat: add headerblock

### LINKED ISSUE #171: Component: HeaderBlock
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3659%3A24767

### Things to consider
- Which elements are visible inside that component?
- What versions exist of that component?
The outline in Figma implies that there are two/four variants. If necessary clarify with Eric what the differences are
  - "Large" or "Regular"
  - "With icon" or "Without icon". 
- If you're confused by the designs (which element belongs to which version?), check the grouping on the left hand side. This should clarify things.

### Expected outcome
A fully responsive new component :sparkles:

milljoniaer:
### Description
This PR adds a HeaderBlock according to the Design in Figma. It can have a Topline, Headline, TimeStamp and Author and as Children the Text itself.
This Component should be responsive to mobile devices where the illustration should be underneath the text on the right side.

---
### Some Screenshots with random configurations

Large with Icon (Desktop)
image

Small without Icon (Desktop)
image

Large with Icon without Date (Mobile)
image

Small without date and Timestamp (Mobile)
image


src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,111 @@
...
+  `;
+
+  const Timestamp = styled.p`
+    ${TextS}
+    letter-spacing: -0.01em;
+    color: rgba(0, 0, 0, 0.5);
+    margin-top: ${props.large ? '32px' : '16px'};
+    margin-bottom: 32px;
+  `;
+
```
georgiee: always define them outside. this should cause a warning in the console, maybe you have missed it?

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <BlockWrapper>
+      <TextWrapper>
+        <Topline>{props.topline}</Topline>
+        <Headline>{props.headline}</Headline>
+        {(props.date || props.author) && (
+          <Timestamp>
+            {props.date}
+            {props.date && props.author && ' · '}
+            {props.author}
+          </Timestamp>
```
georgiee: I would not recommend to rely on author & date instead people should just pass in the full string so we can display any text there. I think the name `timestamp` is misleading because it can contain the author too and I think virtually any content should be allowed. It's just an additional slot besides headline and topline. Something like `bottomline`, `subline` or `metaline` (I like this)

Makes this component much simpler and the user of the component can then put together any string. Please change 🙌

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,111 @@
...
+        )}
+        <HeaderBlockText>{props.children}</HeaderBlockText>
+      </TextWrapper>
+      {props.illustration && (
+        <IconWrapper>
+          <Illustration
+            show={props.illustration}
+            size={IllustrationSize.LARGE}
+          />
```
georgiee, line 86: I like that you rely on a prop and use the illustration directly 👍

src/components/header-block/header-block.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+    eget risus varius blandit sit amet non magna.
+  </HeaderBlock>
+);
+
+export const Regular = Template.bind({});
+Regular.args = {
+  topline: 'Topline',
+  headline: 'Headline',
+  date: 'vor 3 Tagen',
+  author: 'von Fabian Dietenberger, Frontend Engineer, Satellytes',
```
georgiee: should be replaced by a string prop, see my other comments related to this in the component

src/components/header-block/header-block.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  illustration: 'rocket_011',
+  large: true,
```
georgiee, line 25: what's thay. do we have different sizes? If you look careful, the design demonstrates the mobile version too, so maybe you mixed that up ?
Screen Shot 2021-11-19 at 14 21 31

src/components/header-block/header-block.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+} as ComponentMeta<typeof HeaderBlock>;
+
+const Template: ComponentStory<typeof HeaderBlock> = (args) => (
+  <HeaderBlock {...args}>
+    Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Morbi
+    leo risus, porta ac consectetur ac, vestibulum at eros. Maecenas sed diam
+    eget risus varius blandit sit amet non magna.
+  </HeaderBlock>
+);
+
```
georgiee, line 17: small and effective story 👍

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,111 @@
...
+  font-weight: normal;
+  font-size: 20px;
+  line-height: 150%;
+`;
+
+const Text = css`
+  font-weight: normal;
+  font-size: 16px;
+  line-height: 150%;
+`;
```
georgiee: as agreed it's inline for now. thanks!
how did you like using them later in the component. does it feel good?

georgiee (changes requested):
Thank you! 👍

I have some comments, one blocker (merge date/author) and one question (large property) which might be a blocker.

src/components/header-block/header-block.tsx
milljoniaer: I started with having them outside, but then I could not access the props values inside the css. Is there another way to do this?

src/components/header-block/header-block.tsx
milljoniaer: Yes, I do like it. Very clean approach.

src/components/header-block/header-block.stories.tsx
milljoniaer: But there is for both, Mobile and Desktop, always a Regular and a Large state, isn't it?

src/components/header-block/header-block.tsx
georgiee: you can pass props to styled components, here a simplfied exampel taken from the repo somewhere (search to find the place. there are plenty of other examples like this. `styled.div<>` contains the prop definition for this specific styled component to tell what's being passed in. You can try it to define but if you struggle remove it (defaults to any I think)

```
<AuroraBackground source={getSource(type)} />

const AuroraBackground = styled.div<AuroraBackgroundProps>`
  background-image: url(${(props) => props.source});
`;

```

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+import styled from 'styled-components';
+import { css } from 'styled-components';
+import { Illustration, IllustrationSize } from '../illustration/illustration';
+import { IllustrationType } from '../illustration/illustration-set';
+import { down, up } from '../breakpoint/breakpoint';
```
feedm3: 2 imports are not used and can be removed:

Screenshot 2021-11-23 at 11 17 11

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,123 @@
...
+  line-height: 150%;
+`;
+
+const TextL = css`
+  font-weight: normal;
+  font-size: 20px;
+  line-height: 150%;
+`;
+
+const Text = css`
```
feedm3: If I remember correctly those are now all defined in `typography-v2` right? So can should be able to update the branch and remove them here.

feedm3 (changes requested):
Looks very good, just 2 small cleanup comments 👍

feedm3 (approved):

src/components/header-block/header-block.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <TextWrapper>
+        <Topline>{props.topline}</Topline>
+        <Headline large={props.large}>{props.headline}</Headline>
+        {props.metaline && (
+          <Metaline large={props.large}>{props.metaline}</Metaline>
+        )}
+        <HeaderBlockText large={props.large}>{props.children}</HeaderBlockText>
+      </TextWrapper>
+      {props.illustration && (
+        <IconWrapper>
```
georgiee, line 82: tip for next time so save one div 🙌

```
const IllusrationStyled = styled(IconWrapper)`
..
`

// and then

<IllusrationStyled/>
```
instead of 

```
<IconWrapper>
  <Illustration/>
</IconWrapper>
```

georgiee (approved):
Let's merge this now💫

PR #199, opened 2021-11-22
feat: add RegularButton component

### LINKED ISSUE #167: Component: Button (Regular)
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3669%3A12612

### Things to consider

milljoniaer:
### Description
This adds the Regular Button just like designed [here](https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3669%3A12612). The label is passed as `children` of the Component.

`onClick` can be passed as property.

hover rules are followed.

---

#### Screenshots

image

hover:
image

---

closes #167


src/components/buttons/RegularButton.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+
+const toplineR = css`
```
feedm3: What does the `R` mean? Better not use abbreviations in general, only if they are super obvious like HTTP.

If this variable is specifically for styles, I would also recommend to add the postfix `Styles` (like `toplineRStyles`)

src/components/buttons/RegularButton.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
feedm3: Filenames in this project always use kebab case names, so `regular-button.strories.tsx` and not camel case.

For consistency this should be used on all new files. In general it's a matter of personal taste, although I find using kebab case names more simple to read when working in the console (not sure if I this theory can be baked by real science :D)

src/components/buttons/RegularButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface RegularButtonProps {
+  children: React.ReactNode;
+  onClick: React.MouseEventHandler;
+}
+
+const RegularButton = ({ children, onClick }: RegularButtonProps) => {
+  return (
+    <StyledButton onClick={(event) => onClick?.(event)}>
+      <ButtonText>{children}</ButtonText>
+      <StyledIcon>
```
feedm3: Shouldn't the icon be optional? I guess the icon is only used when the button acts like a link.

If most of the buttons are links, I would recommend that you create a separate `LinkButton`, which always shows this icon.

src/components/buttons/RegularButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+  ${toplineR}
+  color: #ffffff;
+`;
+
+interface RegularButtonProps {
+  children: React.ReactNode;
+  onClick: React.MouseEventHandler;
+}
+
+const RegularButton = ({ children, onClick }: RegularButtonProps) => {
```
feedm3: Naming it `Button` will automatically make it a regular button.

src/components/buttons/RegularButton.tsx
feedm3: I just saw that you use this style only in one place. In this case, why not put it directly there?

src/components/buttons/RegularButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+const ButtonText = styled.span`
+  padding-left: 16px;
+
+  ${toplineR}
+  color: #ffffff;
+`;
+
+interface RegularButtonProps {
+  children: React.ReactNode;
+  onClick: React.MouseEventHandler;
```
feedm3: Nice typing!

feedm3 (changes requested):

src/components/buttons/RegularButton.tsx
milljoniaer: ToplineR is one of the Text styles defined in figma (and it is called there just like this, I also do not know what the `R` means).

I did not use this css in the styled component directly since @georgiee und @fischeversenker wanted to have a new typography.tsx with those text styles and until we have that one, I shall just do it like above.

src/components/buttons/RegularButton.tsx
milljoniaer: In Figma, there is just one Example for the Button (Regular) and that one has this icon. 
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3669%3A12612

Should the Icon although be optional?

src/components/buttons/RegularButton.tsx
feedm3: All right, if that's the name in Figma that is fine. Maybe you can ask Design if they can use more expressive names, but we don't need to wait for feedback here 👍

src/components/buttons/RegularButton.tsx
feedm3: You are right, every button has this Icon. Unusual, I would bet that it will change in near future 😄  But then it's ok to hardcode it for now.

feedm3 (approved):

PR #198, opened 2021-11-22
feat: add teaser overview story

hofmann-felix:
Resolves issues from https://github.com/satellytes/satellytes.com/issues/173#issuecomment-975220091
Storybook URL: https://deploy-preview-198--satellytes-website-storybook.netlify.app

## What's included? 
- add a story "Teaser Grid" to showcase the natural habitat of a teaser: A grid that holds multiple instances of a teaser. 
- Teaser type (Regular, with Image or with Illustration) can be changed in the controls
- hide `cover` prop in the `WithImage` story

## Preview
Bildschirmfoto 2021-11-22 um 12 10 24

PR #200, opened 2021-11-22
feat: add mdx support to storybook

### LINKED ISSUE #185: Docs & MDX Support
I would like to write a few words about our new components so people understand how and why you will use it. That's available in storybook through DocsPages.  It either extracts comments from a component or you start with mdx in the beginning, basically writing markdown content mixed with your components mark up.

See the [Docs Overview](https://github.com/storybookjs/storybook/tree/next/addons/docs) and the [MDX description](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/mdx.md).

Right now my comments above components are not extracted. Maybe someone knows if this is already supported out of the box and then let's discuss if we would benefit from mdx.

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/185
Storybook: https://deploy-preview-200--satellytes-website-storybook.netlify.app
## What's included?
- [Storybook Docs ](https://github.com/storybookjs/storybook/blob/next/addons/docs/README.md) was added and configured. This allows us to add a mdx documentation to each existing story. 
- A mdx documentation was added to the teaser story. This documentation still needs to be improved, currently it only serves to show the functionality of the plugin. 

## Notes
There are several ways to add documentation: Currently in the Teaser component, the existing story is kept (`teaser.stories.tsx`) and only the documentation (`teaser.stories.tsx`) comes from the mdx file. This allows (as before) the pure component to be displayed under "Canvas" and the mdx documentation under "Docs". 
Bildschirmfoto 2021-11-22 um 14 46 33
Instructions and further options can be found here:
- [How to write CSF stories with MDX docs](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#csf-stories-with-mdx-docs)
- [General story book mdx documentation](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/mdx.md)

## Preview
Bildschirmfoto 2021-11-22 um 14 30 14

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,12 +1,16 @@
 const path = require('path');
 
 module.exports = {
-  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: [
+    '../src/**/*.stories.mdx',
+    '../src/**/*.stories.@(js|jsx|ts|tsx|mdx)',
+  ],
   addons: [
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     'storybook-addon-gatsby',
     'storybook-react-i18next',
+    '@storybook/addon-docs',
   ],
   core: {
     builder: 'webpack5',
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,16 +1,12 @@
 const path = require('path');
 
 module.exports = {
-  stories: [
-    '../src/**/*.stories.mdx',
-    '../src/**/*.stories.@(js|jsx|ts|tsx|mdx)',
-  ],
+  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
   addons: [
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     'storybook-addon-gatsby',
     'storybook-react-i18next',
-    '@storybook/addon-docs',
   ],
   core: {
     builder: 'webpack5',
--- a/src/components/teasers/teaser.stories.mdx
+++ b/src/components/teasers/teaser.stories.mdx
@@ -1,0 +1,32 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Teaser } from './teaser';
+import {Regular, WithImage, WithIllustration} from './teaser.stories'
+
+
+<Meta title="Components/Teaser" component={Teaser} />
+
+# Teaser
+
+Teasers are used to link to another page and give the user a brief overview of the content of that page.
+This requires a `headline`, the path to the page (`linkTo`) and a short text, which is entered as a child.
+In addition, an illustration or an image (`cover`), a formatted date (`dateFormatted`) and a `topline` can be displayed.
+
+
+
+## Teaser Layouts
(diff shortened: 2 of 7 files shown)


.storybook/main.js
```diff
@@ -1,12 +1,16 @@
...
+  stories: [
+    '../src/**/*.stories.mdx',
+    '../src/**/*.stories.@(js|jsx|ts|tsx|mdx)',
+  ],
   addons: [
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     'storybook-addon-gatsby',
     'storybook-react-i18next',
+    '@storybook/addon-docs',
```
fischeversenker: According to the [documentation](https://storybook.js.org/docs/react/writing-docs/introduction) and the [npm package](https://www.npmjs.com/package/@storybook/addon-essentials) this is already part of `@storybook/addon-essentials`. Are you sure we need this explicitly (both in this fiel as well as a dependency in the `package.json`)?

feedm3 (approved):
One step closer to a very cool component library 👍

src/components/teasers/teaser.stories.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+# Teaser
+
+Teasers are used to link to another page and give the user a brief overview of the content of that page.
+This requires a `headline`, the path to the page (`linkTo`) and a short text, which is entered as a child.
+In addition, an illustration or an image (`cover`), a formatted date (`dateFormatted`) and a `topline` can be displayed.
+
+
+
+## Teaser Layouts
```
fischeversenker: Those are a\
\
\
lot of new lines. Do we need them?

src/components/teasers/teaser.stories.tsx
```diff
@@ -1,16 +1,11 @@
...
 import { ILLUSTRATION_NAMES } from '../illustration/illustration-set';
 
-export default {
-  component: Teaser,
-  title: 'Components/Teaser',
-} as ComponentMeta<typeof Teaser>;
-
 const Template: ComponentStory<typeof Teaser> = (args) => <Teaser {...args} />;
 
 export const Regular = Template.bind({});
```
fischeversenker, line 11: Even though it's not part of this PR: Would you mind changing the default value for the `title` property in these stories to `Teaser title` instead of `Leadbox title`?

fischeversenker (commented):
Thanks for taking care of this. It works nicely.

I do wonder if we need to explicitly state `addon-docs` as a dependency. According to the documentation it's already part of `addon-essentials`. Did you check?
I can imagine that this PR does not need to touch the storybook config, the `package.json` nor the `yarn.lock`.

.storybook/main.js
hofmann-felix: Yes you are definitely right. It is already included in `addon-essentials`. I removed the changes

src/components/teasers/teaser.stories.tsx
hofmann-felix: I've already changed it in https://github.com/satellytes/satellytes.com/pull/198. Thanks anyway 👍

fischeversenker (approved):
Thanks for the cleanup. Looks great now :raised_hands:

PR #202, opened 2021-11-22
feat: build storybook on netlify

feedm3:
I added a build pipeline for Storybook on our Netlify account. Storybook should be built with every pull request now. The link will be posted as comment into the PR, like Gatsby Cloud also does.

The `storybook-static` folder is now included in the `.gitignore`. The files there should never be commited.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,9 @@ yarn storybook
 yarn build-storybook
 ```
 
+The `main` branch automatically gets deployed to https://satellytes-website-storybook.netlify.app. Pull Requests on 
+Github will also be updated with a separate deployment URL.
+
 ## Writing a blog post
 
 You can add a blog post by adding a single markdown file. The markdown file contains the blog post itself and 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,6 +31,9 @@ yarn storybook
 yarn build-storybook
 ```
 
+The `main` branch automatically gets deployed to https://satellytes-website-storybook.netlify.app. Pull Requests on
```
fischeversenker: Maybe we can make it more clear that this is about storybook and not the gatbsy application itself? I do realize that this is below the headline "Component development with Storybook" and that the URL contains `storybook` - but I still think we should mention it in the plain text.
```suggestion
The storybook for the `main` branch automatically gets deployed to https://satellytes-website-storybook.netlify.app. Pull Requests on 
```

fischeversenker (approved):
Nice! Thanks for wiring this up :pray:

I left a comment regarding the added documentation. I think it needs to be more precise.
To not block this PR I'm already approving 👍

README.md
feedm3: Yep, agree. Can be a bit confusing if you are not familiar with the setup 👍

PR #203, opened 2021-11-22
chore: add new text styles to a new typograph-v2

### LINKED ISSUE #201: New Typography File for the new Font styles
### Description
Figma defines text styles for each Component. 
All those styles should be added into a new `typography.tsx` to prevent us from implementing them for each component locally.

Those styles should be used in all the new components.

milljoniaer:
Add the new Text styles from Figma into a new `typography-v2.tsx`.
It closes Issue #201 .

refine them in the new components:
- Breadcrumb
- Callout
- Checkbox
- Leadbox
- Navigation
- Textfield

I left the Teaser out since @hofmann-felix is still working on it, right?

I have cross checked the font styles in the new components with storybook and figma, so they should be still/now correct


src/components/typography/typography-v2.tsx
```diff
@@ -0,0 +1,165 @@
+import React from 'react';
```
feedm3: Can be removed and the file type changed to `.ts` instead of `.tsx`. We don't have any JSX in here.

This is what Github is telling me (nice feature):
Screenshot 2021-11-22 at 17 07 02

feedm3 (changes requested):
Nice to have all text styles in one place 💪🏻

I'm thinking about if it would be simpler to put them into an object and export only a `TextStyles` object:
```js
const ToplineR = css`...`;

export const TextStyles = {
  ToplineR,
}
```

and then use it like 
```jsx
import { TextStyles } from '../typography/typography-v2';

const SomeComponent = styled.span`
  ${TextStyles.ToplineR}
`;
```

Why? Because it's simpler to work with auto-complete. Right now you either need to know the exact name or you need to search in the `typography-v2.tsx` name for it. If there is just one `TextStyles` object you can import this and then find the appropriate style via auto-completion. 

I usually don't create those object containers, but in exactly this case I think it could be a good idea 🤔

src/components/typography/typography-v2.tsx
milljoniaer: Thank you, I've implemented the requested Changes. Please have another look at the PR.

src/components/typography/typography-v2.ts
```diff
@@ -0,0 +1,185 @@
+import React from 'react';
```
fischeversenker: This is still unused. Please remove this import :pray:

src/components/typography/typography-v2.ts
```diff
@@ -0,0 +1,185 @@
...
+  textL,
+  textR,
+  textS,
+  textXS,
+  quote,
+  code,
+  menuItemR,
+  menuItemS,
+  menuMeta,
+};
```
fischeversenker, line 150: I like the "export everything as an object approach" that @feedm3 suggested.

I would suggest to define these styles inline, though. Right now, if you wanted to rename or add a style you would need to touch two places. First the variable name that holds the style as well as this object that is built out of all these variables. We could simplify this like this:
```ts
export const TextStyles = {
  timestamp: css`
    font-weight: normal;
    font-size: 14px;
    line-height: 110%;
    letter-spacing: -0.01em;
    color: #000000;
  `,
  toplineR: css`
    font-weight: bold;
    font-size: 16px;
    line-height: 110%;
    color: #000000;
  `,
  ...
};
```

In the end this boils down to personal preference. Opinions about this are very welcome :raised_hands:

fischeversenker (commented):
This is great! Thanks a lot for the cleanup and collecting all these styles. Not only did you collect all the existing styles but you also added the missing ones from Figma. Thanks 🙏 

I like the approach that @feedm3 suggested to group all these styles into a `TextStyles` object. The only difference that I suggest is to define the styles inline to avoid duplication of the variable names. I claim this will be easier to maintain. See my comment. 

There still also is the unused `React` import in the new typography file ☝️

src/components/typography/typography-v2.ts
feedm3: Yep, was also thinking about this. Maybe inlining all those objects can become a bit messier (for example when you want to use some extra constants or so), but right now it would be more simple than the separate objects.

milljoniaer: Yes, I do agree, I implemented the changes. Please have another look at the changes.

feedm3 (approved):

src/components/typography/typography-v2.ts
fischeversenker: Thanks for the feedback :pray:

fischeversenker (approved):
Thanks for the updates. Let's merge this :rocket:

PR #205, opened 2021-11-24
feat: add accordion with story

### LINKED ISSUE #178: Component: Accordion ContentBlock
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3717%3A16145

### Things to consider
- Can there be more than one open accordion item?
- Take a look at the outline on the left hand side to find the states for `Open Mobile` and `Open Desktop`

georgiee:
close #178 
initially based on the previous PR #124  👍

---

Preview here:
https://deploy-preview-205--satellytes-website-storybook.netlify.app/

# Description
This brings the Accordion component to our website. I have abstracted away the different parts from the underlying reachui accordion to make usage easier. I like the API of using the `Accordion` plus several `AccordionSection`.

I really disliked the Storybook default of showing the Canvas. I have adapted the Storybook integration globally to show the docs by default and renamed `Canvas` to `Code` and `Docs` to `Documentation`. A lot of inspiration came from the [World Food Program Designsystem](https://github.com/wfp/designsystem).

I found a [great animation example](https://github.com/raunofreiberg/reach-ui/blob/6742f0cdcef05ba23077f63c21e3ede3bbfcae52/packages/accordion/examples/animated.example.js#L74-L89) based on  [React Spring](https://react-spring.io/basics) a pretty nifty toolkit to animate things. The animation of our accordion looks nice and I think this is a great template for future animations in our website.

# Usage
ideally check the Story but that's the code in a nutshell

```
<Accordion>
  <AccordionSection title="Text with illustration" illustration="some-keyword">
    Lorem ipsum dolor sit amet, consectetur adipisicing elit. Ab dolorem
    itaque nam, quisquam quos sed totam voluptas voluptatem voluptatibus.
    Debitis dicta dolores enim esse fugit impedit odio sed sunt tenetur!
  </AccordionSection>
  
  <AccordionSection title="Text, no illustration">
    {...}
  </AccordionSection>
</Accordion>
```
# Preview
Here an image preview
Screen Shot 2021-11-24 at 10 32 49

# Remarks
+ The animation looks not good during the initial render. I spend some time figuring it out but I would like to postpone this into a second PR.
+ I wrote a super descriptive PR before but I lost the window and the content did not restore so it's a little bit less of text know. Probably good for all of us 👹 Most of the stuff is described in the docs anyway and I will comment the PR a little bit anyway.
+ I had to create a story `Getting Started/Intro` to fix an edge case with the new default of showing the docs panel. It's not working on the initial render (visiting the root). The new page has no canvas so it's forced to show the correct panel. I think this is a great place to write about typography, colors and such anyway so it's okay to start almost empty for now to fix the issue.
+ MDX is great but the rendering of the props is not working as expected. You can pass in a story but you always get the rendering of the involved components instead of the story props. That's not expected but I don't know if I did wrong of if it's a bug.
+ I derived from the design for the mobile view. The design wants the icon as a block. We can do much better and we already did for the `Callout` component. I have used `shape-outside: circle()` the first time ever, which made the text flow around more appropriately. 
Screen Shot 2021-11-24 at 10 31 38


.storybook/preview.js
```diff
@@ -14,6 +14,15 @@ addDecorator((storyFn) => (
 ));
 
 export const parameters = {
+  viewMode: 'docs',
```
georgiee, line 17: This is the magic to make the docs view the default. That is very desirable so I push this change together with this PR to save me from extracting those changes 🙏

.storybook/preview.js
```diff
@@ -14,6 +14,15 @@ addDecorator((storyFn) => (
...
+  previewTabs: {
+    // move the docs tab to the front
+    'storybook/docs/panel': { index: -1, title: 'Documentation' },
+    'storybook/canvas/panel': { title: 'Code' },
+    canvas: {
+      title: 'Code',
+    },
+  },
```
georgiee, line 25: Now good 🙌
Screen Shot 2021-11-24 at 10 36 35

before bad 🤓
Screen Shot 2021-11-24 at 10 36 54

.storybook/preview.js
```diff
@@ -27,4 +36,9 @@ export const parameters = {
     en: 'English',
     de: 'Deutsch',
   },
+  options: {
+    storySort: {
+      order: ['Getting started', 'Components'],
+    },
+  },
```
georgiee, line 43: push the helping story to the front. This is desirable anyway and helps acting as the workaround for the wrong tab selection (not docs but canvas) in the initial rendering

package.json
```diff
@@ -81,6 +82,7 @@
     "react-i18next": "^11.14.2",
     "react-leaflet": "^3.2.2",
     "react-share": "^4.4.0",
+    "react-spring": "^9.3.1",
```
georgiee, line 87: for our accordion height animation

src/components/accordion/accordion-animated-panel.tsx
```diff
@@ -0,0 +1,53 @@
+import { animated, config, useSpring } from 'react-spring';
+import * as ReachAccordion from '@reach/accordion';
+import styled from 'styled-components';
+import { useAccordionItemContext } from '@reach/accordion';
+import { useDivHeight } from './use-div-height';
+import React from 'react';
+
+const AnimatedAccordionPanel = animated(ReachAccordion.AccordionPanel);
```
georgiee, line 8: This is a wrapper coming from react spring and works like styled components as it extends the capabilities of the target component but otherwise doesn't change it. 


> You need the animation-primitive itself, and a special factory called animated. This library animates outside React (for performance reasons). Your view has to know how to handle the animated props you pass to it. This is what animated is there for, it extends native elements to receive animated values.

See here https://react-spring.io/basics for details

src/components/accordion/accordion-animated-panel.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * https://reactjs.org/docs/forwarding-refs.html
+ *
+ * Based on:
+ * https://github.com/raunofreiberg/reach-ui/blob/6742f0cdcef05ba23077f63c21e3ede3bbfcae52/packages/accordion/examples/animated.example.js#L74-L89
+ * 1. We removed the forwardRef as we don't benefit from forwarding the ref as of now)
+ * 2. Added the aria-hidden flag based on this feedback: https://github.com/reach/reach-ui/pull/500#discussion_r403603183
+ */
+export const AccordionAnimatedPanel = ({ children }) => {
+  const { isExpanded } = useAccordionItemContext();
+  const { ref, height } = useDivHeight();
```
georgiee: Don't get confused. `useDivHeight` is a custom hook, see the other file in this PR

src/components/accordion/accordion-header.tsx
```diff
@@ -0,0 +1,84 @@
...
+`;
+
+const resetButton = css`
+  padding: 0;
+  margin: 0;
+  appearance: none;
+  background: 0;
+  border: 0;
+  box-shadow: none;
+`;
```
georgiee, line 33: I had the feeling this makes the actual button styles more readable so I grouped that reset. Might be useful globally but for now local is good.

src/components/accordion/accordion-header.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+interface AccordionChevronProps {
+  open: boolean;
+  className?: string;
+}
+
+const AccordionTitleHeadline = styled.span`
+  ${TextStyles.headlineXS}
+  ${ellipsis()}
```
georgiee, line 21: this is coming from polish. We already have it in the code base so I replaced my manual text ellipsis code with this shortcut. See https://polished.js.org/docs/#ellipsis

src/components/accordion/accordion-animated-panel.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * Read this if forwardRef is unclear:
+ * https://reactjs.org/docs/forwarding-refs.html
+ *
+ * Based on:
+ * https://github.com/raunofreiberg/reach-ui/blob/6742f0cdcef05ba23077f63c21e3ede3bbfcae52/packages/accordion/examples/animated.example.js#L74-L89
+ * 1. We removed the forwardRef as we don't benefit from forwarding the ref as of now)
+ * 2. Added the aria-hidden flag based on this feedback: https://github.com/reach/reach-ui/pull/500#discussion_r403603183
+ */
+export const AccordionAnimatedPanel = ({ children }) => {
+  const { isExpanded } = useAccordionItemContext();
```
georgiee, line 31: This took years until I found it. It's in the documentation but virtually not described. It's a great technique to expose the expanded state of the accordion panel in any component that is interested in. I mean I know the context hooks but I just did not realize how flexible it can be used 👍 I'm glad I found this because without it you don't have access to the currennt state of the panel.

src/components/accordion/accordion.stories.mdx
```diff
@@ -0,0 +1,45 @@
...
++ You can't define your own `AccordionButton`, `AccordionPanel` and such, everything is predefined and bundled in the `AccordionContent` component.
+
+<Canvas>
+  <Story story={Regular}/>
+</Canvas>
+
+
+## Component API
+
+<ArgsTable of={Accordion} />
```
georgiee, line 26: Because I defined a subcomponent of interest in the meta data (`AccordionSection`) we get a second tab.

Screen Shot 2021-11-24 at 10 47 20

The behavior is not expected actually. I thought I pass in the story and I get the story props but it always shows me the two tabs, regardless of the story I pass in.

src/components/accordion/accordion.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+        This section is opened by default as we defined the defaultIndex to be
+        1.
+      </AccordionSection>
+    </Accordion>
+  );
+};
+
+export const AccordionSectionExample = (args) => {
+  return (
+    <Accordion defaultIndex={[0, 1]}>
```
georgiee, line 59: `defaultIndex` is the internal reach ui name. You can pass in a number of an array of number to tell what is opened on the initial rendering. You can't control the panels during runtime. 

I thought about renaming this but I found no better words so I pick the same name as the internal one from reach ui.

src/components/accordion/use-div-height.ts
```diff
@@ -0,0 +1,31 @@
...
+      resizeObserver.observe(ref.current);
+    }
+
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, []);
+
+  return { ref, height };
+}
```
georgiee: I always need some time to understand custom hooks. I never can make any sense out of them on the first sight. I really love this one to always get the current height of a component 🌟

src/components/illustration/illustration.tsx
```diff
@@ -56,7 +60,10 @@ export const Illustration = (props: IllustrationProps) => {
   const IllustrationSvg = IllustrationSet[props.show];
 
   return (
-    <IllustrationLayout size={props.size ?? IllustrationSize.NORMAL}>
+    <IllustrationLayout
+      className={props.className}
+      size={props.size ?? IllustrationSize.NORMAL}
+    >
```
georgiee, line 66: I need to directly wrap this component for the Accordion implementation. I think it's never wrong to be a good citizen of the ecosystem and by supporting a famous framework like styled components (but also others using the same technique).

I could have prevented this by wrapping the Illustration component in an additional styled div. Much easier now. 
Find `const PanelIllustration = styled(Illustration)` in the code to understand.

.storybook/preview.js
feedm3: Nice, this makes storybook so much more user-friendly.

src/components/accordion/use-div-height.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Lovely height measurement, from here:
+ * https://github.com/raunofreiberg/reach-ui/blob/6742f0cdcef05ba23077f63c21e3ede3bbfcae52/packages/accordion/examples/animated.example.js#L19-L39
+ */
+export function useDivHeight() {
+  const ref = useRef(null);
+  const [height, setHeight] = useState(0);
+
+  useEffect(() => {
+    const resizeObserver = new ResizeObserver(([entry]) => {
```
feedm3, line 11: What is the `entry`? I thought that the `ResizeObserver` need some reference to the `div` that gets measured, but there is nothing here.

src/components/accordion/accordion.tsx
```diff
@@ -0,0 +1,98 @@
...
+`;
+
+const PanelContainer = styled.div`
+  ${up('sm')} {
+    display: flex;
+  }
+`;
+
+const PanelIllustration = styled(Illustration)`
+  shape-outside: circle();
```
feedm3: Interesting didn't know this neat trick. But tbh, I think it looks better and is easier to read if we just add a little bit more padding to the right and bottom.

Screenshot 2021-11-24 at 13 44 15

Something to discuss with design.

src/components/accordion/accordion.tsx
```diff
@@ -0,0 +1,98 @@
...
+   */
+  illustration?: IllustrationType;
+}
+
+const AccordionItem = styled(ReachAccordion.AccordionItem)`
+  border-top: 1px solid #eeeeee;
+  border-bottom: 1px solid #eeeeee;
+
+  & + & {
+    margin-top: 34px;
```
feedm3: Looking at the design, the margin should be 5px on desktop and 0 on mobile (probably -1 to overlap the borders)

src/components/illustration/illustration.tsx
feedm3: Classic technique to pass down `className` for styled-components 👍

src/components/accordion/accordion-animated-panel.tsx
feedm3: The hook probably "only" works inside the `Accordion`, as the accordion is creating the context for it's children. So not on `any component` but on any `child component`.

src/components/accordion/use-div-height.ts
feedm3: Finally got how it works: the `ref` return variable is used as input for the function. I think this is a bit unintuitive, as I would expect the input to be passed as a function argument. 

I checked a popular hooks library how it's done there with similar use cases.  Check this code here how they use the `useIntersection `:
https://github.com/streamich/react-use/blob/master/docs/useIntersection.md 

They put the `ref` outside of the function so that you can pass it in as a function argument.

```js
const intersectionRef = React.useRef(null);
const intersection = useIntersection(interestectionRef)
```

I think this is simpler (or more intuitive) to read and would therefore recommend to put the `ref` outside.

feedm3 (changes requested):
Nice implementation. Like the use of react-spring, a library that I wanted to try since years 😄  Finally there is a project to use it 💪 

I had 2 points during the review:
1. The used margins don't fit to the design (too large). And I think the `circular` shape makes the text harder to read.
2. The custom hook is a bit hard to understand, as the return object of the hook is used as input. I added an example of a different hook on how to put the `ref` outside of it to make it more intuitive to read and use.

src/components/accordion/use-div-height.ts
georgiee: Thanks for the link to useIntersection. Makes sense to make this more flexible. I copied this part from the linked sources without thinking too much over details like that. I very much appreciate that specific point though and I'm going to refactor the code.

src/components/accordion/accordion.tsx
georgiee: I will remove this for now to be more conservative. I think the tech dragged me into this direction and not the esthetics anyway 😁

preview for completion and later discussion:
Screen Shot 2021-11-25 at 15 07 13

src/components/accordion/accordion.tsx
georgiee: Interesting, I only checked the reference in the component library. Now looking into the actual design I can see the first time the mobile distance and the distance on desktop is clearly more condensed. Those information are missing in the reference document I think.

Screen Shot 2021-11-25 at 15 10 11
Screen Shot 2021-11-25 at 15 09 40
:

georgiee: I asked in the design team. They suspect some error happened in the design. Now every panel have a border at the top and I add an additional border at the bottom for the last (`:last-of-type`) element. In addition I reduced the spacing. We will go with this first version for now.

georgiee:
I found another issue with the border being too short due to the padding on the accordion container and I fixed the rest. Thanks for your review. Please have another look and I'm happy to merge then.

feedm3 (approved):

PR #208, opened 2021-11-24
chore: provide CODEOWNERS file

georgiee:
Add a CODEOWNERS file to provide automatic PR assignments for the entire team. We use this technique with many teams in a client project's monorepo.


georgiee:
That will work in any case. I will check if we can get automatic assignment without this file in parallel (through the review settings for the repo)

georgiee:
Ok this was quick.

The review settings for team say this:
>  This will automatically route this team's code review requests to individual team members. 

So it's based on the assumption that the team is assigned from someone and then the load balancing kicks in. The settings do not assign the team. This is done through this file. Let's merge and check if this works

fischeversenker (approved):
Let's try 🚀

PR #210, opened 2021-11-24
fix: add custom font to storybook

### LINKED ISSUE #206: Fonts are not loaded properly in Storybook
Currently the request to load `fonts.css` from inside Storybook fails, which leads to `CocoGothic` not being loaded.

image

### Expected outcome
"CocoGothic" is used by all our components inside Storybook.

feedm3:
Custom fonts are now available in Storybook.

I also align the storybook commands to align them with the others and added a small library to also serve the build output.

- `yarn storybook`
- `yarn storybook:build`
- `yarn storybook:start`

The old `yarn build-storybook` is still in until all pull requests with it have been merged or updated. After that we can update the command in Netlify.

Closes #206


ghost:
**satellytes.com-main**

:tada: Your build was successful! [See the Deploy preview here.]([object Promise])

## Build Details

[View the build logs here.](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/674be368-adf0-4ac5-8fae-f9a4493adc51/builds/1e4b73d4-e0ee-4b38-92ae-62cb6fe57314/details)

:clock1: Build time: **3m**

## Performance

### Lighthouse report

| Metric | Score |
| ----- | ---- |
| Performance |   :large_orange_diamond: 71   |
| SEO |   :large_orange_diamond: 85   |

:link: [View full report](https://build-1e4b73d4-e0ee-4b38-92ae-62cb6fe57314.gtsb.io/reports/lighthouse/index.html)

src/components/seo.tsx
```diff
@@ -107,29 +107,6 @@ const SEO: React.FC<SeoProps> = ({
...
-        crossOrigin="anonymous"
-      />
-      <link
-        rel="preload"
-        as="font"
-        href={CocoGothicBoldWoff2}
-        type="font/woff2"
-        crossOrigin="anonymous"
-      />
-
```
fischeversenker, line 132: Thanks for removing this. Please also remove the two unused imports 🙏 
image

gatsby-browser.js
```diff
@@ -1,4 +1,3 @@
-import './src/components/layout/fonts/fonts.css';
```
fischeversenker, line 1: With this removed, where do the fonts come from now? Is the `gatsby-ssr` script run for all requests?

gatsby-ssr.js
```diff
@@ -1,6 +1,39 @@
...
+        as="font"
+        href="/fonts/CocoGothic.woff2"
+        type="font/woff2"
+        crossOrigin="anonymous"
+      />,
+      <link
+        key="preload-bold"
+        rel="preload"
+        as="font"
+        href="/fonts/CocoGothic-Bold.woff2"
```
fischeversenker, line 29: Since the fonts now live in `static/fonts/` how are they being picked up by this? I couldn't find any configuration for the `static` folder that Gatsby would be aware of. Is this a Gatsby convention?

gatsby-ssr.js
```diff
@@ -1,6 +1,39 @@
...
+        type="font/woff2"
+        crossOrigin="anonymous"
+      />,
+      <link key={'fonts-css'} rel="stylesheet" href="/fonts/fonts.css" />,
```
fischeversenker: The other two `<link>`s have a regular string `key`. For the sake of consistency and to avoid confusion I suggest to use this here as well:
```suggestion
      <link key="fonts-css" rel="stylesheet" href="/fonts/fonts.css" />,
```

package.json
```diff
@@ -25,8 +25,10 @@
     "test:format": "prettier \"**/*.{tsx,ts,md}\" --check",
     "postinstall": "husky install",
     "precommit": "lint-staged",
-    "storybook": "start-storybook -p 6006 -s ./",
-    "build-storybook": "build-storybook"
+    "storybook": "start-storybook --port 6006 --static-dir ./static",
+    "storybook:build": "build-storybook --static-dir ./static",
+    "storybook:serve": "serve storybook-static",
```
fischeversenker, line 30: What do we need this for?

fischeversenker (commented):
Nice! I like the new `package.json` script names. I do wonder why we need the `storybook:server`, though. I also left other comments, please take a look 🙏

src/components/seo.tsx
feedm3: Ups, overseen. Thanks for pointing it out 👍

gatsby-browser.js
feedm3: `gatsby-ssr` runs for all HTML files and adds the link to the font file into the header. The output is nearly the same as importing it here, but it's a bit more separated which makes it simpler to use in Storybook.

gatsby-ssr.js
feedm3: Yes,  that is Gatsby convention. Gatsby copies all files from `static` to `public`.

https://www.gatsbyjs.com/docs/how-to/images-and-media/static-folder/

package.json
feedm3: To answer this question and your question in the summary at once: We need this to check Storybook build in the browser on our local machines. Gatsby brings this feature out of the box with `gatbsy serve`, but Storybook doesn't include a server. 
Otherwise you would need to install a webserver on your machine to verify the output folder.

Usually the development and build output should be the same, but since different things happen there, sometimes the dev server works and the build doesn't.

gatsby-ssr.js
fischeversenker: Nice, thanks :+1:

package.json
fischeversenker: > Usually the development and build output should be the same, but since different things happen there, sometimes the dev server works and the build doesn't.

Does that happen with Storybook? Has it happened to you in the past? I'm not convinced that we need to do this for Storybook...

I'd rather use `npx serve storybook-static` when I need to debug something instead of having a solid dependency to `serve`. All in all I'm fine with this, either way. It certainly doesn't do any harm. But it's yet another dependency that we will probably need and use only once in a leap year.

fischeversenker (approved):
Thanks for the changes and the explanations. Looks great now 🙌 

I'm still not convinced we need the `storybook:serve` script. See my comment.

To not block this PR I'm approving so you can merge this if you insist to keep the script and the new dependency 👍

PR #211, opened 2021-11-24
fix: use correct css value and non-null check

feedm3:
This fix fixes 2 things:
1. The annoying assertion warning under the files tab on every pull requests Screenshot 2021-11-24 at 14 58 22
2. My editor told me that the `max-aspect-ratio` value is wrong, so I fixed it according to the CSS docs: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/aspect-ratio


fischeversenker (approved):
Thanks for this 🙌

PR #212, opened 2021-11-24
feat: add custom title to storybook

feedm3:
Just a little bit of customization. The best name I could imagine was "Satellytes UI", feel free to propose something else.

I first checked how WFP is doing it: They have a custom React component to display different styles and a dynamic version number with a dropdown. They then have a somehow hacky webpack call to replace the official logo component with theirs: https://github.com/wfp/designsystem/blob/master/.storybook/main.js#L39-L45

After fiddling out with that I decided to Google the problem, as it felt too hacky to change a simple title. I then found out that there is a theming object where you can just set the title 😄

### DIFF
--- a/.storybook/theme.js
+++ b/.storybook/theme.js
@@ -1,6 +1,6 @@
 import { create } from '@storybook/theming';
 
-export default create({
+export const customTheme = create({
   base: 'light',
   brandTitle: 'Satellytes UI',
 });
--- a/.storybook/theme.js
+++ b/.storybook/theme.js
@@ -1,0 +1,6 @@
+import { create } from '@storybook/theming';
+
+export default create({
+  base: 'light',
+  brandTitle: 'Satellytes UI',
+});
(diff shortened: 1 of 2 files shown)


.storybook/theme.js
```diff
@@ -0,0 +1,6 @@
+import { create } from '@storybook/theming';
+
+export default create({
```
fischeversenker: I'm not a fan of default exports. Does this need to be one?

fischeversenker (approved):
Nice that you figured this out 🙌 
Regarding the name: I can't think of anything that would be more fitting right now, also I do like "Satellytes UI" so I'm perfectly fine with it. 
One question regarding the `default` theme export, please see my comment.
It does not block this PR so I'm approving.

.storybook/theme.js
feedm3: Didn't even think of it as those configuration files usually rely on the default export. But this shouldn't be the case for this file. I change it to `export const customTheme = create(...)`. Let's wait for the build if it still works. If yes, I will merge.

feedm3: It works, will merge it. Thanks.

PR #213, opened 2021-11-25
feat: add file dropper component

### LINKED ISSUE #166: Component: File dropper
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3708%3A17734

### Things to consider

### Expected outcome
- clicking on "browse" opens the OS native file picker
- supports multiple files

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/166
Storybook URL: https://deploy-preview-213--satellytes-website-storybook.netlify.app

# What's included?
- new file dropper component 
- file overview of all uploaded files
- uses npm package [react-dropzone](https://react-dropzone.js.org/)
- mdx FileDropper documentation

# URLs
- [Figma: FileDropper without active files](https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3708%3A17734)
- [Figma: FileDropper with active files](https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3735%3A35936)

# Preview
FileDropper without active files:
Bildschirmfoto 2021-11-25 um 11 38 54

FileDropper with active files:
Bildschirmfoto 2021-11-29 um 14 22 13


src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,88 @@
...
+import styled from 'styled-components';
+import { theme } from '../layout/theme';
+import { Illustration, IllustrationSize } from '../illustration/illustration';
+import { IllustrationType } from '../illustration/illustration-set';
+import { Trans } from 'gatsby-plugin-react-i18next';
+import { TextStyles } from '../typography/typography-v2';
+
+interface FileDropperProps {
+  /**
+   * (Optional) Pass in a valid illustration keyword to show the according illustration.
```
feedm3: `(Optional)` is not needed here, that's already clear with the type.

feedm3 (changes requested):
Looks very good and is working as expected. Thanks 👍 

Some small detail: If you hover over with the mouse and the area is clickable, the mouse cursor should change to "pointer". Not sure if the Design has something else in mind, best to clarify that with them.

src/components/file-dropper/file-dropper.tsx
georgiee: I did that before. That's not clear from the table being generated later so I think it's good for the story to include.

georgiee:
How does it relate to the [current form](https://satellytes.com/career/senior-backend-engineer/) on our career page? When I drop a pdf file there the drop zone hides (or shrinks) and I'm presented with the file list in addition. Is this just the first piece? Because from the overall context I would have expected to see something like the following at least mentioned or included? Are we going to work on the actual file input in a following PR?

Screen Shot 2021-11-25 at 16 43 41

feedm3:
The wrong font is used on the dropped files buttons.

Screenshot 2021-11-30 at 06 42 16

src/components/file-dropper/file-list-item.tsx
```diff
@@ -0,0 +1,103 @@
...
+  margin-right: 24px;
+
+  border-bottom: rgba(0, 0, 0, 0) 1px solid;
+  &:hover {
+    border-bottom: ${theme.palette.text.topline} 1px solid;
+  }
+`;
+
+const RemoveButton = styled.button`
+  ${TextStyles.textR}
```
feedm3, line 59: Not sure if the style is working if you don't end the line with `;`.

georgiee: 1. Missing semicolons are silently ignored and causes the line being merged with the following statement. We need a linter to prevent that in general for styled components. 
2. In this case it's an interpolation and should resolve to the correct values stored in `textR` and a semicolon would be wrong I think, but I have never though about that. Worth to check the final part in css here. Maybe felix can post some observations>

src/components/file-dropper/file-dropper.stories.mdx
```diff
@@ -0,0 +1,17 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { FileDropper } from './file-dropper';
+import {Regular} from './file-dropper.stories'
+
+
+<Meta title="Components/FileDropper" component={FileDropper} />
```
georgiee: Please nest within a form group: `Components/Form/FileDropper` and let us collect everything around forms in there.  I want it like here in the long term 🙏 https://uikit.wfp.org/docs/index.html?path=/docs/components-forms-checkbox--regular

Screen Shot 2021-11-30 at 09 57 48

src/components/file-dropper/file-dropper.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { FileDropper } from './file-dropper';
+
+const Template: ComponentStory<typeof FileDropper> = (args) => (
+  <FileDropper {...args} />
+);
+
+export const Regular = Template.bind({});
+Regular.args = {
+  illustration: 'monitor_024',
+  maxFiles: 1,
```
georgiee, line 12: I can select more than one file 😬
Screen Shot 2021-11-30 at 10 05 03

src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,164 @@
...
+  maxFiles?: number;
+  /**
+   * (Optional) Categories that the user can add to each uploaded file (e.g. 'CV' or 'Cover Letter')
+   * */
+  fileCategories?: string[];
+  /**
+   * (Optional) Pass in a valid illustration keyword to show the according illustration.
+   */
+  illustration?: IllustrationType;
+}
```
georgiee, line 50: So you have basically checked what we are using right now from the dropzone handling in a rather uncoordinated way (https://github.com/satellytes/satellytes.com/blob/main/src/components/career-form/career-file-upload.tsx) or what was your approach?

src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const FileDropperContainer = styled.div<{
```
georgiee, line 56: looks great 🌟

src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,164 @@
...
+    css`
+      align-items: flex-start;
+      padding: 12px 16px;
+    `}
+
+  &:hover {
+    span {
+      border-bottom: 1px solid ${theme.palette.text.link.default};
+    }
+  }
```
georgiee: Is this to draw the underline on the `browse` link? Please make this more explicit by using a className or so.

src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,164 @@
...
+  flex-direction: column;
+  align-items: center;
+  padding: 28px 0;
+  background-color: ${theme.palette.background.leadbox};
+  cursor: pointer;
+  border: 3px solid
+    ${(props) =>
+      props.isDragActive
+        ? theme.palette.text.link.default
+        : theme.palette.background.leadbox};
```
georgiee, line 71: that looks good when I drag a file over it 👍

src/locales/en/translations.json
```diff
@@ -28,6 +28,7 @@
     "salary-expectations": "Salary expectations",
     "cover-letter": "Cover letter",
     "file-category": "File category",
+    "remove-file": "Remove",
```
georgiee, line 31: thanks for handling i18n 👍

src/components/file-dropper/file-dropper.tsx
```diff
@@ -0,0 +1,164 @@
...
+  ${TextStyles.timestamp};
+  letter-spacing: 0;
+  margin: 0;
+`;
+
+const Highlight = styled.span`
+  color: ${theme.palette.text.link.default};
+`;
+
+export const FileDropper = ({
```
georgiee, line 91: I wonder how we are going to use it later. Will it work? Did you spend enough thoughts on that matter or will we discover issues when we integrate it in a form? That could be one nice example in the stories that you can offer after the PR in terms of "how do I actually use it with a form bases on `react-hook-form` (our default) "

georgiee (approved):
Looks good! Thank you. There are some small things that I would like to see delivered with this PR but I give you my approval already. Would be nice if you could change the grouping, check the file limit issue and explain how you came to this implementation to know if this is really something we can use later on in any form without many changes 🙏

src/components/file-dropper/file-list-item.tsx
hofmann-felix: This seems to be a global problem, as it affects all new components that use a `<button />` or a `<select />`.

hofmann-felix: I created an issue for it: https://github.com/satellytes/satellytes.com/issues/235

src/components/file-dropper/file-dropper.tsx
hofmann-felix: Yes, I checked the current file-upload component and also had some thoughts about it: The logic in the current component is only made for the use case "career form".  My plan was to make the new FileDropper component more universal so that it can be used in other places in the future if necessary. All the additional logic used in the current file-upload component will be applied in the new component through the parent.

feedm3 (approved):

PR #215, opened 2021-11-25
chore: remove font color from typography

### LINKED ISSUE #214: Typography: Remove font color
In typography-v2.tsx all text styles are set to black. The colors should not be part of the typography and can therefore be removed from the file. 

Bildschirmfoto 2021-11-25 um 10 28 20

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/214

**What's included?**
- remove colors from typography-v2


fischeversenker (approved):
Thanks for the cleanup 🙏

PR #216, opened 2021-11-25
feat: add quote component

### LINKED ISSUE #177: Component: Quote ContentBlock

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/177
Storybook: https://deploy-preview-216--satellytes-website-storybook.netlify.app

### What's included?
- new Quote content block

### Preview
Bildschirmfoto 2021-11-25 um 13 42 27


src/components/quote/quote.tsx
```diff
@@ -0,0 +1,21 @@
...
+  color: ${theme.palette.text.topline};
+  padding-left: 20px;
+  margin: 0;
+`;
+
+const Quote = ({ children }: QuoteProps) => {
+  return <QuoteText>{children}</QuoteText>;
+};
+
+export default Quote;
```
feedm3: Named exports are preferred over default exports. Some reasons why: https://stackoverflow.com/a/68665805/3141881

So please change it to `export const Quote`.

feedm3 (changes requested):
Simple approach, like it :)

src/components/quote/quote.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { ReactNode } from 'react';
+import styled from 'styled-components';
+import { TextStyles } from '../typography/typography-v2';
+import { theme } from '../layout/theme';
+
+interface QuoteProps {
+  children: ReactNode | ReactNode[];
+}
+
+const QuoteText = styled.p`
```
georgiee: `blockquote` is the correct semantic tag to use for quotes. See here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote.

src/components/quote/quote.tsx
```diff
@@ -0,0 +1,21 @@
...
+  ${TextStyles.quote}
+  color: ${theme.palette.text.topline};
+  padding-left: 20px;
+  margin: 0;
+`;
+
+const Quote = ({ children }: QuoteProps) => {
+  return <QuoteText>{children}</QuoteText>;
```
georgiee, line 18: The design doesn't show any cite sources and such but the this [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) shows a great example of how to do it with `figure` combined with the blockquote. Maybe we can work on that in a following PR to make this component more useful as I can imagine to use this in blog posts if available.

```
<figure>
    <blockquote cite="https://www.huxley.net/bnw/four.html">
        <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
    </blockquote>
    <figcaption>—Aldous Huxley, <cite>Brave New World</cite></figcaption>
</figure>
```

src/components/quote/quote.tsx
georgiee: I fully support this. You almost never need default exports. Sometimes I try it out to challenge my thoughts about this but it's much more powerful sticking to the named exports.

src/components/quote/quote.stories.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+<Meta title="Components/Quote" component={Quote} />
+
+# Quote
+
+Special text formatting for quotes. Used mainly in blog posts. Text is passed through `children`.
+
+<Canvas>
+  <Story story={Regular}/>
+</Canvas>
```
georgiee, line 13: Thanks for taking care of writing at least a little bit 🙌

georgiee (changes requested):
Thanks for that 👍 The wrong tag one is blocking for me otherwise only a few comments.

src/components/quote/quote.tsx
hofmann-felix: Good suggestion, I created an issue (https://github.com/satellytes/satellytes.com/issues/221) for it

georgiee: Thanks 👍

src/components/quote/quote.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactNode } from 'react';
+import styled from 'styled-components';
+import { TextStyles } from '../typography/typography-v2';
+import { theme } from '../layout/theme';
+
+interface QuoteProps {
+  children: ReactNode | ReactNode[];
+}
+
+const QuoteText = styled.blockquote`
```
georgiee, line 10: Thanks for changing 👍

georgiee (approved):
Perfect, let's merge 💪
I have checked the design and found some invisible box shadow. I guess this is an error on the design but I asked for the intention. Too late for this PR, everything looks good now. Thank you.

georgiee (approved):
That's good now. Thanks for the correct tag change. Let's merge 👍

feedm3 (approved):

georgiee:
@hofmann-felix we forgot the border to the left. That was the invisible shadow box.  Look at the left side to spot the border.

Screen Shot 2021-11-29 at 15 16 23

PR #217, opened 2021-11-25
feat: add expandable component

### LINKED ISSUE #174: Component: Expandable ContentBlock
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3721%3A19690

### Things to consider
- Are there supposed to be animations? Animating the height is fairly difficult using CSS so we might want to avoid any animation here initially.

### Expected outcome
Interactive component that allows to show/hide its content.

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/174
Storybook: https://deploy-preview-217--satellytes-website-storybook.netlify.app

### What's included?
- add Expandable content block to show expandable text in blogposts

### Notes
- Currently, there is no animation (except the chevron transition) when the user toggles the expandable. Does it make sense to use more animations? I would be happy to hear a few opinions about it

### Preview
Bildschirmfoto 2021-11-29 um 10 04 11
Bildschirmfoto 2021-11-29 um 10 04 26


src/components/expandable/expandable.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { ReactNode, useState } from 'react';
+import styled, { css } from 'styled-components';
+import ChevronSVG from './expandable-chevron.svg';
```
georgiee: this will fail soon, see the pending PR https://github.com/satellytes/satellytes.com/pull/222 you have to use the named export `ReactComponent`.

src/components/expandable/expandable.stories.mdx
```diff
@@ -0,0 +1,17 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Expandable } from './expandable';
+import {Regular} from './expandable.stories'
+
+<Meta title="Components/Expandable" component={Expandable} />
+
+# Expandable
+
+The Expandable content block is used to show expandable text in blogposts.
```
georgiee: You could add a reference to the [detail disclosure element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) as it's what we are imitating here. Something like "this works similar to...". I think this is good for completion.

src/components/expandable/expandable.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return <Expandable {...args}>{args.children}</Expandable>;
+};
+
+export const Regular = Template.bind({});
+Regular.args = {
+  children: `Lorem ipsum dolor sit amet, consectetur adipiscing elit, 
+sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
+Ut enim ad minim veniam.`,
+  label: 'Label',
+};
```
georgiee: I think the API is good. For a short moment I though about using the prop `summary` instead of `label` just to be closer to the standard component, but I think it's okay, as we are not rebuilding it and we might confuse people otherwise? (or do they feel more comfortable reading the same notation?)

src/components/expandable/expandable.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const Expandable = ({ label, children }: ExpandableProps) => {
+  const [open, setOpen] = useState<boolean>(false);
+
+  return (
+    <ExpandableContainer>
+      <LabelContainer onClick={() => setOpen(!open)}>
+        <ExpandableChevronStyled open={open} />
+        <Label>{label}</Label>
+      </LabelContainer>
+      {open && <ExpandableText>{children}</ExpandableText>}
```
georgiee: You remove the entire content here. Can you check what's the best approach from accessibility view and how the accessibility tree reacts when it's hidden and then shown? Do we need something like live update?

Writing about accessibility: Would it have been possible to base this component on the details disclosure component from the beginning or do we run into styling issues then?

src/components/expandable/expandable.tsx
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+export const Expandable = ({ label, children }: ExpandableProps) => {
+  const [open, setOpen] = useState<boolean>(false);
+
+  return (
+    <ExpandableContainer>
+      <LabelContainer onClick={() => setOpen(!open)}>
+        <ExpandableChevronStyled open={open} />
+        <Label>{label}</Label>
```
georgiee: something is wrong with the alignment:
Screen Shot 2021-11-29 at 11 51 24

georgiee (changes requested):
I have few comments. The design error is blocking
and I would appreciate if you could take some time to dive into the accessibility aspects of showing & hiding content which is super important for screen readers. If you have no experience with that you can ask the team or just start reading a little bit.

src/components/expandable/expandable.stories.tsx
hofmann-felix: Since I am now using the detail disclosure element, I have also renamed `label` to `summary`

src/components/expandable/expandable.tsx
hofmann-felix: I couldn't reproduce the wrong alignment. In which browser did you observe this behavior?

src/components/expandable/expandable.tsx
```diff
@@ -0,0 +1,88 @@
...
+  color: ${theme.palette.text.topline};
+  list-style-type: none;
+
+  &::marker {
+    display: none;
+  }
+
+  &:hover {
+    color: ${theme.palette.text.default};
+  }
```
hofmann-felix: Is it useful to use a vendor prefix here to make it temporarily usable in Safari as well?

```suggestion
  &::marker {
    display: none;
  }
  
  &::-webkit-details-marker{
    display: none;
  }

  &:hover {
    color: ${theme.palette.text.default};
  }
```

feedm3: Yes do it, it's a valid edge case in my opinion. Adding extra tooling for just this case is too much overhead.

src/components/expandable/expandable.tsx
georgiee: Screen Shot 2021-12-06 at 15 18 02

Firefox on Mac. Image shows Firefox left, Chrome right.
Do you only use Chrome by chance? We can postpone this fix to get this change merged though.

src/components/expandable/expandable.stories.tsx
georgiee: 👍

src/components/expandable/expandable.stories.mdx
georgiee: thanks for the note

src/components/expandable/expandable.tsx
georgiee: Time flies, we have now an icon component. Please use the right icon instead of importing the file. You can do this after this PR.
Screen Shot 2021-12-06 at 15 24 40

georgiee (approved):
Let's merge this one. I really appreciate that you went the way to use the standard element. Makes so much sense here. There are two small issues. You can do them afterwards. Just don't forget.

+ Fix the icon alignment
+ Use the icon component we have now

PR #218, opened 2021-11-26
Add Image Content Block

### LINKED ISSUE #179: Component: Image ContentBlock
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3732%3A26922

### Things to consider
Is the "Bottom" layout supposed to be the mobile layout for both "Left" and "Right"?

milljoniaer:
### What is included
ImageContentBlock component with description at the right, left or bottom.
There can also be a Creator Tag at the bottom right.

On mobile devices there will always the bottom Configuration be displayed.

---

### Screenshots
image

image

image


src/@types/global.d.ts
```diff
@@ -3,6 +3,7 @@ declare module '*.woff';
 declare module '*.eot';
 declare module '*.ttf';
 declare module '*.png';
+declare module '.*.jpg';
```
feedm3: Please use the same pattern as for other file types: `*.jpg`

src/components/image-contentBlock/image-contentBlock.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
feedm3: Only use kebap-case filenames, don't mix with camel-case: `image-content-block.*`

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+      : 'margin-top: 12px;'}
+
+  ${({ textAlign }) =>
+    textAlign === 'bottom' ? TextStyles.textXS : TextStyles.textS}
+
+  ${({ textAlign }) =>
+    textAlign === 'bottom' ? 'width 0; min-width:100%;' : ''}
+  ${({ textAlign }) => (textAlign !== 'bottom' ? 'max-width: 228px;' : '')}
+
+  ${down('sm')} {
```
feedm3: Default CSS is always mobile (mobile first principle). So instead of using `down('sm')`, we use `up(md)` for desktop styles. 

There are some exceptions when you can use desktop first, but it shouldn't be necessary here.

feedm3 (dismissed):
Good first version.

Usually, we would not use the native `img` tag but always use the `GatsbyImage` component. Not sure if it's possible, but can you add a Storybook example with `GatsybImage`? If not, is it already possible to use this component on the old [office page](https://satellytes.com/office/)?

src/components/image-contentBlock/image-contentBlock.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import ExampleImage from '../../images/office/sy-office-01.jpg';
+import ImageContentBlock from './image-contentBlock';
+import styled from 'styled-components';
+
+export default {
+  component: ImageContentBlock,
+  title: 'Components/ImageContentBlock',
```
georgiee: That name is no good. Check the other names, do you see a pattern? 

Screen Shot 2021-11-29 at 06 43 44


We always remove the `ContentBlock` postfix. It's of no use for use. Please rename including the name of the component. I think `<Image/>` will work for us as it differs from `GatsbyImage` if not then we can still rename in the future.

I now see another pattern, you named `HeaderBlock` with a postfix `Block` as it would collide with the actual `Header` probably? I guess you just followed the idea for naming here? I would go backwards and say, that we should rename that component before using it extensively though. Something like `ArticleHeader` as it's the text at the beginning of an article in the best case. The designer took a shortcut and used this component then for entire content blocks (like on Services) which blurs the borders too much. Anyway, I have created an issue for that: https://github.com/satellytes/satellytes.com/issues/220

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+const BlockWrapper = styled.div<{
+  textAlign: 'right' | 'left' | 'bottom';
+}>`
+  display: inline-flex;
+  flex-direction: ${({ textAlign }) =>
+    textAlign === 'right'
+      ? 'row'
+      : textAlign === 'left'
+      ? 'row-reverse'
+      : 'column'};
```
georgiee: this is not readable at all, please create a local function at the top of the file to "calculate" instead of having such an inline definition. I don't like a single ternary operator for readability reasons, chaining it makes it worse so I suggest to use a dedicated `if` or surch in the function.

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import styled from 'styled-components';
+import { down } from '../breakpoint/breakpoint';
+import { theme } from '../layout/theme';
+import { TextStyles } from '../typography/typography-v2';
+
+const BlockWrapper = styled.div<{
+  textAlign: 'right' | 'left' | 'bottom';
```
georgiee: I think we don't have any rules here yet, but I prefer an isolated `BlockWrapperProps` instead of inlining the definition. You can that type (use an interface though) just before this line.

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  ${({ textAlign }) =>
+    textAlign === 'bottom' ? 'width 0; min-width:100%;' : ''}
+  ${({ textAlign }) => (textAlign !== 'bottom' ? 'max-width: 228px;' : '')}
+
+  ${down('sm')} {
+    ${TextStyles.textXS}
+
+    margin: 12px 0;
+    width 0;
```
georgiee: I think that's not good here

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+    min-width: 100%;
+    max-width: 100;
+  }
+
+  ${theme.palette.text.default};
+`;
```
georgiee: I think you arrived at a place that works by hard work and trying out a lot. I would ask you to refactor the component by using either flex or grid. This will make it much more elegant. In both worlds you can use an attribute `order` to change the order. If you don't like any layout (like in the text below image) just disable the flex or grid layout.

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+  description,
+  attribution,
+}: ImageContentBlockProps) => {
+  return (
+    <BlockWrapper textAlign={textAlign}>
+      <ImageWrapper>
+        {children}
+        {attribution && (
+          <AttributionContainer>
+            Photo by{' '}
```
georgiee: This is not covered by the translation system. Have you though about it ?

src/components/image-contentBlock/image-contentBlock.tsx
```diff
@@ -0,0 +1,116 @@
...
+const ImageContentBlock = ({
+  children,
+  textAlign,
+  description,
+  attribution,
+}: ImageContentBlockProps) => {
+  return (
+    <BlockWrapper textAlign={textAlign}>
+      <ImageWrapper>
+        {children}
```
georgiee: I think it's good to pass in the children. That way image & GatsbyImage works in the best case. I guess we will hit a few issues with the layout though as use superimpose a specific layout where the gatsby image might have different ways of including an image (block vs inline an such)

georgiee (changes requested):
Thanks for this.
1. I would like to ask you to refactor this specific component again. See my comment in the `TextWrapper`. You made it work 💪🌟 Now I would like to guide you into a way that is more modern by relying on flex or grid css. Ask the team in the chat for details if you need help.  This is the reason to reject this PR. I'm torn though. We want to move fast to get things done, so I can perfectly imagine to merge this PR with a **mandatory** refactoring directly afterwards. Let's talk what's the best.

2. I have also commented on the name. Something is off.

3. The component still feels weird with the text included. I would like to hear about the opinion of the team @satellytes/website-crew. Does it feel good to have a single image component in the storybook that ships with the layout + text possibility? I would only see the image + attribution as the core element. Everything besides that is an additional component. I encouraged to move forward with the current version to get things done, so I think it's okay but a discussion about this feels never wrong.

src/@types/global.d.ts
milljoniaer: Oops, typo :+1:

src/components/image-contentBlock/image-contentBlock.stories.tsx
milljoniaer: I am sorry, I have forgotten it another time...
Fixed it, and will remember it

milljoniaer:
@feedm3 
> Not sure if it's possible, but can you add a Storybook example with `GatsybImage`?

I tried it, but when I use GatsbyImage, the story disappears....

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: Thanks for the tip, I implemented it :+1:

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: I actually thought about that, I have tried the "interface-way" in another component, and was thinking if I like the way with less variables better, but I have implemented it using an interface

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: I did not thought about it (probably since in Figma there is only "Photo by") and I thought that for this creator tag the english should then be fine (since the rest of the design is in german)...
Shall I support translation ? And who do I do this?

georgiee: Don't trust the design too much 🤓 Think about the final result and imaging being a Germany visitor. Would you like to read random english words while the page is German? Please fix.

Please ask @hofmann-felix for support as he used it before too. We use https://react.i18next.com/ so please read the documentation in addition a little bit to know the foundation as this will hunt you in many places anyway 👹

src/components/image-contentBlock/image-contentBlock.tsx
georgiee: @milljoniaer any thoughts on my feedback?

src/components/image-contentBlock/image-contentBlock.tsx
georgiee: Less is not better here as it makes reading the code more difficult because you mix two very different things. Usually we have defined rules that ideally are enforced through the linter for things like that, so your machine would complain. Right now we don't have any rule, so let's see how you think about that the next time.

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: Okay, changed it

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: I am using flex, but afaik I did that also before, but I now used gap for the space between the elements. The change of the flex-direction seemed like a good approach to me

src/components/image-content-block/image-content-block.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  </Image>
+);
+
+export const Regular = Template.bind({});
+Regular.args = {
+  textAlign: 'right',
+  description:
+    'Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Morbi leo risus, porta ac consectetur ac, vestibulum at eros. ',
+  attribution: {
+    source: 'www.somelink.de',
```
georgiee: always use `www.example.com` for random urls you use for testing. That's the official test domain by IANA. Nobody can ever register it, so it's safe. Imagine www.somelink.de points to some NSFW content in the future, I think you don't want to be the person to have put it here 🤓

The same is valid and even more important fore mails. Don't use random mail addresses when testing or even sending mails. Always use `info@example.com` and such. The domain `example.com` again is the important part. Otherwise you annoy other people or leak internal information from some client.

src/components/image-content-block/image-content-block.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+export default {
+  component: Image,
+  title: 'Components/Image',
+  argTypes: {
+    textAlign: {
+      options: ['left', 'right', 'bottom'],
+      control: { type: 'radio' },
+    },
+  },
+} as ComponentMeta<typeof Image>;
```
georgiee: this part is usually ignore if you provide a mdx files so you can omit it. You define this through the meta tag in the mdx file then.

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,116 @@
...
+const BlockWrapper = styled.div<TextAlignProps>`
+  display: inline-flex;
+  flex-direction: column;
+
+  gap: 12px 24px;
+
+  width: 100%;
+
+  ${up('sm')} {
+    flex-direction: ${({ textAlign }) => mapTextAlignToFlex(textAlign)};
```
georgiee: much better with the external mapping and the flex layout as the base 👍

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: Okay, yes, I implemented localisation

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: 👍

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: Oops, thanks, fixed it

src/components/image-contentBlock/image-contentBlock.tsx
milljoniaer: So, do we leave it like that?

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,117 @@
...
+      textAlign === 'bottom' &&
+      css`
+        ${TextStyles.textXS}
+      `}
+
+    ${({ textAlign }) =>
+      textAlign !== 'bottom' &&
+      css`
+        ${TextStyles.textS}
+        max-width: 228px;
```
georgiee: we have a lack of requirements. I would have defined this as a shrinking and growing flex area with a percentage. But actually it feels okay with the fixed size so let's keep it.

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,117 @@
...
+import styled, { css } from 'styled-components';
+import { up } from '../breakpoint/breakpoint';
+import { theme } from '../layout/theme';
+import { TextStyles } from '../typography/typography-v2';
+
+const mapTextAlignToFlex = (textAlign: 'right' | 'left' | 'bottom') => {
+  if (textAlign === 'right') {
+    return 'row';
+  } else if (textAlign === 'left') {
+    return 'row-reverse';
```
georgiee: I mentioned the `order` attribute which is much more explicit than just reversing everything. This works here as we only have two elements but remember the `order` property to swap children to arbitrary places in the flex layout 👍

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,117 @@
...
+    return 'column';
+  }
+};
+
+interface TextAlignProps {
+  textAlign: 'right' | 'left' | 'bottom';
+}
+
+const BlockWrapper = styled.div<TextAlignProps>`
+  display: inline-flex;
```
georgiee: Why would you define this an an inline container? Just make it flex which renders as a block to spread to the full width. That way you can also remove the width 100%.

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,117 @@
...
+  ${TextStyles.textXS}
+
+  max-width: 100%;
+  color: ${theme.palette.text.default};
+
+  ${up('sm')} {
+    ${({ textAlign }) =>
+      textAlign === 'bottom' &&
+      css`
+        ${TextStyles.textXS}
```
georgiee: Interesting that the text has different sizes even though it's displayed for the same image in the same viewport. That just makes no sense to me and I have expected a unique size across all variants.

src/components/image-content-block/image-content-block.tsx
```diff
@@ -0,0 +1,117 @@
...
+      css`
+        ${TextStyles.textS}
+        max-width: 228px;
+      `}
+  }
+`;
+
+interface ImageContentBlockProps {
+  children: React.ReactNode;
+  textAlign: 'right' | 'left' | 'bottom';
```
georgiee: I really don't like the name `textAlign` because you are changing the layout and not the text alignment. You are creating a false expectation from the API. I would name this `textPlacement` or maybe you have some other alternative.

src/locales/de/translations.json
```diff
@@ -160,5 +160,8 @@
       "link": "Karriere"
     }
   },
+  "image": {
+    "attribution": "Foto von "
+  },
```
georgiee, line 178: Ideally you should provide an interpolation for things like that as you can't know the sentence structure of the future translation. See https://www.i18next.com/translation-function/interpolation. Something like `"key": "{{what}} is {{how}}`.

For here it's okay and makes things easier as we have a component involved anyway.

src/components/image-content-block/image-content-block.stories.mdx
```diff
@@ -0,0 +1,23 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Image } from './image-content-block';
+import { Regular } from './image-content-block.stories';
+
+<Meta title="Components/Image" component={Image} />
+
+# Image
+
+The ContentBlock/Image Component can be used to display pictures with a description and an optional `attribution` tag in the bottom right corner.
+The image itself is passed as children. (e.g. as GatsbyImage Component, but always with `width: 100%;`for correct layouting).
```
georgiee: > but always with `width: 100%`

Is this something we expect or something we apply? It's not clear from the text.

georgiee (changes requested):
Thanks for the next iteration. I still have a few remarks but it looks good already. Coudl you update this PR another time? Thank you!

src/components/image-content-block/image-content-block.stories.tsx
milljoniaer: That is a point, thank you for the information about ww.example.com 👍 
I changed it.

src/components/image-content-block/image-content-block.stories.tsx
milljoniaer: Oh, yes, that was a fail, fixed it 👍

src/components/image-content-block/image-content-block.tsx
milljoniaer: 👍

src/components/image-content-block/image-content-block.tsx
milljoniaer: I have already talked to Fabian about that, and he also said, that it is okay for now 👍

src/components/image-content-block/image-content-block.tsx
milljoniaer: Okay, yes, this way felt quite cool to me, since I could also do the bottom thing within.
But I will read about `order` for the next time.

src/components/image-content-block/image-content-block.tsx
milljoniaer: Way cleaner, thank you 👍

src/components/image-content-block/image-content-block.tsx
milljoniaer: Yes it seems a little bit weird (but it is just like that Figma).

src/components/image-content-block/image-content-block.tsx
milljoniaer: 👍

src/components/image-content-block/image-content-block.stories.mdx
milljoniaer: We expect it, I have changed it

src/locales/de/translations.json
milljoniaer: I will remember that for next time 👍

milljoniaer:
Thank you for our feedback. I think everything should be implemented now.
Additionally I have changed the file names of the now called "image" component into "image.*" and moved the whole folder to the new components.

src/components/image-contentBlock/image-contentBlock.tsx
georgiee: ok good I have no agenda to force order 🙏

src/components/image-contentBlock/image-contentBlock.tsx
georgiee: yeah I like to give positive head-ups instead of criticizing only in PRs 👍

src/components/image-contentBlock/image-contentBlock.tsx
georgiee: thanks 🙌

georgiee (approved):
Let's merge that one finally. Thanks for your efforts

src/new-components/image/image.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const Regular = Template.bind({});
+Regular.args = {
+  textAlign: 'right',
+  description:
+    'Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Morbi leo risus, porta ac consectetur ac, vestibulum at eros. ',
+  attribution: {
+    source: 'https://www.example.com',
+    creator: 'Banksy',
+  },
+};
```
fischeversenker, line 26: Let's only use the `mdx` format for documentation. You can define this story in the `stories.mdx` file so we can completely remove the `stories.tsx` file.

src/new-components/image/image.tsx
```diff
@@ -0,0 +1,116 @@
...
+`;
+
+interface ImageContentBlockProps {
+  children: React.ReactNode;
+  textAlign: 'right' | 'left' | 'bottom';
+  description: string;
+  attribution?: {
+    source: string;
+    creator: string;
+  };
```
fischeversenker, line 87: I suggest to do this the other way around: the image should be passed as a property, the description should be the children. Why? Do we know if it's always enough to have `description` as a plain string? Do we want to be able to have bold/italic text in the Image description? Also, it is really the "description" of the image or is it just some content that happens to be accompanied by an image?
I do understand that having the "description" passed in as a property makes it a lot easier to deal with the different font-sizes and responsiveness... but still, this feels upside down.

I'm sorry for the late but fundamental feedback, but I'd really like to see this topic addressed. Or you convince me that this way (Image as child, description as property) is solid?

fischeversenker (commented):
Well. Now it's already merged :joy:

I admit that this feedback might come too late. I would have loved to see a few things addressed in here. Maybe these are now topics for a future refactoring of this new component :sparkles:

src/new-components/image/image.stories.tsx
georgiee: oh snap, you are right I missed this

src/new-components/image/image.tsx
georgiee: I think we are heading in some wrong direction.
1. I don't want to use an image tag like this: `<Image image={}>lorem ipsim</Image>`
2. I never wanted to have an image with text block combine with attribution but I forced myself to sit tight until we have more time for refactorings.

Maybe we should go back two steps, what do we want to solve?

1. Attribution. Something I desperately wanted in the blog so we have it here. Maybe a simple attribution wrapper helps here or we create a [higher order component (hoc)](https://reactjs.org/docs/higher-order-components.html) for this specific purpose?
2. Text. This is probably what Felix writes about. From user perspective I see this:

```
<ImageWithCaption image={}>
 lorem ipsum
</ImageWithCaption>
```

This could also work as a HOC too wrapping any image.


This means if we rename this component to `ImageWithCaption` and swap the props description & children(image) we should be fine. In addition we should remove the attribution from this component and let some other component handle it. Best case:

```
const ImageWithCaptionAndAttribution = withAttribution(ImageWithCaption)

<ImageWithCaptionAndAttribution image={} attribution={}>
 lorem ipsum
</ImageWithCaptionAndAttribution>
```

that's a nice DX and a great split of responsibilties. What do you think?

fischeversenker: Where else would you use that `withAttribution` hoc? The concept is nice... but in this particular case it looks very much overengineered to me.
I do like the very explicit `<ImageWithCaption/>`. However, I think whether the `<Image/>` that this component receives contains an attribution or not should not bother the `ImageWithCaption` component. I'd rather have an `ImageWithAttribution` component that can be used as a regular `Image` everywhere. Whether this is `withAttribution(Image)` or just a new regular component doesn't matter. I'd be careful with these advanced concepts like HOCs for our rather simple website. I'd very much argue that we want our website to be as junior-friendly as possible.

PR #222, opened 2021-11-29
fix(build): force parity of building with SVGR in storybook and gatsby

georgiee:
This will fix the build in the context of SVGR. @feedm3 reported that the Gatsby build can't use the Storybook export as the exports svg files are not matching. Storybook uses the `default` export, while Gatsby creates two exports `default` and `ReactComponent`. Actually the config is virtually matching, this is expected behavior by SVGR when svg files are processed by multiple loaders. See https://react-svgr.com/docs/webpack/#use-with-url-loader-or-file-loader.

We don't have two loaders in storybook, hence the default export.
The fix is forcing SVGR to always export. Together with the fix I can use `<Illustration/>` in both worlds as expected ✅

Remarks: This requires SVGR in version 6 which was released just 12 hours ago. Only there you have the feature to force the named export. See the [changelog](https://github.com/gregberge/svgr/releases/tag/v6.0.0). An alternative workaround would have been to add a second loader to match the Gatsby environment, which we might want to do in the future to have a full parity of build systems anyway? Good for now.

Thanks for reporting @feedm3 🙌


georgiee:
There was a build error:
> Error: src/components/accordion/accordion-header.tsx(55,8): error TS2604: JSX element type 'ChevronSVG' does not have any construct or call signatures.

The chevron is a direct import in the accordion header and relied on the default export. I fixed it and updated the PR.

hofmann-felix (approved):

PR #223, opened 2021-11-29
chore: enhance HeaderBlock documentation

milljoniaer:
### What is included
- Short Documentation for the HeaderBlock (`header-block.stories.mdx`)
- Including the Example from the Landing-page
- Making the props `large`, `metaline` optional


src/components/header-block/header-block.tsx
```diff
@@ -8,14 +8,14 @@ import { TextStyles } from '../typography/typography-v2';
 interface HeaderBlockProps {
   topline: string;
   headline: string;
-  metaline: string;
+  metaline?: string;
   children: string;
-  large: boolean;
-  illustration?: IllustrationType | null;
+  large?: boolean;
+  illustration?: IllustrationType;
```
fischeversenker, line 14: I like these changes... but they should not be hidden away in a PR to enhance the documentation. In this case it's fine but please don't hesitate to open multiple very small PRs that are very focused in the future (like: "make some header props optional").

src/components/header-block/header-block.stories.tsx
```diff
@@ -1,18 +1,14 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
-import HeaderBlock from './header-block';
+import { HeaderBlock } from './header-block';
 
 export default {
   component: HeaderBlock,
   title: 'Components/HeaderBlock',
 } as ComponentMeta<typeof HeaderBlock>;
```
fischeversenker: You need to remove this entire `default` export for the `mdx` to have an effect.

fischeversenker (commented):
Thanks for enhancing the documentation 👍 

This is almost good to go :rocket:

I wonder... when you ran this locally, didn't you notice that your content from the MDX does not have any effect? You need to remove the export of the `ComponentMeta` from the `stries.tsx` file. See my comment.

I also left another comment regarding the changes to the component that you secretly provide in this PR. See my other comment.

src/components/header-block/header-block.tsx
milljoniaer: Okay, I'll do it like that next time

src/components/header-block/header-block.stories.tsx
milljoniaer: It is kind of weird though, since I saw the changes locally in the storybook, but in netlify they have not been visible.
But now it works everywhere :+1:

src/components/header-block/header-block.stories.tsx
```diff
@@ -1,18 +1,9 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
```
fischeversenker: `ComponentMeta` is not used in this file anymore. Please remove it :pray:

I recommend to always look at your own PR to make sure you didn't forget something obvious. You would've probably seen this notification:
image

```suggestion
import { ComponentStory } from '@storybook/react';
```

fischeversenker (approved):
Thanks for the updates. Looks fine now ⭐ 
There is one unused import, see my comment. It's not a blocker so I'm approving this already. It would still be nice if you took care of it before you merge (you can "commit my suggestion" so you don't need to do it locally, see the comment with the suggested change) 👍

src/components/header-block/header-block.stories.tsx
milljoniaer: Makes sense, thank you

Issue #136, opened 2021-09-16
Show visual feedback after copying a code block


feedm3:
After clicking on "Copy", we want to show the user visual feedback.

Github does it like this:

https://user-images.githubusercontent.com/7814926/133559909-c80b1b41-d688-433f-abfb-22fea2ed94b6.mp4

@georgiee already did some experiments, which you can read here: https://github.com/satellytes/satellytes.com/pull/130#issuecomment-913462511

georgiee:
Jonathan is working on that.

PR #227, opened 2021-11-29
feat: add IconButton

### LINKED ISSUE #169: Component: Button (copy)
### Design
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3708%3A14588

### Things to consider

### LINKED ISSUE #228: Button Issues
Our button is based on a real button element. That's good.

The implementation looks odd though:
https://github.com/satellytes/satellytes.com/blob/07e2040d4e548c0a6d4c368b1369b5a8acd9ab95/src/components/buttons/button.tsx#L29-L37

You can see that there is a dedicated `onClick` event handler whereas I would have expected to spread the props to pass through any option that is valid for a button (disabled, onClick and such).

Something like that:

```
export interface ButtonProps extends React.ComponentPropsWithoutRef<"button"> {
// additional props if any
}

const Button = (props: ButtonProps) => {
  return (
    <StyledButton {...props}>
      <ButtonText>{props.children}</ButtonText>
```

This is a much nicer citizen of a UI component library, isn't it ?

milljoniaer:
### Description
This adds an IconButton, which displays Icons which should be passed as children (Icon components).

Additionally this wraps the stories of the Regular Button and the IconButton into one.

image

---

closes #169


feedm3:
You can make this a lot more reuseable, by not creating a `CopyButton` (I would not call this `Copy`, as this name is also used for other components like text stuff) but a generic  `IconButton`. Because you `CopyButton` is just an `IconButton` with the `navigator.clipboard.writeText(text);` click handler. The copy can be a good example in Storybook ;-) 

Look how material UI is doing the implementation ([Docs](https://mui.com/components/buttons/#upload-button)), maybe you can do it like that:
```
<IconButton aria-label="upload picture">
  <PhotoCamera />
</IconButton>
```

src/components/buttons/copy.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { Copy } from './copy';
+
+const Template: ComponentStory<typeof Copy> = (args) => <Copy {...args} />;
+
+export const Regular = Template.bind({});
+Regular.args = {
+  textToBeCopied: 'Nothing at all!!!',
```
georgiee: I know this is fun for yourself but please don't leak informal palceholders like that into the sources. You should use lorem ipsum or some descriptive text instead.

georgiee (changes requested):
My first feeling also was: Oh gosh, this is a missed opportunity for an icon button (like Fabian commented) but I know where you are coming from. We have an issue for exactly this button type so you made the right conclusions: issue -> new component -> new story.

Variant 1:
Here I would appreciate if you remove the story and just include the button in the existing button story. In addition we need a little bit more to make this a useful component. Foremost the copy feedback. Otherwise this component really don't bring much.

See the linked issue about [copy feedback](https://github.com/satellytes/satellytes.com/issues/136) and maybe ask the design team for guidance.


Variant 2:
Postpone this PR and deliver a truly icon component. Combined with the new icon component I'm offering through a PR (#230) this should be not that complex I hope.

I think variant 2 can be quicker but variant 1 could be fun with some small animation combined with useSprings and such. Your choice and I put this PR on hold until then. 

I want to emphasize: This is by no means a mistake by you but a situation where the lack of requirements causes some additional thoughts and adjustment the moment you opened this specific task. 

Thank you!💪

milljoniaer:
Thank you for comments. Together with @fischeversenker, I have decided to change the Copy component into an IconButton component as @feedm3 suggested.

@georgiee I would suggest to do the onClick animation of Icon Buttons or Buttons in general another Issue iteration and stick to the basic design requirements for now, if that is okay?

Please have another look at the changes I have made.

src/components/buttons/button.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+import { IconButton } from './icon-button';
+import { Button } from './button';
+import { Icon } from './../../new-components/icon/icon';
+
+<Meta title="Components/Button" component={(Button, IconButton)} />
+
+# Button
+
+There are multiple Button components for different situations. They all take an `onClick` property to control its functionality.
+The displayed values are passed as `children` property. Only the valid types of those property differ for the different components.
```
fischeversenker: > Only the valid types of those property differ for the different components.

I don't understand. What do you mean by that?

src/components/buttons/button.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+  <Story name="Regular">
+    <Button onClick={() => console.log('Regular Button clicked')} children="">
+      Alle Services
+    </Button>
+  </Story>
+</Canvas>
+
+#### IconButton
+
+This is the Icon Button. It is always of the size `24x24`. You can choose the displayed Icon by passing an [Icon](?path=/docs/components-icon--regular) as children.
```
fischeversenker: Well done, providing a link to the `Icon` component 👌

src/components/buttons/button.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+#### Regular
+
+This is the Regular Button, which should be used in all the standard use cases. Pass the `label` as `children` to the Button.
+
+<Canvas withSource="open">
+  <Story name="Regular">
+    <Button onClick={() => console.log('Regular Button clicked')} children="">
+      Alle Services
+    </Button>
```
fischeversenker: The empty `children` property is confusing here. Looks like a leftover?

Also the indentation in this story leads to weird indentation in Storybook:
image

I couldn't find a way to get rid of this behavior except to write everything in one line. Let's ignore this for now, but let's at least remove the empty `children` property:
```suggestion
    <Button onClick={() => console.log('Regular Button clicked')}>
      Alle Services
    </Button>
```

fischeversenker (commented):
Thanks for the recent changes and the additional documentation for the existing `Button` 👌 
I left a few comments. Especially the one regarding the first paragraph of the documentation is something I'd like to clarify before we merge this.

Also, one more thing that would be nice to have in this PR would be move both of these new components into the `new-components` folder. It should be as easy as moving these three files since they are not yet used anywhere.

src/components/buttons/button.stories.mdx
feedm3: Also note that `children` is a special property in React that should never be set as an attribute.

src/components/buttons/icon-button.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+`;
+
+interface CopyProps {
+  onClick: React.MouseEventHandler;
+  children: React.ReactNode;
+}
+
+export const IconButton = ({ onClick, children }: CopyProps) => {
+  return <StyledButton onClick={onClick}>{children}</StyledButton>;
```
georgiee: You only expect to receive an icon through children. You never expect more. Why would you use the children? There is no rule to use children so I would omit it. Instead I would try a similar technique as with the illustrations:

1. Hard code the icon here and only pass in the icon name
2. The prop for the icon can be typed similar to how the accordion types the illlustration type. Great for auto completion and icon safety.

See the accordion:

https://github.com/satellytes/satellytes.com/blob/5cc2990d9777d0791d165ae8ca12570fcf59ab0d/src/components/accordion/accordion.tsx#L36

src/components/buttons/icon-button.tsx
```diff
@@ -0,0 +1,35 @@
...
+    background: #3e61ee;
+  }
+
+  &:active {
+    background: ${lighten(0.1, '#3e61ee')};
+  }
+`;
+
+interface CopyProps {
+  onClick: React.MouseEventHandler;
```
georgiee: I would not expose the `onClick` instead see  this general issue:
https://github.com/satellytes/satellytes.com/issues/228

Passing in `onClick` feels super weird for things that support that by default. It's always a big warning sign so please be careful with that in the future.

georgiee (changes requested):
I have left a comment about the props mainly which I would like to see addressed or discussed. Thanks.

src/components/buttons/icon-button.tsx
fischeversenker: To make it more clear: The idea is to use `export interface CopyProps extends React.ComponentPropsWithoutRef<"button">` and then to spread the `props` (by the way, they should really be called `IconButtonProps` instead of `CopyProps`).

I don't think of this as such a big deal. The component receives an `onClick` and passes it on to the native `button` element. What's wrong with that approach? It's very explicit but other then that I dont't see any flaws. To the outside (users of this component) it doesn't make a difference, does it?

georgiee: Yes it works, but it's not idiomatic to React. That's the deal. I could ask instead why not use `type CopyProps = any`, spread the props and say that people don't need the types because buttons always have `onClick`. Still works and has no flaws.

Instead I went the route following the framework's rails and suggested to use the given button type that helps us to map react components back to the web standard so you can use the `<IconButton/>` as you would use any other `<button/>`. This resonates so much more with me then any custom mapping of `onClick` which creates more noise then the special type `ComponentPropsWithoutRef` I referenced before.

src/components/buttons/button.stories.mdx
milljoniaer: Due to some functionality changes, I have left out the whole sentence.

src/components/buttons/button.stories.mdx
milljoniaer: I had that in the previous story to let the user change the label of the button in the canvas. It was a leftover. I fixed it.

src/components/buttons/icon-button.tsx
milljoniaer: Thank you for your suggestions, I do like @georgiee idea and changed it, just like he suggested it. I hope that is fine for everyone.

src/components/buttons/icon-button.tsx
milljoniaer: Nice idea, way cleaner, I have done it like that 👍

milljoniaer:
Thank you for all your suggestions. I have now changed the following:
- I moved the whole `buttons` folder into the `new-components` folder
- The IconButton now passes a `show` property to the hardcoded Icon Component 
- I removed the unnecessary children property everywhere (especially in the story)
- I am now using `React.ComponentPropsWithoutRef<'button'>` and spread the props into the `<button>` 

Please have another look.

milljoniaer:
To make both buttons equal in use, I let also the RegularButtonProps extend React.ComponentPropsWithoutRef<'button'> and spreaded them over the native <button>.

I am not sure if that is the right PR for this. But in that way, both buttons are immediately equal in use.

src/new-components/buttons/icon-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+    background: #3e61ee;
+  }
+
+  &:active {
+    background: ${lighten(0.1, '#3e61ee')};
+  }
+`;
+
+interface IconButtonProps extends React.ComponentPropsWithoutRef<'button'> {
+  show: IconType;
```
georgiee: please use `icon`. We use `show` on the icon because it feels awkward to use `<Icon icon="..."`.

src/new-components/buttons/button.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+# Button
+
+There are multiple Button components for different situations.
+They all take all the props of a normal `<button>`, e.g. the `onClick` property.
+
+### Overview of Buttons
+
+---
+
+#### Regular
```
georgiee: Why h3 + h4 ? I would just go straight to `## Regular`

georgiee (approved):
Thanks, looks good. Let's go with that. I have some small little things. Please adjust but you get my approval already 🙌

PR #229, opened 2021-11-30
feat: landing page

### LINKED ISSUE #189: Page: Landingpage
Let's list here what components we need to update our landingpage. That way we have a clear goal which components we need to update. We don't even have to provide all of the updated components. I'm for example fine with the current navigation.

# List

+ ⏳ #173
+  #167 
+  #152 
+ Footer (missing story ?)
+ Some unclear ones. 

# Unclear
## This is NOT the header
This is a special canvas/stage for the landingpage.

Screen Shot 2021-11-17 at 06 59 19

## What's that as a whole ?
You can see that through the entire page but not in this specific shape (image, subtitle, headline, paragraph and some teasers). On other sites there is an illustration, no image or so. I don't think we should generalize this and let it leave only on the landingpage. I think we can deliver a component like this for the landingpage. Everything as a prop and the main content are the teasers?

```
<ContentHighlight
  image={SomeGatsbyImage}
  title="Kundenbeziehngen statt Kundenprojekte"
  link={<MyButton>Alle Kunden</MyButton>}
  subitle="Unsere Kunden">
  <Teaser/>
  <Teaser/>
  <Teaser/>
</ContentHighlight>

```

Screen Shot 2021-11-17 at 07 06 30

The particular parts are not existing too:
1. This is not a header block:

Screen Shot 2021-11-17 at 06 56 30

2. This is not an Image ContentBlock, isn't it ? There is no image credit in the lower corner, so is this a special version and we are good to use that component than?

Screen Shot 2021-11-17 at 06 57 47

# Source
https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3708%3A18248

<details>
<summary> Preview </summary>

d-home

</details>

feedm3:
This PR implements the new home page. 
URL: https://satellytescommain21751-featlandingpage.gtsb.io/

**Done**
- put Aurora effect into hero
- introduce new `TeaserGrid` component: [Storybook](https://deploy-preview-229--satellytes-website-storybook.netlify.app/?path=/docs/components-teasergrid--four-children&globals=locale:en;locales.en:English;locales.de:Deutsch)
- put Teasers and Callouts (HeaderBlocks) on the landing page

> I created a `LayoutV2` component instead of renaming or changing the existing `Layout` to keep the changeset small. I will refactor the `V2` components in a separate PR.

**Todo**
- [x] Add full width image -> Image not added, but a `FullWidth` container that can contain any element
- [x] Add `.mdx` to `TeaserGrid`
- [x] Remove `Grid` from layout
- [x] make `margins` and `paddings` pixel perfect
- [ ] put text into message bundles

Closes #189


georgiee:
Awesome we are getting closer 💫

georgiee:
Did you have time to work on this page? Do you need support? Thanks 💪

georgiee (approved):
ok let's go with this and fix the rest

PR #230, opened 2021-11-30
feat: add icon component

### LINKED ISSUE #224: Component: Icon
Provide a component to host our icons from the design using the same approach as `<Illustration/>` for a win.We don't have a lot of icons but it's better to distribute them through a single component anyway.

georgiee:
Close #224 

---

## Description
We finally have an `<Icon/>` component which means together with the `<Illustration/>` that we don't have to include any svg files directly anymore.

The component is fixed to a square size. The default size is `24x24px` but you can make it inherit the font-size. I chose to use a em size of 150% to make the icon 24px at a text size of 16px. That should reflect the usual ratio between the text and our icons. That's because our bounding box is pretty big, the actual shapes are smaller so we have to scale up a little bit.

There is an active discussion about the bounding boxes of the buttons. Team design doesn't want a fixed bounding box while we developers would like to handle icon through a fixed size. I think the developers trump here any decision as long as we can guarantee that the icons will look like in the design. 

I'm not sure how other icon libraries handle this. Font Awesome is based on a font, which means they do not have a fixed bounding box. They are embed in the text flow correctly at any time. Google Fonts is similar of course. Do we know any successful SVG based icon library to get some inspiration of the best settings? 

I chose a few settings for the beginning:
+ Fixed bounding box of 24px of the source SVG files.
+ SVG sources files inherit the color of any surrounding text
+ I apply a vertical align `middle` plus a display of `inlin-block` to align the icons. That is in most cases not enough as the text usually flows differently than our icon which is arbitrarily aligned within the bounding box. We have no way of makign it flow naturally as the bounding box kind of "swallows" the natural shape inside. That's also the reason why I placed a magic number for the vertical align in the first adaption I included: the button.

## Remarks
1. I had some issues with Storybook not applying correct controls while using the mdx document. Pretty annoying and that's when I stopped to improve the UX/DX with it. I wanted to include a filter, changing colors (to proof they adapt) and such but it's too much fighting and I think we can come back once storybook stabilizes (well I hope it gets stable since version 1 😬).
2. I did not include every icon as I had to exclude those with a bounding box different then 24x24px. Once available we can add them:
3. While adapting the button I found that we have to update the button, it's API with the `onClick` is not good and we should spread the props instead.


src/components/accordion/accordion-header.tsx
```diff
@@ -23,15 +24,6 @@ const AccordionTitleHeadline = styled.span`
...
 
-const resetButton = css`
-  padding: 0;
-  margin: 0;
-  appearance: none;
-  background: 0;
-  border: 0;
-  box-shadow: none;
-`;
-
```
georgiee, line 34: I chose to make this reset available for sharing to use in other button-like components.

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
-  border-radius: 30px;
-  border: none;
 
   &:hover {
     background: #3e61ee;
   }
+
+  &:active {
+    background: ${lighten(0.1, '#3e61ee')};
+  }
```
georgiee: I'm sorry, this might be beyond this PR of introducing the icon. I started adapting the button to make it work with the icon component and I got carried away a little bit by adding some active state and the cursor. Those are minimal changes and I didn't want to revert that useful stuff here.

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
   background: linear-gradient(275.41deg, #543fd7 0%, #2756fd 100%);
-  border-radius: 30px;
-  border: none;
 
   &:hover {
     background: #3e61ee;
   }
+
+  &:active {
+    background: ${lighten(0.1, '#3e61ee')};
```
georgiee, line 23: This lighten helper from polished is my simple way of defining some lighter color here. That `active` state (when mouse is down on the button) is not defined in the design and I don't want to postpone every single small decision until the design team delivers. Here a best guess to be fixed and aligned later 👍

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
+  }
 `;
 
-const StyledIcon = styled.span`
-  padding-left: 24px;
-  padding-right: 16px;
+const StyledIcon = styled(Icon)`
+  padding-right: 10px;
+  // pull the fixed size icon closer to the text
+  margin-left: -4px;
```
georgiee, line 30: Adjustments like that are what we have committed for when we chose the fixed bounding box.

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
+  /**
+    This is a magic number to match the vertical middle of the button
+    with the icon which has a fixed bounding box which won't allow us to align
+    it based on the actual shape of the icon.
+   */
+  vertical-align: -18%;
```
georgiee, line 36: Same here, the bounding box doesn't allow us to let the shape of the icon "flow" with the text. If we had a proper shape without huge amounts of spare space we could use the different modes of vertical-align. Icon fonts can do that because we basically have no bounding box per se (but usually they get one from the css of the icon framework)

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
 `;
 
 const ButtonText = styled.span`
   padding-left: 16px;
-
-  ${TextStyles.toplineR}
```
georgiee, line 25: I chose to remove this as it includes a line height and I moved it to the root of the component so that the icon can inherit the size and color aspects.

src/components/buttons/button.tsx
```diff
@@ -35,22 +51,7 @@ const Button = ({ children, onClick }: RegularButtonProps) => {
...
-        >
-          <path
-            d="M2 2L5 6L2 10"
-            stroke="white"
-            strokeWidth="2.5"
-            strokeLinecap="round"
-          />
-        </svg>
-      </StyledIcon>
+      <StyledIcon show={'chevron_right'} size={IconSize.NORMAL} />
```
georgiee, line 53: deleting code is the best discipline 😁

src/components/icon/icon-preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+`;
+
+const Name = styled.span`
+  font-weight: bold;
+  margin-top: 10px;
+  ${ellipsis()}
+`;
+
+const IconPreviewLayout = styled.div`
+  aspect-ratio: 1;
```
georgiee: that forces the preview to have a square layout in the grid, so good 🌟

src/components/buttons/button.tsx
fischeversenker: I'd like to point out that [SCSS also provides this](https://sass-lang.com/documentation/modules/color#scale). No need for an additional library.
I know that we already have `polished` as a dependency so might as well use it. I just think it's worth knowing that SCSS has an insane amount of functionality built in.

src/components/buttons/button.tsx
fischeversenker: What's the point of having these helpers then? Why not use it and overwrite what you need to overwrite, like the `line-height` instead of copying all the styles from this helper?

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
 
 const StyledButton = styled.button`
-  height: 40px;
+  ${resetButton};
+  color: #ffffff;
+  font-size: 16px;
+  font-weight: bold;
+  line-height: 40px;
+  border-radius: 30px;
+  vertical-align: middle;
```
fischeversenker: This will make sure that the button, if used as an inline element, will be middle-aligned. It does not affect the icon. Was that your intention?

src/components/buttons/button.tsx
```diff
@@ -1,29 +1,45 @@
...
+  cursor: pointer;
 
   background: linear-gradient(275.41deg, #543fd7 0%, #2756fd 100%);
-  border-radius: 30px;
-  border: none;
```
fischeversenker, line 10: Why did you decide to move these two properties up between the other props but leave the `background-color` down here, separated by a new line? Is there a reason for this or just personal preference? The newline confuses me (I know that it existed before - it was equally as confusing back then :joy:) and the change itself makes the review more complicated than it would need to be.

src/components/buttons/button.tsx
georgiee: That's irritating first but styled components does not provide scss ☝️ It supports a small set of feature like the nesting only. That's why polished is offered by the maintainers of styled components.

src/components/buttons/button.tsx
georgiee: I thought the button component is a very special component but you might be right, I got carried away a little bit.

src/components/icon/icon.stories.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+
+## Usage
+The icon can be used in various situations. Here some examples with buttons and inside a text.
+<div style={{margin: '20px 0'}}>
+  <Callout illustration={'scientist_042'}>
+    This is not a component that magically works everywhere.
+    You want to adjust the icons in particular situations to fit the given design expectations.
+  </Callout>
+</div>
```
fischeversenker: While reading the documentation I got very confused by this. Do we really want to use component A in the documentation for component B? I don't think so. I suggest to use a different technique to highlight this paragraph. Maybe a quote works here (`> ...`)?

src/components/icon/icon.stories.mdx
```diff
@@ -0,0 +1,88 @@
...
+### With Buttons
+
+<Canvas>
+  <Story name="With Buttons">
+    <div>
+      <button>
+        <Icon show={'theme_switch'} size={IconSize.NORMAL} />
+      </button>
+      <button>
+        <Icon show={'burger_nav'} size={IconSize.NORMAL} /> Menu
```
fischeversenker: As far as I understood, `IconSize.NORMAL` is the default. Why would we want to explicitly specify this here, then? It makes the code for this story more confusing then it would need to be in my opinion.

src/components/icon/icon-preview.tsx
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components';
+import { Icon } from './icon';
+import React from 'react';
+import { ellipsis } from 'polished';
+
+export const PreviewGrid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(auto-fit, 150px);
+  gap: 10px;
+`;
```
fischeversenker: I was confused why this is part of this file. Why not define this in the `stories` file? I'm torn... I do like the idea to not define any (styled) components in the `stories` files but here it feels like an artificial separation.

src/components/icon/icon.tsx
```diff
@@ -0,0 +1,76 @@
...
+   */;
+  size?: IconSize;
+  /**
+   * Let's support styled components and friends by accepting this property
+   */
+  className?: string;
+}
+
+const transformSizeToDimension = (size: IconSize) => {
+  if (!Object.keys(ICON_SIZE_MAP).includes(size)) {
```
fischeversenker: Do we really need this `if`? Through the power of the typescript compiler I feel like we should be perfectly safe here...

src/components/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from 'styled-components';
+
+export const resetButton = css`
+  padding: 0;
+  margin: 0;
+  appearance: none;
+  background: 0;
+  border: 0;
+  box-shadow: none;
+`;
```
fischeversenker: I like this. I'm not sure about the file name, though. Maybe we can be more explicit? `shared-styles`, `css-collection`? I can't think of nice examples so we might as well leave it as `utils.ts`... it was just a thought.

fischeversenker (commented):
Thanks for making sure that we are using harmonized icons and integrating it into the `button` already.
I left a few suggestion, some of them I'd like to see resolved before we merge this.

src/components/buttons/button.tsx
fischeversenker: Thanks for the info. This is in fact irritating... We don't need new standards at every corner 😭

src/components/buttons/button.tsx
georgiee: Leftover from debugging. Thank you catching that. I will remove it.

src/components/buttons/button.tsx
georgiee: I guess my mind saw this neat little package of things dealing with the background😬 I remove the extra line and promise to think of you the next time I'm on the way doing such a reordering because I understand that it makes the review more complex.

Screen Shot 2021-12-01 at 07 29 21

src/components/buttons/button.tsx
georgiee: You will get the hang of it. I'm actually happy there is no scss implemented but that they offer the ampersand 🙏 Imagine they invented yet another token to replace the ampersand. That would be bad.

src/components/buttons/button.tsx
georgiee: I have brought this back now. Thanks for pointing out 👍

src/components/icon/icon.stories.mdx
georgiee: Oh really? I did that with good intentions. That callout will be used to highlight information in the blog post later so why not use this awesome component for our own purpose? I mean yes, when opening this discussion people will ask why not picking up the text styles, headlines and such but it was pretty easy and neat for this one so I think I was happy trying it out. I never expected to cause any confusion with that. I will remove it for now and replace it with a markdown blockquote as suggested

Screen Shot 2021-12-01 at 07 36 47

src/components/icon/icon.stories.mdx
georgiee: You are 100% right. I switched the default mid-implementation so I had explicit assignments for both worlds at some point. I removed these ones. Thanks!

src/components/icon/icon.tsx
georgiee: I had an exception while working on it that's when I did that but actually I should have though like you and ask why is TS not covering my ass. That's the reason:

```
const ICON_SIZE_MAP = {
  [IconSize.NORMAL]: {
    width: '24px',
    height: '24px',
  },
```

This object is not typed even though the only keys are enums. We need to do it like that for enums:

```
type IconSizeMap =  {[key in IconSize]: Dimensions};

const ICON_SIZE_MAP: IconSizeMap = {
```

I have inserted this to make sure values are correct. Thanks.

src/components/utils.ts
georgiee: I didn't want to spend on naming that thing yet. I took some time now to think about it and `css-helpers.ts` it is now. I hope that suits you better, I like it more than the generic utils barrel

georgiee (commented):
Thanks @fischeversenker for your review. I handled and commented every single comment of you and I think we are good know 💪

src/components/icon/icon-preview.tsx
georgiee: I just want to lock away the stuff I wrote to support the story. This gets more important if we write mdx files as this is for sure something I won't write in mdx files for example.

src/components/icon/icon.stories.mdx
fischeversenker: I feel like the documentation is a lot cleaner if there is really only a single one of our components involved. I appreciate your understanding 🙏

src/components/icon/icon-preview.tsx
fischeversenker: Got it. Fine for me 👌

fischeversenker (approved):
Thanks for the updates. Looks solid now 👍 ✔

Regarding your review:
Don't include comment responses as part of your review. They will appear without any reference to the original comment in your review:
image

I very much recommend to adding comment responses as "Add single comment" instead of as part of a review 🙏

georgiee:
@fischeversenker I have seen that effect too. Pretty stupid without the reference. I though I save you a few mails by batching them through the review. Next time I stick to the default way of commenting

fischeversenker:
Maybe you can spend another minute on this PR so that the new component is added to `/new-components` instead of `/components`?

georgiee:
Yes I will do that.

src/new-components/icon/icon.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
 import { Overview, Regular } from './icon.stories';
-import { Icon, IconSize } from './icon'; import Callout from '../callout/callout';
+import { Icon, IconSize } from './icon'; import Callout from '../../components/callout/callout';
```
georgiee: unused and my IDE hid it in that line 😁

georgiee: removed

PR #231, opened 2021-12-01
refactor: group textfield & checkbox in category `forms`

georgiee:
I did a little refactoring exercise to improve our structure in Storybook.

+ Introduce folder `new-components` so we can collect new ones or updated ones. I only put the refactored ones in there but we should continue doing so.
+ Renamed `Textfield` to `TextInput` to be more specific.
+ Introduced the mdx format. I was able to remove the entire `*.ts` file 👍 Spot the `mdx` column in the [docs](https://storybook.js.org/docs/react/essentials/actions). I never noticed it but it helps dramatically to write code for mdx (right now we let both files coexist which we often don't need)
Screen Shot 2021-12-01 at 07 01 58

+ I added a note ` ⚠️ API not usable yet, design focused only` to the two components as they are really not usable. We discussed that matter and said it's okay to learn for our folks but we should not forget to tackle them.

+ There are plenty of inline styled components which is wrong and generates plenty of warnings. I have pointed out that problem in a later component implemented by @milljoniaer (see https://github.com/satellytes/satellytes.com/pull/195/files/9572c9238cb2b629d9dc04953b412443f421f5a8#r753180436). I suggest once this PR is merged we fix this issue immediately 🙏 It's a good practice to reflect on gained knowledge by thinking about your own past components like the two in this PR.

+ We don't have an issue to create the different form controls. For example `TextArea` is missing. I'm going to create an issue so we don't forget them.


feedm3:
Just today I have talked with Felix about how we don't like to setup 2 files for documentation 😄  Only having `.mdx` is really convenient, looking forward to seeing our design system getting better and better!

feedm3 (approved):

fischeversenker (approved):
Looks great! Let's get this merged so the other open PRs can also feed into `new-components` asap 👏

Issue #220, opened 2021-11-29
HeaderBlock naming & usage issue: split into two components


georgiee:
This is not an urgent issue, but I think it's worth the discussion in general. From the design I can see the `HeaderBlock` being used in three ways:

## 1. As standalone content

![Screen Shot 2021-11-29 at 06 57 40](https://user-images.githubusercontent.com/1701755/143816183-84be0211-e16a-49bf-b42d-f31b65749f1a.png)

## 2. As an "introductionary" for teasers

![Screen Shot 2021-11-29 at 06 58 08](https://user-images.githubusercontent.com/1701755/143816286-2dbee6c2-d373-4b83-8b6f-8feba4df6386.png)

## 3. As an actual "header block" in terms of being a header of an article
![Screen Shot 2021-11-29 at 06 59 16](https://user-images.githubusercontent.com/1701755/143816317-e107445f-4fe8-4567-9fca-8697dfb6b85d.png)

This kind of blurs the scope for the user and makes the naming especially hard. `HeaderBlock` does work right now because it helps talking about the same things and the designers name it like that. It's a wrong name for me though. 

+ First of all, everything in the design, almost everything got the postfix `block` in the desing anyway. It's obsolete because we are talking about content parts that we want to reuse. It's blocks but also components or whatever category we would like to choose for them. It doesn't need to be part of the name.
+ Yeah, it's something like a header as much as the [`<header/>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) is a header for almost everything because you have to decide: Page Header, Article Header, whatever header.

# Suggestions
+ I think `ArticleHeader` defines the original use case (see Nr. 3) the best and it should be based on an actual `header` tag like so. There is never an illustration.

```
<article>
  <header>
    <span>kicker</h2>
    <h2>headline</h2>
    <p>byline/subtitle</p>
    <p>lead/first paragraph</p>
  </header>
 <!-- here the rest of content -->
  
</article>

```

+ The content ones outside of a header deserve their own component. It's like being `SectionHeader` or `ContentHeader` with the following markup. There is never a byline/meta text to describe the author.

```
  <header>
    <span>kicker</h2>
    <h2>headline</h2>
    <p>lead/first paragraph</p>
    <illustration/>
  </header>
```

[...]

PR #233, opened 2021-12-01
feat: add blog header

georgiee:
Relates to #220 

---

This is part 1 of 2 refactoring the `HeaderBlock` into two distinct components. Here the Blog Header to be used exclusively in the blog as the header (hence the name).

I basically took the existing implementation, removed the obsolete parts and refactored the rest. I updated the naming following the guide pdf in the linked issues (kicker, byline, leader).

I skipped the `.stories.ts` file as we should use not both but one of the story formats which works pretty good.

I like how the component feels from the API usage and I was able to make the implementation more compact. Thanks @milljoniaer for the existing implementation, I was able to pick some styles easily 👍

## Changes
+ I focused on html tags. Something like that `const Headline = styled.p` is really bad from an accessible view and we should not let thigns like that slip into our codebase 🙏 In addition I based the entire component on a `header` tag plus `h1` to provide a correct markup in the context of blog posts. This also means this component is not meant to be used somewhere else.

+ I removed the entire exception handling of the `large` display. This is not relevant at all for the blog article and I will take a very careful look on this when working on part 2.

+ renamed the different parts (meta -> byLine, topline -> kicker and such) which feels much more solid now 💪
+ I moved some margins around so we don't end up with obsolete margins if an element is removed (e.g. Topline -> Headline, margin 16px is now on the topline)
+ I found a color with alpha 50%. That's not good we do not work with alpha values in text. I have normalized it to a hex value without alpha and commented in Figma.

Part 2 will cover the content-related component with headline + icon + text. I hope you see how divide and conquer helps providing two very useful components instead of one big blob of features that correlate in markup but not semantically (a blog header is never a content section).


src/new-components/blog-header/blog-header.tsx
```diff
@@ -0,0 +1,61 @@
...
+`;
+
+const HeadlineStyled = styled.h1`
+  ${TextStyles.headlineXL}
+  margin: 0;
+  color: #202840;
+  margin-bottom: 16px;
+`;
+
+const BylineStyled = styled.span`
```
hofmann-felix, line 39: Typography does not match the design. Applying the timestamp text style should solve it: 

```suggestion
const BylineStyled = styled.span`
  ${TextStyles.timestamp}
```

src/new-components/blog-header/blog-header.tsx
```diff
@@ -0,0 +1,61 @@
...
+`;
+
+const LeadStyled = styled.div`
+  ${TextStyles.textL}
+  margin-top: 48px;
+
+  ${up('md')} {
+    margin-top: 32px;
+  }
+`;
```
hofmann-felix: The design uses another font style on mobile: 

```suggestion
const LeadStyled = styled.div`
  ${TextStyles.textR}
  margin-top: 48px;

  ${up('md')} {
    margin-top: 32px;
    ${TextStyles.textL}
  }
`;
```


Edit: The headline is also smaller on mobile (Headline L instead of XL)

hofmann-felix (changes requested):
Useful changes, especially regarding the accessibility and readability of the code 👍 In two places the font does not fit.

fischeversenker (approved):
I agree with the changes that @hofmann-felix requested. Apart from that I'm approving this.

src/new-components/blog-header/blog-header.tsx
georgiee: thanks for the catch and the helpful tip 🙏

georgiee:
@hofmann-felix I did not notice at first but you made suggestion for the actual fixes. Thank you for that, this really helped moving forward while being busy with other things. I merged them and I'm happy to merge once approved.

hofmann-felix (approved):

PR #234, opened 2021-12-01
feat: add section header

georgiee:
Relates to #220 

this component does not provide the illustration. There will be a third component derived from the header block for this.

---

Following the blog header from #233 we are introducing another type of header for our general content on the page. Use this component to open new sections including the start of a page. 

I found that semantic behavior while browsing through the figma designs. I have marked this component where it's clearly opening up a new section.

<details>

Screen Shot 2021-12-01 at 17 33 52

</details>

Other parts of the design look very similar to this component:
+ The ones with the illustration in "Services"
+ The ones in "Kunden"

The both share the fact that they belong to a single section on the page and they act as content. So they are clearly no section headers although they look really similar. 

Some technical details:
1. I have picked a headline of `h2` as it's a great fit for the general page outline. You should only have a single h1 per page and that's the one we will place in our header area through other means.
2. I have picked a `div` tag to wrap this component. First I chose a `section` but I'm not sure about the semantics. The a11y document outline reflects both div and section as `sections`. [Some more reading on this](https://www.scottohara.me/blog/2021/07/16/section.html)
3. This component deliberately does not contain any illustration which is only available on the content variants described before.

## Usage

I imagine a markup like this (here the index page as an example). You can see how the section header works really good to open a new section to fulfill its semantic.

```

<Main>
  <h1>Satellytes</h1>

  <SectionHeader headline="Development & Design" [...]>
  <SomeTeaserGroup/>
   <Button/>

  <Image/>

  <SectionHeader headline="Clients [...]>
  <SomeOtherTeaserGroup/>
   <Button/>

  <Image/>

  <SectionHeader headline="Work with us" [...]>
  <YetSomeOtherTeaserGroup/>
   <Button/>

  <Image/>

  <SectionHeader headline="Ping" [..] >
  <EvenMoreTeaserGroups/>
   <Button/>
</Main>

```


src/new-components/section-header/section-header.tsx
```diff
@@ -0,0 +1,55 @@
...
+    ${TextStyles.textL}
+  }
+`;
+
+export const SectionHeader = (props: BlogHeaderProps) => {
+  return (
+    <div>
+      {props.kicker && <KickerStyled>{props.kicker}</KickerStyled>}
+      <HeadlineStyled>{props.headline}</HeadlineStyled>
+      <ContentStyled>{props.children}</ContentStyled>
```
fischeversenker, line 52: We need to have a naming convention for these styled components. It's `ContentStyled` in some places and `StyledContent` in others. I prefer using `Styled` as a prefix (e.g. `StyledContent`) because it feels more natural while reading ("styled content" vs. "content styled").
In the end I'm fine with both, but let's be consistent.

fischeversenker (approved):
I like it. Nice and simple :sparkles:
Approved :+1:

In your PR description you mention that the sections on the "Kunden" page are not real sections. I'd like to disagree.
Looking at your screenshot and the designs they very much look like sections to me. What's your reasoning here? What constitutes a "section"?
image

Also, in your "Usage" example I would have loved to see some actual `<section>` elements, something like this:
```html

<Main>
  <h1>Satellytes</h1>

  <section>
    <SectionHeader headline="Development & Design" [...]>
    <SomeTeaserGroup/>
    <Button/>
  </section>

  <Image/>
  
  <section>
    <SectionHeader headline="Clients [...]">
    <SomeOtherTeaserGroup/>
    <Button/>
  </section>
</Main>
```

src/new-components/section-header/section-header.tsx
georgiee: it's just the rule I'm missing, my mind constantly goes around the available possibilities as I can't remember a single way 👍

georgiee:
@fischeversenker 

When I looked through the different parts I recognized why we previously introduced `large` as an option. The header variant of the HeaderBlock always have `Text L` applied while the ones below (Kunden & Services) do have `Text R`. I refactored the ones with Text R into `Blog Header` and `Section Header`. 

The rest of the variants use Text R which indicates to me that they might belong to a similar group and indeed, see this.

# Services
SectionHeader openes with "Development & Design" and then one level into this you find the actual content: `Development`, `Reviews` & `Design`.

The same happens on "Kunden":
SectionHeader opens the page with "Kundenbeziehungen.." and then the content follows: the four different clients. They are clearly no sections but content to me. That's why I concluded that they are content and not sections which would open the document outline for yet another level.

Actually those headers (Blog & Section) are good what I don't like are exactly the content things we are talking about now. I would rather see single headlines & paragraph, no need to group them so strictly anyway? You could provide a header component to provide illustrations, different headline sizes or so but really not entire blocks. You can create them as local components to have `ServiceSection`, `ClientReference` and such later.

 I think it's easier to speak about that topic though.

PR #236, opened 2021-12-02
fix: apply font style to button and select

### LINKED ISSUE #235: Font issues in button and dropdown
Our font is not applied to `<button />` and `<select />` in all components:
Bildschirmfoto 2021-12-02 um 09 54 02

One possible fix I found on the fly would be to set the font explicitly for `select` and `button` in `global-styles.ts`:
```
  button, select {
    font-family: ${(props) => props.theme.typography.fontFamily}, 
                "Helvetica", 
                "Lucida Grande", 
                Tahoma, 
                Sans-Serif;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
  }
```

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/235
Storybook: https://deploy-preview-236--satellytes-website-storybook.netlify.app

**What was the problem?** 
- The font was not applied to the `<button />` and `<select />` tags in all components

**How it is solved:** 
- In gloabl-styles.ts the font is now explicitly applied to select and button 

**Preview:**
Before: 
Bildschirmfoto 2021-12-02 um 12 05 22

After: 
Bildschirmfoto 2021-12-02 um 12 05 07


src/components/layout/global-style.ts
```diff
@@ -21,6 +21,16 @@ export const GlobalStyle = createGlobalStyle<{ $lightTheme?: boolean }>`
...
+
+  button, select {
+    font-family: ${(props) => props.theme.typography.fontFamily},
+    "Helvetica",
+    "Lucida Grande",
+    Tahoma,
+    Sans-Serif;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+  }
```
fischeversenker: I suggest to include `input` and `textarea` here as well. Also, instead of duplicating the font definition it's enough to specify `font-family: inherit;`. The font-smoothing properties are already inherited.
```suggestion
  button, select, input, textarea {
    font-family: inherit;
  }
```

fischeversenker (changes requested):

src/components/layout/global-style.ts
hofmann-felix: @fischeversenker Thanks for your review 🙏 It looks much clearer now

fischeversenker (approved):
Thanks for the update 🙏

PR #237, opened 2021-12-02
feat: add translation keys to landing page

hofmann-felix:
What's included?
- translation keys are added to the landing page
- Headlines/Toplines are translated into English


fischeversenker (approved):
I've started the app locally and had a quick look over the texts. It looks great so far.
I don't like that the index page translations are stored under `main`, but that's not in the scope of this PR, so I'm happily approving 🙌

PR #238, opened 2021-12-02
feat: add new services page

### LINKED ISSUE #207: Page: Services
Text is missing. I need to write a draft for the technical ones. 

+ Development/Softwaretwicklung
+ Beratung & Technische Analyse 
+ UX/UI

hofmann-felix:
URL: https://satellytescommain21751-featservicespage.gtsb.io/services/

## What's included?
- new services page
- mainly `SectionHeader` and `HeaderBlock` were used
- As in `index.tsx`, the vertical spacing is established via `<ContentBlockContainer />`.

## Notes
- Currently, the text for `HeaderBlock` and `SectionHeader` is taken unformatted from `translations.json`. I tried to give the text as markdown directly into `translation.json` and then parse it. However, this was a bit more complicated than I thought, so I suggest a separate PR for it
- There are some issues with the font style of the `HeaderBlock` on mobile. To keep the services page PR clean, I would also suggest a separate PR here 

## Preview
Bildschirmfoto 2021-12-06 um 11 31 27


feedm3:
Thanks for the services page!

I have 2 small things:

1. The leadbox icon doesn't work


2. In the design there is an example with Markdown. Can you try to add markdown as text and see if it works? If it is too much effort to make it work, let's split it up into a separate PR.

src/locales/en/translations.json
```diff
@@ -131,7 +131,28 @@
   },
   "services": {
     "title": "Services",
-    "subheading": "Satellytes – passionate developers and designers.",
+    "header": {
+      "title": "Development & Design",
+      "kicker": "Our Services",
+      "text": "Wir haben großen Spaß an Technologie und freuen uns auf neue Herausforderungen. Dabei fokussieren wir uns auf langfristige Engagements im Konzerngeschäft.\nNeuen Aufgaben begegnen wir immer mit angemessenem Respekt. Wir streben stets hochwertige und zeitgemäße Lösungen an – die Wahl der Technologie ist für uns dabei sekundär. Gern unterstützen wir auch Sie in den folgenden Disziplinen:"
```
georgiee, line 114: I really don't like to put so much text in translations. It just cries for markdown files or such. But it's the correct thing to do for this PR 👍

src/pages/services.tsx
```diff
@@ -36,23 +55,59 @@ const ServicesPage = ({ data, location }: ServicePageProps) => {
...
+          <>
+            <Aurora />
+            <HeroContainer>
+              <ServicesPageTitle>{t('services.title')}</ServicesPageTitle>
+            </HeroContainer>
+          </>
+        }
+      >
+        <ContentBlockContainer>
+          <SectionHeader
```
georgiee, line 73: Thanks for using the section header 👍

src/pages/services.tsx
```diff
@@ -36,23 +55,59 @@ const ServicesPage = ({ data, location }: ServicePageProps) => {
...
+            {t('services.header.text')}
+          </SectionHeader>
+        </ContentBlockContainer>
+        <ContentBlockContainer>
+          <HeaderBlock
+            headline={t('services.contentblocks.first.title')}
+            illustration="rocket_011"
+            large
+            topline={t('services.contentblocks.first.kicker')}
+          >
```
georgiee, line 86: I have created the section header and blog header from the  header block. What's missing is this part. It's okay for this PR but my main goal is to solve the question how to provide larger blocks of text & headlines which will include this part for sure.

I see that this special arrangement of headline + illustration and text is only available on services, so a local component could work wonder here before creating some shareable component. You would be very fast to assemble the different components without the story overhead for example.

georgiee (approved):
I'm fine with that. Let's merge. I'm writing/updating the text content so we might change things (content wise) here soon anyway so I would like to move forward. Thanks!

PR #239, opened 2021-12-03
refactor(leadbox): use illustrations and update data structure

georgiee:
I have this PR in the pipeline for quite some days and I finally got all pieces together. I started with integrating our illustration component but I found some issues to address:

+ The spacing to the left is not existent while it should be 108px, so I updated the css layout
+ The component includes a lot of margin and padding to create the illusion it's behind the footer. That should not be part of the component. Now I provide `LeadboxFooterContainer` to use in our layout files but we need to make sure to move that responsibility into the layout soon.
+ I have updated the props structure to have dedicated fields for `contact` or `link` instead of mixing both scenarios together.
+ I have created two stories to showcase the two cases and removed the ts story file.
+ I moved the component into the `new-components' folder.
+ I updated our pages (currently no leadbox is visible in the production for some reason ?)

I hope the changes make sense👍


georgiee:
Before/After in the story.

Screen Shot 2021-12-03 at 13 11 22
Screen Shot 2021-12-03 at 13 11 26

src/pages/clients.tsx
```diff
@@ -10,6 +10,7 @@ import { MarkdownAst } from '../components/markdown/markdown-ast';
 import { LocalesQuery } from './index';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { SpaceShuttle } from '../components/icons/illustrations/space-shuttle';
```
feedm3: Here are 2 unused imports:

Screenshot 2021-12-06 at 08 01 52

src/templates/blog-post.tsx
```diff
@@ -16,7 +16,7 @@ import { LocalesQuery } from '../pages';
 import FollowPanel from '../components/social-panel/follow-panel';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { Astronaut } from '../components/icons/illustrations/astronaut';
```
feedm3: Also an unused import.

feedm3 (approved):
Look very good 👍 

Besides the 3 unused imports I don't have anything to complain 🙂 I approve in advance.

src/pages/clients.tsx
georgiee: I learned to ignore the github stuff from the times we had the errors in our repo. I know that you cleaned that up so I promise to check them better. Thanks for the catch 👍

src/templates/blog-post.tsx
georgiee: I can tell my IDE to clean up. I think I should do this on save of even better we have some commit hook for things like that too. Thanks for the catch here too.

PR #241, opened 2021-12-06
chore: remove clients

georgiee:
this will remove the clients by doing the following. 
Our expectations: Clients will come back with a real portfolio or not. That's why we can remove everything related.

+ remove obsolete client components not even used currently
+ remove any translation related to client
+ remove navigation entries
+ remove clients markdown content
+ remove index section
+ remove clients data json (not even used right now)

in addition I saw and fixed:
+ scroll bar on index on the teaser grids (cc @feedm3 )
+ wrong links on the service teasers (went to clients instead of services cc @feedm3 )

## Remarks
1. The current branch locally show no career job posting. I think this is not related to this PR and might be some problem with the dependency update or just locally a hiccup ?
2. The index page doesn't reference the office page or the blog right now. I think this is also something we should keep track of.


fischeversenker (approved):
It's always nice to see code being deleted. In this case its a little sad because it's an entire page of our website... but I get the point. It will probably return one day, shinier than ever :sparkles:

I couldn't find any leftover reference to the clients page so this looks great and I'm approving 🙌 

One small thing, that is not even related to this PR:
There is an unused import in `office.tsx`. Since you already take care of 2 unrelated things in this PR, maybe a third one doesn't hurt?
image

georgiee:
I updated `office` as suggested and enabled auto merge for this PR now, thanks 👍

PR #243, opened 2021-12-06
feat: add dropdown component

### LINKED ISSUE #225: Component: Dropdown
Currently, there are several dropdowns (e.g. in the header, in the pagination and in the FileDropper) on the page, which have all been built with the native `select` element. However, this element has the following limitations:
- The content of the expanded dropdown and the active value is always the same. For the header it would be good to have "Deutsch" and "English" in the expanded state and only the abbreviation "DE" or "EN" in the collapsed state. 140717922-1b9a6f1c-6aa6-46fa-85ca-0f3b748590b1
- The dropdown always has the width of the longest value. For the FileDropper it would be practical if the length always adapts to the active value. 
Bildschirmfoto 2021-11-29 um 14 46 13
Bildschirmfoto 2021-11-29 um 14 46 22

Ideas to solve the problem:
- own reusable dropdown component
- @georgiee suggested this library: https://reach.tech/listbox

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/225
Story: https://deploy-preview-243--satellytes-website-storybook.netlify.app/?path=/docs/components-form-dropdown--regular
Preview: https://satellytescommain21751-featdropdown.gtsb.io/

## What's included?
- new dropdown component based on the [Listbox of Reach UI](https://reach.tech/listbox/)
- add new dropdown to existing components ([language switch in the header](https://satellytescommain21751-featdropdown.gtsb.io/), [pagination](https://deploy-preview-243--satellytes-website-storybook.netlify.app/?path=/docs/components-pagination--with-mock-data) and [file-dropper](https://deploy-preview-243--satellytes-website-storybook.netlify.app/?path=/docs/components-form-filedropper--regular))
- apply some universal styles which are needed in every dropdown (e.g remove border, cursor = pointer, ...)

## Why is the native `<select />` not enough?
- The content of the expanded `select` and the active value is always the same. For the header it would be good to have "Deutsch" and "English" in the expanded state and only the abbreviation "DE" or "EN" in the collapsed state. That's now possible: 
Bildschirmfoto 2021-12-06 um 16 34 13

- The `select` always has the width of the longest value. For the FileDropper it would be practical if the length always adapts to the active value. That's now possible too: 
_Before:_ Bildschirmfoto 2021-12-02 um 11 30 29
_Now:_ Bildschirmfoto 2021-12-06 um 16 27 13 Bildschirmfoto 2021-12-06 um 16 27 19

## Notes
- The expanded dropdown (= `ListboxPopover`) has the standard Reach UI style yet. Maybe we have to adjust this style in the future.


georgiee:
this looks super promising 🌟

src/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,73 @@
...
+  cursor: pointer;
+  padding: 0;
+`;
+
+const StyledListboxButton = styled(ListboxButton)`
+  border: none;
+  cursor: pointer;
+`;
+
+interface DropdownProps {
```
feedm3: See #228: It should be possible to extend the react components interface and have all default `Select` props at once: 
```
interface DropdownProps extends React.ComponentPropsWithoutRef <'select'> {
  ...
}
```

Maybe I'm wrong because react-ui handles that differently though...

hofmann-felix: `React.ComponentPropsWithoutRef <'select'>` doesn't work, because the reach ui listbox uses other props/types than the native `select` element. 
Reach UI provides type definitions for some components, unfortunately I didn't find any for the listbox

src/components/dropdown/dropdown.stories.mdx
```diff
@@ -0,0 +1,27 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Dropdown, DropdownOption } from './dropdown';
+import { Icon } from '../../new-components/icon/icon';
+
+<Meta
+  title="Components/Dropdown"
```
georgiee: I would prefer the `Forms/` bucket, what do you think?

package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@reach/accordion": "^0.16.1",
+    "@reach/listbox": "^0.16.2",
```
georgiee, line 35: 👍

src/components/dropdown/dropdown.stories.mdx
```diff
@@ -0,0 +1,27 @@
...
+# Dropdown
+This component is used to build more flexible dropdowns. It is based on the [Listbox of Reach UI](https://reach.tech/listbox/).
+
+<Canvas withSource="open">
+  <Story name="Regular">
+    <Dropdown defaultValue='default' arrow={<Icon show={'chevron_down'}/>}>
+        <DropdownOption popoverLabel={"Choose an option"} value='default' disabled>Choose an option</DropdownOption>
+        <DropdownOption popoverLabel={"Option 1"} value='option1'>1</DropdownOption>
+        <DropdownOption popoverLabel={"Option 2"} value='option2'>2</DropdownOption>
+        <DropdownOption popoverLabel={"Option 3"} value='option3'>3</DropdownOption>
```
georgiee: I think `popoverLabel` is your naming? It feels not good almost counter intuitive. It's the value displayed in the select when not opened right? I think we can do better. Can you propose a few names?

src/components/dropdown/dropdown.stories.mdx
```diff
@@ -0,0 +1,27 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Dropdown, DropdownOption } from './dropdown';
+import { Icon } from '../../new-components/icon/icon';
+
+<Meta
+  title="Components/Dropdown"
+  component={{ Dropdown }}/>
```
georgiee: storybook has the concept of `subcomponents`. Is this relevant here? I don't know, because actually you can just pass  `DropdownOption` to the args table anyway?

https://storybook.js.org/docs/react/workflows/stories-for-multiple-components

src/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+  ListboxOption,
+} from '@reach/listbox';
+import '@reach/listbox/styles.css';
+import styled from 'styled-components';
+
+const StyledListboxInput = styled(ListboxInput)`
+  cursor: pointer;
+`;
+const StyledListboxPopover = styled(ListboxPopover)`
+  z-index: 200;
```
georgiee: You should really not pick random numbers. Do we have other z indices already? Otherwise would a `1` or `2` work already?

src/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+  ListboxButton,
+  ListboxPopover,
+  ListboxList,
+  ListboxOption,
+} from '@reach/listbox';
+import '@reach/listbox/styles.css';
+import styled from 'styled-components';
+
+const StyledListboxInput = styled(ListboxInput)`
+  cursor: pointer;
```
georgiee: So there is no pointer by default. Maybe that's okay? Can you compare with the native select? Thanks.

src/components/dropdown/dropdown.tsx
georgiee: Maybe you can dig into the sources of the component by cloning the repo? Did you try that? It's maybe a mistake of not exporting something on their side ?

src/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+      value={value}
+      onChange={onChange}
+    >
+      <StyledListboxButton arrow={arrow} {...props} />
+      <StyledListboxPopover>
+        <ListboxList>{children}</ListboxList>
+      </StyledListboxPopover>
+    </StyledListboxInput>
+  );
+};
```
georgiee: looks great👍

src/components/dropdown/dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+interface DropdownOptionProps {
+  /**
+   * Value which is returned when `onChange()` of the parent is called
+   * */
+  value: string;
+  /**
+   * The label which is shown in the expanded dropdown. If no `popoverLabel` is given, the same value as in `children` is displayed
+   * */
+  popoverLabel?: string;
```
georgiee: as said before this is not an ideal name. let's think about alternatives at least once.

src/components/file-dropper/file-list-item.tsx
```diff
@@ -84,13 +83,17 @@ export const FileListItem = ({
...
           >
-            <option disabled>{t('career.file-category')}</option>
+            <DropdownOption value={t('career.file-category')} disabled>
+              {t('career.file-category')}
+            </DropdownOption>
             {fileCategories.map((category) => (
-              <option key={category}>{category}</option>
+              <DropdownOption value={category} key={category}>
+                {category}
+              </DropdownOption>
```
georgiee, line 97: That looks straight forward. How did it feel? Is it a good replacement?

georgiee (changes requested):
Thanks for this. Awesome to have it finally 🙌
I left a few comments and things to change. Holding this back for now because of this. Looking forward to round 2.

src/components/dropdown/dropdown.tsx
hofmann-felix: yes, the header has 100. I was thinking of a global z-index file, maybe in another PR 🤔

src/components/dropdown/dropdown.tsx
hofmann-felix: The native select has no pointer. I removed it from the dropdown component

src/components/dropdown/dropdown.stories.mdx
hofmann-felix: I now understand how to use the ReachUI props and have imported them. Here the prop is simply called `label`. Is there a way in TS to rename a property name of an interface that you extend?

src/components/dropdown/dropdown.stories.mdx
hofmann-felix: Since I'm now using the interfaces from ReachUI, the args table isn't working anymore. I spent some time trying to fix it but it didn't work. Do you have a good suggestion on how to best display the args from an external component?

src/components/file-dropper/file-list-item.tsx
hofmann-felix: I think the changes definitely make sense since we were able to fix all the problems we had with native select. 
The only thing that bothers me is that on mobile you no longer have the select popout that was shown with the normal select.

src/components/dropdown/dropdown.tsx
georgiee: later through the theme is best. thanks

src/components/file-dropper/file-list-item.tsx
georgiee: maybe there is some issue/workaround for reach ui because this is a pretty standard behaviour.

georgiee (approved):
As agreed:
+ rename stories (form/s)
+ move (new-components)

and then ready to merge 🙌

Afterwards we should not forget to think about:

**How can we trigger mobile native ui/behavior ?**
I quickly checked the reach repo and also the most famous select repo: https://github.com/JedWatson/react-select both don't even mention "native + mobile". I simply can't find people talking about that fact. I don't understand that at all, because it's something to see missing on mobile 😳

PR #244, opened 2021-12-07
feat: create css touch detection util

feedm3:
I created a utility function to add a CSS media query to detect touch and non-touch devices. As the `breakpoints` file is also a CSS utility, I decided to put both files into the `style-utils` folder.

1. I don't feel 100% comfortable with that name, but I didn't want to spend more time on finding a better name / structure.
2. The detection of touch / no-touch isn't perfect either. I wanted to detect a touchpad to treat it like a touch device, but this didn't work. A touchpad is like a mouse in CSS world, as it has pointer and hover functionality. It's probably sufficient for 99% of the use cases, so better than nothing.

URL: https://satellytescommain21751-choretouchdetection.gtsb.io/

There is now also some outer right padding in the teaser grid:

<details>
<summary>Perfect fit on desktop</summary>
Screenshot 2021-12-07 at 09 18 42
</details>

<details>
<summary>Mobile initial overflow</summary>
Screenshot 2021-12-07 at 09 19 16
</details>

<details>
<summary>Gap to outer right when scrolled to the end</summary>
Screenshot 2021-12-07 at 09 20 24
</details>


src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -25,13 +39,26 @@ const StyledTeaserGrid = styled.div<{ amountOfChildren: number }>`
...
+  /**
+   * we give the last element 24px more space for the padding 
+   * -> distance to outer window
+   */
   grid-template-columns: ${(props) =>
-    `repeat(${props.amountOfChildren}, minmax(224px, 1fr))`};
+    `repeat(${props.amountOfChildren - 1}, minmax(224px, 1fr)) 248px`};
+
+  > :last-child {
+    padding-right: 24px;
```
feedm3, line 51: This took me ages to figure out 😄 I was always thinking of an `::after` element like it's necessary when you do it with Flexbox (https://stackoverflow.com/questions/38993170/last-margin-padding-collapsing-in-flexbox-grid-layout)

src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -1,9 +1,21 @@
...
+
+  ${isNoneTouch()} {
+    overflow-x: scroll;
+
+    /**
+     * we push to content box over it's boundaries to have some distance to
+     * to scroll bar
+     */
+    padding-bottom: 16px;
+    margin-bottom: -16px;
```
georgiee, line 15: I can see the difference. Is it worth though? Or does it look better for me?

Screen Shot 2021-12-08 at 18 05 49

Screen Shot 2021-12-08 at 18 05 46

src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -1,9 +1,21 @@
...
 import styled from 'styled-components';
-import { up } from '../../breakpoint/breakpoint';
+import { up } from '../../style-utils/breakpoint';
+import { isNoneTouch } from '../../style-utils/is-touch';
 
 const ScrollContainer = styled.div`
-  overflow-x: auto;
+  overflow: hidden;
+
+  ${isNoneTouch()} {
```
georgiee, line 9: I love this 🥳 I think we should move the teaser grid in the teaser component. It's always used together so why split them up ? This is a great feature addition. Thank you!

src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -1,9 +1,21 @@
...
+    overflow-x: scroll;
```
georgiee: why not auto? This always forces a scrollbar? I changed this before to auto because of this
Screen Shot 2021-12-08 at 18 08 56

Can you explain the problem with scroll between mobile & desktop again? I never had any special issues in my mind. Why change it across device types?

src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -25,13 +39,26 @@ const StyledTeaserGrid = styled.div<{ amountOfChildren: number }>`
...
 
   justify-items: stretch;
+
+  /**
+   * we give the last element 24px more space for the padding 
+   * -> distance to outer window
+   */
   grid-template-columns: ${(props) =>
-    `repeat(${props.amountOfChildren}, minmax(224px, 1fr))`};
+    `repeat(${props.amountOfChildren - 1}, minmax(224px, 1fr)) 248px`};
```
georgiee, line 48: isn't it possible to use `auto-fill` and friends here? Less explicit but we can get a similar behavior I bet.
https://css-tricks.com/auto-sizing-columns-css-grid-auto-fill-vs-auto-fit/

src/components/teasers/grid/teaser-grid.tsx
```diff
@@ -25,13 +39,26 @@ const StyledTeaserGrid = styled.div<{ amountOfChildren: number }>`
   gap: 24px;
 
   justify-items: stretch;
+
+  /**
+   * we give the last element 24px more space for the padding 
+   * -> distance to outer window
+   */
```
georgiee: can you explain with an image? I can't understand the padding. Because of the horizontal scrolling ?

georgiee (changes requested):
Thank you for that nice tool. Might be handy in the future 🙌
I also think it belongs in the same location as breakpoint. `style-utils` is okay for now but the `components/` folder feels awkward. We will do some refactoring tomorrow anyway, maybe it makes sense then. Why not outside `components` will be a question to challenge this tomorrow.

I have some questions before approving 🙏

src/components/teasers/grid/teaser-grid.tsx
feedm3: If you have a "permanent" scroll bar, the browser adds some extra padding by default. But if you are on Mac with a touchpad, it doesn't. The appearing scroll bar will flow over the arrow.
I wanted to add a screenshot, but the scroll bar disappears too fast.

src/components/teasers/grid/teaser-grid.tsx
feedm3: Thanks :)

Good idea to bring them closer together, will do.

src/components/teasers/grid/teaser-grid.tsx
feedm3: The big issue is the different scroll bar handling between touchpad (on Mac) and a real mouse. Because the browser adds some padding to permanent scroll bars. I was probably just lost in the issue, but you are right: `overflow-x: auto` works when it's added to the mobile layout. But the extra `padding-bottom` on mobile is needed for touchpads, as the scrollbar will otherwise flow over the arrow.

src/components/teasers/grid/teaser-grid.tsx
feedm3: On mobile, the `teaser-grid` needs to overflow the pages padding on the right side. 

Screenshot 2021-12-09 at 07 56 15

To compensate for this overflow, the last element needs to be `24px` larger with a padding of the same size. I tried other solutions but couldn't make it work. Also spent too much time on this already.

I try to make the comment above the code more clear.

src/components/teasers/grid/teaser-grid.tsx
feedm3: Couldn't make it work with `auto-fill`

src/components/teasers/grid/teaser-grid.tsx
georgiee: I always forget about the trackpad and I have problems reproducing it. I have an internal trackpad + safari + mouse (also tried to disable bluetooth). Any way I trust you here, let's move forward!

src/components/teasers/grid/teaser-grid.tsx
georgiee: thanks!

src/components/teasers/grid/teaser-grid.tsx
georgiee: another PR another try I guess. Let's try to merge this baby first.

src/components/teasers/grid/teaser-grid.tsx
georgiee: I'm find with that solution for now

src/components/teasers/grid/teaser-grid.tsx
georgiee: Okay as said, I have problems reproducing it. Maybe you can demonstrate it to me at some time. Good for now!

georgiee (approved):
Let's merge. Thank you for this change!

PR #246, opened 2021-12-07
feat: update career page

georgiee:
Let's try to run only a shallow review on this. I will work on the career page with other PRs anyway and I'm trying out a few things in terms of component organization. I'm happy to discuss and talk about things I did, this is not something I want to force. I just want to get some of the content out of my mind to focus on the rest.

---

I have walked over the career page, wrote missing texts and then created section after section. I would like to push this change to get a clean table for the next iteration: adding images that spread across the full width and into the header.

I had big problems finding a good way for the sizing and such. I also didn't want to use the `HeaderBlock` which is kind of obsolete almost (after the split of SectionHeader/Blog Header). What I did is use mostly existing components but for a few headlines or texts I created a local component. 

In the same vein I struggled with the spacing. I use the `ContentBlockContainer` but it feels not ideal but good enough. At some points I inserted a spacer to try out how it feels or I wrapped a component. 

**Important:**
I would like you to focus on technical problems or content-wise (as I wrote a few textst). Please let's ignore the spacing and the actual composition of components. My goal is to have more material with yet another page and then we can hopefully generalize a few concepts.

## Changes/Remarks
+ I had to fix the parsing of our jobs to get the jobbs
+ I decided to create a career page component and isolate it from the actual thin gatsby wrapper in `pages/`. I like how this feels because I can clearly see that the page only depends on `positions`.
+ I have not translated the texts. This is something done by some translation agency later. Rather then translating I would like to disable the English version of we keep mixed content. We should not translate ourselves.
+ I have placed the "partials" of the career page in `page-building/` as I had not better name. Ideally it would be in `components/pages` while our ui components would live in `components/ui` and such. We can do that later.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -5,7 +5,7 @@ const {
 const { createBlogPosts } = require('./gatsby/create-pages/create-blog-posts');
 const {
   createCareerDetails,
-} = require('./gatsby/create-pages/create-carrer-details');
+} = require('./gatsby/create-pages/create-career-details');
 const { createFilePath } = require('gatsby-source-filesystem');
 
 exports.onCreateNode = (gatsbyCreateNodeArgs) => {
--- a/gatsby/create-pages/create-career-details.js
+++ b/gatsby/create-pages/create-career-details.js
@@ -44,13 +44,10 @@ const createCareerDetails = async ({ actions }) => {
       const jobsXml = await jobsXmlResponse.data;
 
       const xmlParser = new XMLParser({
-        tagValueProcessor: (tag, value) => {
-          return decode(value);
-        },
+        tagValueProcessor: (_, value) => decode(value),
       });
       const jobsParse = xmlParser.parse(jobsXml);
       const positions = jobsParse['workzag-jobs'].position;
-
       return { positions, languageKey };
     },
   );
--- a/src/@types/personio.d.ts
+++ b/src/@types/personio.d.ts
@@ -1,0 +1,22 @@
+export interface PersonioJobNameValue {
+  name: string;
+  value: string;
+}
+
+export interface PersonioJobPosition {
+  id: string;
+  office: string;
+  department: string;
+  name: string;
+  jobDescriptions: {
+    jobDescription: PersonioJobNameValue[];
+  };
+  employmentType: string;
+  seniority: string;
+  schedule: string;
+  // comma separated
+  keywords: string;
+
+  satellytesPath: string;
+  satellytesShortDescription: string;
+}
--- a/src/components/accordion/accordion.tsx
+++ b/src/components/accordion/accordion.tsx
@@ -52,6 +52,7 @@ const PanelContainer = styled.div`
 
 const PanelIllustration = styled(Illustration)`
   padding-right: 1em;
+  box-sizing: content-box;
   float: left;
 
   ${up('sm')} {
--- a/src/components/teasers/teaser.tsx
+++ b/src/components/teasers/teaser.tsx
@@ -1,11 +1,6 @@
 import React from 'react';
 import styled, { css } from 'styled-components';
-import {
-  TeaserTitle,
-  Text,
-  Timestamp,
-  Topline,
-} from '../typography/typography';
+import { TeaserTitle, Timestamp, Topline } from '../typography/typography';
 import { Arrow } from '../icons/arrow';
 import { theme } from '../layout/theme';
 import { Link } from '../links/links';
--- a/src/components/teasers/teaser.tsx
+++ b/src/components/teasers/teaser.tsx
@@ -42,7 +42,9 @@ const ToplineContainer = styled.div`
   justify-content: space-between;
 `;
 
-const TeaserText = styled(Text)`
+const TeaserText = styled.div`
+  font-size: 16px;
+  line-height: 150%;
   margin: 16px 0 22px;
 `;
 
@@ -69,7 +71,7 @@ export interface TeaserProps {
   topline?: string;
   dateFormatted?: string;
   cover?: JSX.Element;
-  linkTo: string;
+  linkTo?: string;
 }
 
 /**
@@ -77,6 +79,15 @@ export interface TeaserProps {
  * This requires a headline, the path to the page (linkTo) and a short text, which is entered as a child.
  * In addition, an illustration or an image, a formatted date and a topline can be displayed.
  */
+
+const ConditionalLink = ({ to, children }) => {
+  if (to) {
+    return <Link to={to}>{children}</Link>;
+  }
+
+  return children;
+};
+
 export const Teaser: React.FC<TeaserProps> = ({
   topline,
   title,
@@ -89,7 +100,7 @@ export const Teaser: React.FC<TeaserProps> = ({
 
   return (
     <TeaserContainer>
-      <Link to={linkTo}>
+      <ConditionalLink to={linkTo}>
         {cover && <CoverContainer>{cover}</CoverContainer>}
         {hasToplineContainer && (
           <ToplineContainer>
@@ -101,8 +112,8 @@ export const Teaser: React.FC<TeaserProps> = ({
           {title}
         </StyledTeaserTitle>
         <TeaserText>{children}</TeaserText>
-        <Arrow />
-      </Link>
+        {linkTo && <Arrow />}
+      </ConditionalLink>
     </TeaserContainer>
   );
 };
--- a/src/page-building/career/application-process.tsx
+++ b/src/page-building/career/application-process.tsx
@@ -3,7 +3,6 @@ import {
   Accordion,
   AccordionSection,
 } from '../../components/accordion/accordion';
-import { TextStyles } from '../../components/typography/typography-v2';
 import styled from 'styled-components';
 import { SectionHeader } from '../../new-components/section-header/section-header';
 
--- a/src/page-building/career/application-process.tsx
+++ b/src/page-building/career/application-process.tsx
@@ -14,8 +14,8 @@ export const ApplicationProcess = () => {
   return (
     <>
       <SectionHeader headline="Bewerbungsprozess">
-        Sich auf einen Job bewerben, das kann oft frustrierend sein. Unser Team
-        hat sich deshalb etwas überlegt, was Sinn macht. Schau dir hier unsere
+        Sich auf einen Job zu bewerben kann oft frustrierend sein. Unser Team
+        hat sich deshalb etwas überlegt was Sinn macht. Schau dir hier unsere
         vier Schritte deiner Bewerbung an.
       </SectionHeader>
 
@@ -39,13 +39,13 @@ export const ApplicationProcess = () => {
         >
           Wenn wir deine Bewerbung überzeugend finden, laden wir dich zu einem
           ersten Gespräch ein. Hier werden ein oder zwei deiner zukünftigen
-          Kollegen mit dir sprechen. Darüber wie du zum Programmieren gekommen
-          bist und was dich antreibt. Danach werden werden wir dich besser
+          Kollegen mit dir sprechen und wir reden darüber wie du zum Programmieren gekommen
+          bist und was dich antreibt. Danach werden wir dich besser
           kennen und du kennst die ersten Gesichter hinter Satellytes.
         </AccordionSection>
 
         <AccordionSection title="3. Technisches Interview">
-          Wir machen keine keine Whiteboard Interviews, noch schicken wir die
+          Wir machen weder Whiteboard Interviews, noch schicken wir die
           irgendwelche Programmieraufgaben. Stattdessen wirst du vorab eine Code
           Review durchführen, die wir dann zusammen im Interview durchsprechen.
           Hier werden sich ganz automatische spannende Gespräch ergeben. Wir
@@ -56,9 +56,9 @@ export const ApplicationProcess = () => {
         </AccordionSection>
 
         <AccordionSection title="4. Entscheidung">
-          Du kennst Du hast gezeigt was du kannst und wie du arbeitest. Wir
+          Du hast gezeigt was du kannst und wie du arbeitest. Wir
           haben gezeigt was uns antreibt und wie wir arbeiten. Wenn alles passt
-          lassen wir es dir wissen. Passt etwas nicht, dann lassen wir das dich
+          lassen wir es dich wissen. Passt etwas nicht, dann lassen wir das dich
           ebenfalls wissen.
         </AccordionSection>
       </Accordion>
--- a/src/page-building/career/application-process.tsx
+++ b/src/page-building/career/application-process.tsx
@@ -7,12 +7,6 @@ import { TextStyles } from '../../components/typography/typography-v2';
 import styled from 'styled-components';
 import { SectionHeader } from '../../new-components/section-header/section-header';
 
-const Headline = styled.h2`
-  margin: 0;
-  ${TextStyles.headlineM}
-  margin-bottom: 16px;
-`;
-
 const Spacer = styled.div`
   margin-bottom: 48px;
 `;
--- a/src/page-building/career/application-process.tsx
+++ b/src/page-building/career/application-process.tsx
@@ -32,7 +32,6 @@ export const ApplicationProcess = () => {
           deiner Leidenschaft fürs Coden. Zeig uns deinen GitHub oder GitLab
           Account. Hast du einen Blog? Dann verweise gerne darauf.
         </AccordionSection>
-
         <AccordionSection
           title="2. Kennenlernen"
           illustration={'astronaut_015'}
@@ -43,7 +42,6 @@ export const ApplicationProcess = () => {
           bist und was dich antreibt. Danach werden wir dich besser
           kennen und du kennst die ersten Gesichter hinter Satellytes.
         </AccordionSection>
-
         <AccordionSection title="3. Technisches Interview">
           Wir machen weder Whiteboard Interviews, noch schicken wir die
           irgendwelche Programmieraufgaben. Stattdessen wirst du vorab eine Code
@@ -54,7 +52,6 @@ export const ApplicationProcess = () => {
           wir auch Zeit etwas über unsere Kultur bei Satellytes zu sprechen, um
           sicherzustellen dass du dich bei uns wohl fühlst.
         </AccordionSection>
-
         <AccordionSection title="4. Entscheidung">
           Du hast gezeigt was du kannst und wie du arbeitest. Wir
           haben gezeigt was uns antreibt und wie wir arbeiten. Wenn alles passt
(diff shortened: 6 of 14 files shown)


gatsby-node.js
```diff
@@ -5,7 +5,7 @@ const {
 const { createBlogPosts } = require('./gatsby/create-pages/create-blog-posts');
 const {
   createCareerDetails,
-} = require('./gatsby/create-pages/create-carrer-details');
+} = require('./gatsby/create-pages/create-career-details');
```
georgiee, line 8: found that typo 🤓

gatsby/create-pages/create-career-details.js
```diff
@@ -44,7 +44,7 @@ const createCareerDetails = async ({ actions }) => {
       const jobsXml = await jobsXmlResponse.data;
 
       const xmlParser = new XMLParser({
-        tagValueProcessor: (a) => decode(a), // &#039; -> '
+        tagValueProcessor: (_, value) => decode(value),
```
georgiee, line 47: it's tag & value. I don't know when this changed but we need the value to build our json output.

src/@types/personio.d.ts
```diff
@@ -0,0 +1,22 @@
+export interface PersonioJobNameValue {
```
georgiee, line 1: I decided to put it here in types to share it.

src/components/accordion/accordion.tsx
```diff
@@ -52,6 +52,7 @@ const PanelContainer = styled.div`
 
 const PanelIllustration = styled(Illustration)`
   padding-right: 1em;
+  box-sizing: content-box;
```
georgiee, line 55: The icon will shrinking smaller then the given size because flex is somehow involving the present padding in the shrink/grow calculations.

src/components/teasers/teaser.tsx
```diff
@@ -89,7 +95,7 @@ export const Teaser: React.FC<TeaserProps> = ({
 
   return (
     <TeaserContainer>
-      <Link to={linkTo}>
+      <ConditionalLink to={linkTo}>
```
georgiee, line 98: The design doesn't show it, but perks and such on the career page don't have any destination. So I made the link optional and remove the link component than, hence the conditional component. I decided not to update the hover as it's much more to rework. This component needs some refactoring anyway (starting with providing a story mdx)

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const ApplicationProcess = () => {
+  return (
+    <>
+      <SectionHeader headline="Bewerbungsprozess">
+        Sich auf einen Job bewerben, das kann oft frustrierend sein. Unser Team
+        hat sich deshalb etwas überlegt, was Sinn macht. Schau dir hier unsere
+        vier Schritte deiner Bewerbung an.
+      </SectionHeader>
+
+      <Spacer />
```
georgiee, line 22: I just wanted to try it out instead of wrapping the header or accordion. We are missing something about the spacing and I would like to come back once we know a proper ruling of spaces. This is a local fix inside the component and I think it's okay for now, especially because I will touch the page with some other PRs anyway again.

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import {
+  Accordion,
+  AccordionSection,
+} from '../../components/accordion/accordion';
+import styled from 'styled-components';
+import { SectionHeader } from '../../new-components/section-header/section-header';
+
+const Spacer = styled.div`
+  margin-bottom: 48px;
```
georgiee, line 10: I don't care for mobile right now until we have some spacing agreement

src/page-building/career/career-page.tsx
```diff
@@ -0,0 +1,96 @@
...
+          </Paragraph>
+          <Paragraph>
+            Unser Büro befindet sich in einem wunderschönen Altbau im Herzen
+            Münchens, in der Sendlinger Straße, unweit der gleichnamigen
+            U-Bahn-Station. In Laufweite gibt es nicht nur dutzende Läden,
+            Cafés, Restaurants und den bekannten Viktualienmarkt, sondern auch
+            diverse kleinere und größere Parks. Sogar die Isar ist in der
+            Mittagspause fußläufig oder mit der U-Bahn in nur 3 Minuten zu
+            erreichen.
+          </Paragraph>
```
georgiee, line 71: That's something we don't have covered at all. Paragraphs. We just pass in raw text usually which is wrong. So I made this for some experience on creating atoms like paragraphs. Up to change but a first taste of working with atoms 🙏

src/page-building/career/career-page.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <Culture />
+      </ContentBlockContainer>
+
+      <ContentBlockContainer>
+        <Culture />
+      </ContentBlockContainer>
+
+      <ContentBlockContainer>
+        <Perks />
+      </ContentBlockContainer>
```
georgiee, line 93: I like the structure, each section of the career page is clearly visible. It's not a random mix of components anymore and if I want to update a place I can navigate straight to the right place 👍 

I don't like the noise coming from `ContentBlockContainer` and the extra hierarchy. This is just bringing in some spacing to the bottom. We can do better, why not so? (well I don't like that much either,  but it's a little bit better.

```
<ApplicationProcess />
<SectionSpace/>

<Openings jobs={positions} />
<SectionSpace/>

<Culture />
<SectionSpace/>

<Perks />
<SectionSpace/>
```

I don't want to merge the spacing into the sections itself. It mixes too much of the spacing concerns into an otherwise innocent component that you can check out easily in storybook right now (without any noisy spacings).

What we could do is to have some `SectionContainer` that is used by each of them (Perk, Culture etc) and then we have a rule like this:

```
& + & {
 margin-bottom..
```

That way they only add spacing between them if two or more are in the DOM. Maybe that's something?

src/page-building/career/career-page.tsx
```diff
@@ -0,0 +1,96 @@
...
+    illustration: 'astronaut_012',
+    contact: {
+      headline: t('career.leadbox.subtitle'),
+      title: t('career.leadbox.text'),
+      email: t('career.leadbox.mail'),
+    },
+  };
+
+  return (
+    <LayoutV2
```
georgiee, line 41: I tried out to mix in the Layout but I'm torn where the best place is. Gatsy Page or this isolated content page?

src/page-building/career/career.stories.mdx
```diff
@@ -0,0 +1,51 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { ApplicationProcess } from './application-process'; 
+import { Perks } from './perks'; 
+import { Openings } from './openings';
+import { Culture } from './culture'; import { CareerPage } from './career-page';
+
+<Meta title="Pages/Career" component={{ Image }} />
+
+# Career
```
georgiee, line 9: I like how I can quickly check the sections available on the career page. I used this while developing.

src/page-building/career/culture.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  justify-items: stretch;
+  grid-template-columns: repeat(auto-fit, 250px);
+  ${up('md')} {
+    gap: 70px;
+    grid-template-columns: 1fr 1fr 1fr;
+  }
+`;
+
+const ASPECTS: CultureAspect[] = [
```
georgiee, line 27: hard coded is fine because this component is local and not meant to be reused anywhere. That's why I removed the  props.

src/page-building/career/culture.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { IllustrationType } from '../../components/illustration/illustration-set';
+import { SectionHeader } from '../../new-components/section-header/section-header';
+
+interface CultureAspect {
+  illustration: IllustrationType;
+  title: string;
+  description: string;
+}
+
+const TeaserGrid = styled.div`
```
georgiee, line 15: I copied this from index.tsx and I think we should provide such a grid from within the teaser component package to share them properly 👍

src/page-building/career/openings.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+`;
+
+const SectionHeadline = styled.h2`
+  ${TextStyles.headlineXL}
+  margin: 0;
+  margin-bottom: 80px;
+`;
+
+type PersonioJobPositionPartial = Pick<
```
georgiee, line 26: I need a partial type here otherwise I can't give concise example data in the story. I'm not using more anyway so this is much more precise anyway.

src/page-building/career/openings.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <div>
+      <SectionHeadline>Unsere offenen Stellen</SectionHeadline>
+
+      <TeaserGrid>
+        {props.jobs.map((item) => (
+          <Teaser title={item.name} linkTo={item.satellytesPath} key={item.id}>
+            <div
+              dangerouslySetInnerHTML={{
+                __html: item.satellytesShortDescription,
```
georgiee, line 45: oh personio, stupid html coming in.

src/page-building/career/openings.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export const Openings = (props: OpeningsProps) => {
+  return (
+    <div>
+      <SectionHeadline>Unsere offenen Stellen</SectionHeadline>
+
+      <TeaserGrid>
+        {props.jobs.map((item) => (
+          <Teaser title={item.name} linkTo={item.satellytesPath} key={item.id}>
```
georgiee, line 42: `satellytesPath` comes from the xml parsing and I hope with a custom source plugin that we get more speaking names 👍

src/page-building/career/perks.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Illustration } from '../../components/illustration/illustration';
+import { up } from '../../components/breakpoint/breakpoint';
+import { IllustrationType } from '../../components/illustration/illustration-set';
+import { SectionHeader } from '../../new-components/section-header/section-header';
+
+interface Perk {
+  illustration: IllustrationType;
+}
+
+const TeaserGrid = styled.div`
```
georgiee, line 13: I didn't want to articially share the grid. It's local here and once we refactor and decide for some central location of this grid I'm happy to remove this duplicated code (as it exists in other sections too)

src/templates/career.tsx
```diff
@@ -93,71 +32,28 @@ interface CareerPageProps {
...
-          </GridItem>
-        </Grid>
-      </Layout>
+      <SEO
+        imageUrl={socialCard}
+        title={t('career.seo.title')}
+        description={t('career.seo.description')}
+        location={location}
+      />
+      <CareerPage positions={pageContext.positions} />
```
georgiee, line 51: I love how I can see what this page depends on. Before it was a mess.

georgiee (commented):
here some helping comments for your review 👍

src/page-building/career/career-page.tsx
fischeversenker: I'd still like to see a `<Stack />` approach in real life. `Stack` would be a component used purely for layouting other components.

I have something like this in mind:
Usage:
```html
<Stack [space="small|medium|large"]>
  <Openings jobs={positions} />
  <Culture />
  <Perks />
</Stack>
```
Styles inside `<Stack>`:
```css
:host > *:not(:last-child) {
  margin-bottom: var(--...);
}
```

georgiee: yeah that's nice too 🤘 But for this we need strict rules and the knowledge that we have instead distinct sections everywhere like I created here.

src/components/teasers/teaser.tsx
```diff
@@ -42,7 +37,9 @@ const ToplineContainer = styled.div`
   justify-content: space-between;
 `;
 
-const TeaserText = styled(Text)`
+const TeaserText = styled.div`
+  font-size: 16px;
+  line-height: 150%;
```
fischeversenker, line 42: Does this match any of the existing typography configurations? Let's try to re-use them whenever possible :pray:

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+`;
+
+export const ApplicationProcess = () => {
+  return (
+    <>
+      <SectionHeader headline="Bewerbungsprozess">
+        Sich auf einen Job bewerben, das kann oft frustrierend sein. Unser Team
+        hat sich deshalb etwas überlegt, was Sinn macht. Schau dir hier unsere
+        vier Schritte deiner Bewerbung an.
```
fischeversenker:
```suggestion
        Sich auf einen Job zu bewerben kann oft frustrierend sein. Unser Team
        hat sich deshalb etwas überlegt was Sinn macht. Schau dir hier unsere
        vier Schritte deiner Bewerbung an.
```

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+        <AccordionSection
+          title="2. Kennenlernen"
+          illustration={'astronaut_015'}
+        >
+          Wenn wir deine Bewerbung überzeugend finden, laden wir dich zu einem
+          ersten Gespräch ein. Hier werden ein oder zwei deiner zukünftigen
+          Kollegen mit dir sprechen. Darüber wie du zum Programmieren gekommen
+          bist und was dich antreibt. Danach werden werden wir dich besser
+          kennen und du kennst die ersten Gesichter hinter Satellytes.
```
fischeversenker:
```suggestion
          Wenn wir deine Bewerbung überzeugend finden, laden wir dich zu einem
          ersten Gespräch ein. Hier werden ein oder zwei deiner zukünftigen
          Kollegen mit dir sprechen und wir reden darüber wie du zum Programmieren gekommen
          bist und was dich antreibt. Danach werden wir dich besser
          kennen und du kennst die ersten Gesichter hinter Satellytes.
```

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+        <AccordionSection title="3. Technisches Interview">
+          Wir machen keine keine Whiteboard Interviews, noch schicken wir die
+          irgendwelche Programmieraufgaben. Stattdessen wirst du vorab eine Code
+          Review durchführen, die wir dann zusammen im Interview durchsprechen.
+          Hier werden sich ganz automatische spannende Gespräch ergeben. Wir
+          werden Probleme im Code besprechen und das ein oder andere Fachwissen
+          über relevante Technologien ansprechen. Neben der ganzen Technik haben
+          wir auch Zeit etwas über unsere Kultur bei Satellytes zu sprechen, um
+          sicherzustellen dass du dich bei uns wohl fühlst.
```
fischeversenker:
```suggestion
          Wir machen weder Whiteboard Interviews, noch schicken wir die
          irgendwelche Programmieraufgaben. Stattdessen wirst du vorab eine Code
          Review durchführen, die wir dann zusammen im Interview durchsprechen.
          Hier werden sich ganz automatische spannende Gespräch ergeben. Wir
          werden Probleme im Code besprechen und das ein oder andere Fachwissen
          über relevante Technologien ansprechen. Neben der ganzen Technik haben
          wir auch Zeit etwas über unsere Kultur bei Satellytes zu sprechen, um
          sicherzustellen dass du dich bei uns wohl fühlst.
```

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </AccordionSection>
+
+        <AccordionSection title="4. Entscheidung">
+          Du kennst Du hast gezeigt was du kannst und wie du arbeitest. Wir
+          haben gezeigt was uns antreibt und wie wir arbeiten. Wenn alles passt
+          lassen wir es dir wissen. Passt etwas nicht, dann lassen wir das dich
+          ebenfalls wissen.
```
fischeversenker:
```suggestion
          Du hast gezeigt was du kannst und wie du arbeitest. Wir
          haben gezeigt was uns antreibt und wie wir arbeiten. Wenn alles passt
          lassen wir es dich wissen. Passt etwas nicht, dann lassen wir das dich
          ebenfalls wissen.
```

src/page-building/career/application-process.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </AccordionSection>
+      </Accordion>
```
fischeversenker, line 65: Overall I'm not a fan of these line breaks everywhere in the code in the text blocks... while editing this I'd be confused why my line breaks are not respected on the final page.

src/page-building/career/career-page.tsx
fischeversenker: I'm fine with sticking to the `ContentBlockContainer` that you introduced for now. Nevertheless, we certainly need a more accessible solution for the future.

fischeversenker (approved):
I did a shallow review and left a few comments. Thanks for your initial set of comments, they helped.

The changes are ok. I don't like the name `page-building` for the new folder that you introduced... maybe there is something more descriptive? `page-sections`, `page-partials`, `page-blocks`?

I'll approve to not block this PR (I'm on vacation for the next few days).

src/page-building/career/career-page.tsx
feedm3: Just read this article. Seems like I missed the point where spacer components were cool again 😄 

https://www.joshwcomeau.com/react/modern-spacer-gif/

The article has some good arguments for `<Spacer />`. I guess we need to see how exactly we will use the component. For example, if we iterate over each content block, putting each block into a `<ContentBlockContainer />` feels better then returning `<><Culture /><Spacer/></>`. 

A `Stack` component is something I've never seen in the wild. Also sounds interesting as it is less verbose than both other solutions.  Without digging too deep into it, I wonder if it really works very reliable, as it relies on a certain structure of the children.

hofmann-felix (approved):
I did a shallow review too and just found some typos which were already mentioned in @fischeversenker's review. 

In general I think that the texts are very well done 👍

src/components/teasers/teaser.tsx
georgiee: We need to refactor the teaser story, I would prefer to handle it there. I just copied over the styles of the Text component which is an outdated component.

src/page-building/career/application-process.tsx
georgiee: nice thanks!

src/page-building/career/application-process.tsx
georgiee: oops

src/page-building/career/application-process.tsx
georgiee: well this is common knowledge for markdowns. It was just more comfortable to write the text
in a small block instead of a single line. We will move that into a translation file or such anyway
and that's when you get your one liner I guess 🤓

georgiee:
Thank you folks for the quick review and the text feedback. Very much appreciated.

PR #248, opened 2021-12-07
fix: add responsive headline to SectionHeader

hofmann-felix:
Storybook: https://deploy-preview-248--satellytes-website-storybook.netlify.app/?path=/story/components-sectionheader--regular

While creating the career page I noticed that the SectionHeader heading also needs a responsive font size. 

Bildschirmfoto 2021-12-07 um 14 49 01
Bildschirmfoto 2021-12-07 um 14 49 06


georgiee (approved):
I can confirm this from checking the design. The text is properly covered but the headline was missing. Thank you for spotting this and handling it in a separate PR 👍

PR #249, opened 2021-12-07
feat: update imprint and data-policy page

hofmann-felix:
Imprint Preview: https://satellytescommain21751-featupdateimprintdatapolicy.gtsb.io/imprint/
Data Privacy Preview: https://satellytescommain21751-featupdateimprintdatapolicy.gtsb.io/data-privacy/

## What's included?
- updated imprint and data-policy page design
- both pages are using the light theme
- the content is taken from a markdown file -> The two pages are using the old markdown renderer and components

## Preview
Data Policy:
Bildschirmfoto 2021-12-07 um 16 38 26

Imprint:
Bildschirmfoto 2021-12-07 um 16 39 55


src/components/layout/content-block-container.tsx
```diff
@@ -8,3 +9,11 @@ export const ContentBlockContainer = styled.div`
     margin-top: 180px;
   }
 `;
+
+export const ContentBlockContainerWithoutHero = styled.div`
```
georgiee, line 13: Funny name 😁 This is ok to share from here for now.
Once we have all pages I would suggest to work on the overall layout again to see what we are missing. 

Because a simple page like imprint with no header. That should not require any additional container. The Layotu should handle this. 

Your change is good, just some loud thought👍

georgiee (approved):
Looks splendid 🌟 Thank you for your effort.

PR #250, opened 2021-12-08
feat: add personio source plugin

georgiee:
This PR brings a new source plugin `sy-personio-jobs-source-plugin` and ensures that it's used properly for the career content.

# Includes
+ Source Plugin
+ `CareerPage` moves to `pages/` and uses a proper graphql query to fetch the data now. No manual creation with context anymore (that's why it was in `templates/` before) 👍
+ Career Details page uses a proper grapqhl query instead of a context.
+ Moved the "social card" generation into `onCreateNode` (via `createPreviewCards`) where we process blog previews through the node creation callback mechanism already. I think this is more elegant than coding it into the source plugin hence this split.
+ The career paths `/career/title-of-job`, `/de/career/title-of-job` and such are now created like the slug for markdown files also in `onCreateNode` because it's not the responsbility of the source plugin at all.

# Conclusion
I really love the way it works now. Much more cleaner and I can clearly understand what's happening. Especially the `create-career-pages.js` was a dirty mess with all the `jobDescriptions.jobDescription` and similar the career details. We removed the direct API involvement in nodes and we are back to "normal". That should be a learning for the future, that we should always embrace the internal redux store before using the API + context. I'm still thankful we got such a fast solution back then but it's now obvious that this was some dirty workaround that had an effect in many places.


.eslintrc.js
```diff
@@ -37,5 +37,12 @@ module.exports = {
...
     '@typescript-eslint/explicit-module-boundary-types': 'off',
   },
-  ignorePatterns: ['gatsby-*.js', 'jest', '*.md', 'tooling/**', 'gatsby/**'],
+  ignorePatterns: [
+    'gatsby-*.js',
+    'jest',
+    '*.md',
+    'tooling/**',
+    'gatsby/**',
+    'plugins/**',
```
georgiee, line 46: I got some errors with import statements for our js files. Can't we enable modern node to use imports? It's not optimal to remove all those files from linting.

gatsby-config.js
```diff
@@ -250,10 +251,6 @@ module.exports = {
             matchPath: '/:lang?/career/:id/',
             getLanguageFromPath: true,
           },
-          {
-            matchPath: '/:lang?/career/',
-            getLanguageFromPath: true,
-          },
```
georgiee, line 256: This prevented i18n from creating the career page duplicates internally with the correct language. It took a few moments to realize is hence this comment.

gatsby-config.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
     siteUrl: BASE_URL,
   },
   plugins: [
+    `sy-personio-jobs-source-plugin`,
```
georgiee, line 36: for an awesome win 🤘

gatsby-node.js
```diff
@@ -1,31 +1,43 @@
...
     createNodeField({
-      name: `slug`,
+      name: `path`,
       node,
       value,
     });
   }
+
+  if (node.internal.type === `SyPersonioJob`) {
+    createCareerPagePathField({ node, createNodeField });
```
georgiee, line 33: Instead of adding a `path` to the personio jobs through source we let the source plugin not know about this issue and instead create the required path field here. Only different: It's provided in `{fields: {path}}` instead of `{path}` which caused some refactoring when I changed it.

gatsby-node.js
```diff
@@ -1,31 +1,43 @@
...
+  }
 };
 
 exports.createPages = async (createPagesArgs) => {
-  await createCareerDetails(createPagesArgs);
+  await createCareerPages(createPagesArgs);
```
georgiee, line 38: better name

gatsby/create-node/create-preview-cards.js
```diff
@@ -0,0 +1,68 @@
...
+  const publicUrl = `${siteMetadata.siteUrl}/${fileName}`;
+
+  generateCard({ title: node.name }, outputFile).then(() => {
+    createNodeField({
+      node,
+      name: 'socialCard',
+      value: publicUrl,
+    });
+  });
+};
```
georgiee, line 54: the logic to create a preview card for a given job position. This is again not part of the source plugin. On the other hand seeing this and the path handling which are both attached on `onCreateNode`. Would this be a seprate plugiin (in an ideal world) ? Or would we marry the source plugin with those two responsibilities and remove the `source` part and call it our career plugin ? I like it more like that for now as it's like we would use any other source plugin.

gatsby/create-pages/create-career-pages.js
```diff
@@ -0,0 +1,67 @@
...
+    memo[job.jobId].push(job);
+    return memo;
+  }, {});
+
+  return {
+    get: (position) =>
+      groupedPositions[position.jobId].find((job) => job.id !== position.id) ??
+      null,
+  };
+};
```
georgiee, line 23: This is basically memozing what jobs have the same job id and then we have an easy time to find the other part. By currying we don't need to recalculate the memo everytime. Super premature optimization but I guess my AoC fever got me hire. 

I love how it reads later where it's used 🙏

```
 const complementFinder = createComplementFinder(positions);

  positions.forEach((position) => {
    const complementPosition = complementFinder.get(position);
```

gatsby/create-pages/create-career-pages.js
```diff
@@ -0,0 +1,67 @@
...
+          jobId
+          name
+          slug
+          fields {
+            path
+          }
+        }
+      }
+    }
+  `);
```
georgiee, line 45: There we are, finally consuming our jobs as graphql 🥳

gatsby/create-pages/create-career-pages.js
```diff
@@ -0,0 +1,67 @@
...
+  const complementFinder = createComplementFinder(positions);
+
+  positions.forEach((position) => {
+    const complementPosition = complementFinder.get(position);
+
+    createPage({
+      path: position.fields.path,
+      component: CAREER_DETAILS_TEMPLATE_PATH,
+      context: {
+        id: position.id, // internal object id which is unique across all jobs
```
georgiee, line 57: we only need to pass the id of the node to the career page as it will then query for the details it needs with this id 🙌

package.json
```diff
@@ -89,6 +89,7 @@
     "rehype-react": "^7.0.3",
     "sass": "^1.44.0",
     "slugify": "^1.6.3",
+    "striptags": "^3.2.0",
```
georgiee, line 92: we had some nasty regex and I wanted to have a well maintained library for this (again and again surprisingly) complex topic of stripping html.

plugins/sy-personio-jobs-source-plugin/gatsby-node.js
```diff
@@ -0,0 +1,32 @@
...
+      parent: null,
+      children: [],
+      internal: {
+        type: JOB_TYPE,
+        content: JSON.stringify(position),
+        contentDigest: createContentDigest(position),
+      },
+    });
+  });
+};
```
georgiee, line 32: that's the only magic need to make our jobs available through graphl. Fetch all datasets and then push them into the underlying redux store with `createNode` and gatsby does the rest of inferring the graphql schema. This is awesome everytime I think about it.

plugins/sy-personio-jobs-source-plugin/index.js
```diff
@@ -0,0 +1 @@
+// noop
```
georgiee, line 1: that's common for local gatsby plugins, otherwise it's not recognized as such (or did that change 😬)

plugins/sy-personio-jobs-source-plugin/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "sy-personio-jobs-source-plugin",
```
georgiee, line 2: required for local plugins too

plugins/sy-personio-jobs-source-plugin/personio-jobs-api.js
```diff
@@ -0,0 +1,137 @@
...
+  ]);
+
+  const sections = remainingDescriptions.map((item) => ({
+    headline: item.name,
+    descriptionHtml: item.value.trim(),
+    description: striptags(item.value.trim()),
+  }));
+
+  return { sections, slug, short };
+}
```
georgiee, line 84: Yeah read the code comment above. This was super annoying throughout the career content a we never "normalized" those parts. I wanted to do this here.
1. Remove the special keys and carry around the raw value (slug and short description)
2. Rename the sections and provide text and html variants of the content

src/locales/de/translations.json
```diff
@@ -23,21 +23,7 @@
     "career": {
       "title": "Arbeite mit uns",
       "topline": "Karriere bei Satellytes",
-      "text": "Wir suchen Entwickler:innen aus Leidenschaft! Du hast noch nicht viel Berufserfahrung? Kein Problem. Denn alles, was du wissen musst, lernst Du bei uns. Du kannst schon alles? Dann findest Du bei Satellytes neue Herausforderungen und erfahrene Kollegen, mit denen Du weiter wachsen kannst. Schaue Dir unsere offenen Stellen an. Wir freuen uns auf Deine Bewerbung.",
```
georgiee, line 26: obsolete now as we are loading the real stuff

src/locales/en/translations.json
```diff
@@ -23,21 +23,7 @@
     "career": {
       "title": "Work with us",
       "topline": "Career at Satellytes",
-      "text": "Wir suchen Entwickler:innen aus Leidenschaft! Du hast noch nicht viel Berufserfahrung? Kein Problem. Denn alles, was du wissen musst, lernst Du bei uns. Du kannst schon alles? Dann findest Du bei Satellytes neue Herausforderungen und erfahrene Kollegen, mit denen Du weiter wachsen kannst. Schaue Dir unsere offenen Stellen an. Wir freuen uns auf Deine Bewerbung.",
```
georgiee, line 26: obsolete now as we are loading the real stuff

src/page-building/career/openings.tsx
```diff
@@ -23,28 +23,18 @@ const SectionHeadline = styled.h2`
...
       <TeaserGrid>
         {props.jobs.map((item) => (
-          <Teaser title={item.name} linkTo={item.satellytesPath} key={item.id}>
-            <div
-              dangerouslySetInnerHTML={{
-                __html: item.satellytesShortDescription,
-              }}
-            />
+          <Teaser title={item.name} linkTo={item.fields.path} key={item.id}>
+            {item.short}
```
georgiee, line 37: much better with the parsed short text 🥳

src/pages/index.tsx
```diff
@@ -139,35 +175,7 @@ const IndexPage = ({ location }: IndexPageProps) => {
...
-            </Teaser>
-            <Teaser
-              title={t('main.career.teasers.third.title')}
-              linkTo="/career"
-            >
-              {t('main.career.teasers.third.text')}
-            </Teaser>
-          </TeaserGrid>
-        </ContentBlockContainer>
+        <CareerSection positions={jobPositions} />
```
georgiee, line 178: I think it's always worth to divide and conquer even if you only create local components. Do you agree? Make it much more easy to read and navigate.

src/templates/career-details.tsx
```diff
@@ -70,40 +68,37 @@ export const PersonioHtml = styled.div`
...
   };
   location: Location;
+  data: {
+    syPersonioJob: SyPersonioJob;
+  };
 }
 const CareerPage: React.FC<CareerPageProps> = (props): JSX.Element => {
   const { pageContext } = props;
-  const { t } = useTranslation();
+  const position = props.data.syPersonioJob;
```
georgiee, line 84: I did only the things to keep this working. We still need to update the career details page

georgiee:
Last error was during install on the CI pipeline for standard package `sharp`. Weird. I forced a rebuild.
>Status 503 Egress is over the account limit.

gatsby/create-node/create-preview-cards.js
feedm3: it wouldn't be a "source" plugin, but a normal plugin. 

we could create a "gatsby-plugin-preview-image" plugin and share it on the gatsby plugins page

plugins/sy-personio-jobs-source-plugin/personio-jobs-api.js
```diff
@@ -0,0 +1,137 @@
...
+
+    return memo;
+  }, []);
+}
+
+/**
+ * Fetch all available jobs for the given language
+ * from our Personio API endpoint.
+ */
+async function fetch(language) {
```
feedm3, line 121: would call it `fetchJobPositions` instead of a plain fetch, as it could be misunderstood with the global fetch

feedm3 (approved):
We already talked about this PR, so I kept the review very short. Very good job, well done! This code reads like a book. It's very well structured, clean, simple to read, and fit's perfectly with gatsby. Let's merge it.

plugins/sy-personio-jobs-source-plugin/personio-jobs-api.js
georgiee: I had it like that but I though the package tells the story so no misunderstanding. Let's keep it as is and we can still refactor if continues to upset us 😁

georgiee:
Thank you for your kind words 🙏

Issue #254, opened 2021-12-09
Smaller Issues Bucket


georgiee:
Let's collect a few things to address (please add yours).

+ Teasers need a refactoring. The `cover` is not appropriate anymore. Make sure it uses the mdx story format and it's placed in new-components.
+ Teasers can have no link and should remove the hover effect then.
+ Refactoring: Make sure al gatsby pages are using the new format where the entire content (incl. layout) is wrapped in an additional component and placed in `page-building/` (folder name is going change)

georgiee:
This list is obsolete and outdated. Most things are done.

PR #261, opened 2021-12-13
refactoring(Teaser): use new components

hofmann-felix:
Resolves Teaser issues from https://github.com/satellytes/satellytes.com/issues/254
Storybook: https://deploy-preview-261--satellytes-website-storybook.netlify.app/?path=/docs/components-teaser--regular&globals=locale:en;locales.en:English;locales.de:Deutsch
Preview: https://satellytescommain21751-teaserissues.gtsb.io/

## What's included?
- new `Icon` component (for the arrow) is used
- new `Illustration` component is used: The illustration is now passed as an IllustraionType prop. Therefore the `cover` prop had to be split into `illustration` and `teaserImage`. This also makes sense, since the spacing in the component depends on whether it is an illustration or an image.
- The hover effect is now controlled with the `hoverActive` state -> This allows removing the CSS spaghetti architecture 
- The component can be used without a link now
- the component is moved to the new-components folder
- The whole teaser story is in a `.mdx` file now and the changes of this PR have been added to the story

## Preview
Bildschirmfoto 2021-12-13 um 12 32 20


src/new-components/icon/icon-files.tsx
```diff
@@ -6,3 +6,5 @@ export { ReactComponent as chevron_right } from '../../icons/chevron-right-24.sv
 export { ReactComponent as chevron_up } from '../../icons/chevron-up-24.svg';
 export { ReactComponent as clone } from '../../icons/clone-24.svg';
 export { ReactComponent as burger_nav } from '../../icons/burger-nav-24.svg';
+export { ReactComponent as arrow_right } from '../../icons/arrow-right.svg';
+export { ReactComponent as arrow_left } from '../../icons/arrow-left.svg';
```
fischeversenker: The other icon files are called `...-24.svg` but the new files that you added are not. Please try to stick to existing naming conventions.

src/new-components/teaser/teaser.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+const StyledIcon = styled(Icon)`
+  margin-top: 14px;
+`;
+
+export interface TeaserProps {
+  title: string;
+  topline?: string;
+  dateFormatted?: string | null;
+  teaserImage?: JSX.Element;
```
georgiee: let's name it `image` to be clean, you don't need to remind me that I'm working towards a teaser:

`<teaser teaserImage` vs  `<teaser image`

src/new-components/teaser/teaser.tsx
```diff
@@ -0,0 +1,166 @@
...
+        {teaserImage && (
+          <ImageContainer hover={hoverActive}>{teaserImage}</ImageContainer>
+        )}
+        {illustration && (
+          <StyledIllustration
+            show={illustration}
+            size={IllustrationSize.MEDIUM}
+            hover={hoverActive}
+          />
+        )}
```
georgiee: I think we should **not** allow both at the same time. Let's state that in the description and enforce an order:

something like this:
> You can't pass in an image and illustration. If you pass in an image and illustration the illustration will be taken.

georgiee (changes requested):
Hey that's a nice change, thank you 🙌
1. Could you change a few little things as noted in the review?
2. The `TeaserGrid` is a separate story. Could you please put the teaser grid into the teaser folder and add a story to the teaser stories? They both will never exist in isolation so please marry them 👰‍♀️🤵‍♂️

src/new-components/teaser/teaser.tsx
```diff
@@ -0,0 +1,166 @@
...
+  teaserImage?: JSX.Element;
+  illustration?: IllustrationType;
+  linkTo?: string;
+  className?: string;
+  children: React.ReactNode | React.ReactNode[];
+}
+
+/**
+ * Teasers are used to link to another page and give the user a brief overview of the content of that page.
+ * This requires a headline, the path to the page (linkTo) and a short text, which is entered as a child.
```
fischeversenker: The bit about the "requires [...] path to the page" is not true anymore, right? We can have link-less teasers as far as I understand.
That means this comment is outdated and it's also in the wrong place. It documents `ConditionalLink` and I'm surprised that it is being picked up by Storybook. Looks more like a bug than a feature.
I suggest to add comments like this in the `mdx` file instead of the source. Let's take care of this when we introduce a proper mdx file.

src/new-components/teaser/teaser.tsx
fischeversenker: Do we want some logic here to prevent someone from adding an image as well as an illustration? Maybe a simple `{!teaserImage && illustration && (...)}` for the second condition (or the other way around, like @georgiee suggested)?

fischeversenker (commented):
This looks nice (even though it didn't change anything visually 😂)!

I left a few comments and I'd like to suggest to include the refactoring from `stories.tsx` to `stories.mdx` in this PR. It's a very related changed and the size of the PR is still very manageable in my opinion.

src/new-components/teaser/teaser.tsx
georgiee: We are responsible developers. I think a convention as mentioned is enough, isn't it ?

src/new-components/teaser/teaser.tsx
georgiee: Good catch, I can second this👍

src/new-components/teaser/teaser.tsx
fischeversenker: We certainly are, you are right. However, if we write something like "If you pass in an image and illustration the illustration will be taken." we also need to back that with logic and as far as I can tell it doesn't exist yet, does it?

hofmann-felix: > You can't pass in an image and illustration. If you pass in an image and illustration the illustration will be taken.

I added @fischeversenker's logic to make the component fit the description

src/new-components/teaser/teaser.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+  title="Components/Teaser"
+  component={{ Teaser }} />
+
+# Teaser
+Teasers are used to link to another page and give the user a brief overview of the content of that page.
+This requires a `title` and a short text, which is entered as a child.
+In addition, an `illustration` or an `image`, a formatted date (`dateFormatted`), a link to another page (`linkTo`) and a `topline` can be displayed.
+You can't pass in an image **and** illustration. If you pass in an image and illustration the illustration will be taken.
+
+<ArgsTable of={Teaser} />
```
georgiee, line 15: That looks great and useful in the introduction 👍
Screen Shot 2021-12-14 at 14 48 44

src/new-components/teaser/teaser.tsx
```diff
@@ -0,0 +1,181 @@
...
+          hasTopline={hasToplineContainer}
+        >
+          {title}
+        </StyledTeaserTitle>
+        <TeaserText>{children}</TeaserText>
+        {linkTo && <StyledIcon show={'arrow_right'} />}
+      </ConditionalLink>
+    </TeaserContainer>
+  );
+};
```
georgiee, line 181: Seeing big components like that I always feel an impulse to refactor it.
1. image + illustration is a semantic block which you could handle in a local component
2.  topline with date is a separate part
3. the title, text & icon is another.

That way this part could look much more accessible for a developer. Please give it a try in a future PR 🙏
```
<Teaser>
   <Cover/>
  <Topline/>
  <Content/>
</Teaser>
```

src/page-building/career/culture.tsx
```diff
@@ -82,7 +81,7 @@ export const Culture = () => {
           <Teaser
             title={item.title}
             key={index}
-            cover={<Illustration show={item.illustration} />}
+            illustration={item.illustration}
```
georgiee, line 84: looks much cleaner now

src/page-building/landingpage/service.tsx
```diff
@@ -31,27 +27,21 @@ export const Service = () => {
         <Teaser
           title={t('services.platform.title')}
           linkTo="/services"
-          cover={
```
georgiee, line 34: you harvest the sweet fruits of refactoring here 🍇 good job !

georgiee (approved):
I like it, let's merge.
I have one refactoring suggestion which you could put on your agenda (regarding the teaser grid, to merge it with the teaser folder & stories)


Edit: See my past review, I think you missed my feedback and I would like to address this in a PR.

> The TeaserGrid is a separate story. Could you please put the teaser grid into the teaser folder and add a story to the teaser stories? They both will never exist in isolation so please marry them 👰‍♀️🤵‍♂️

fischeversenker (approved):
Thanks for applying all of the feedback. Let's merge this :rocket:

georgiee:
I'm moving forward amd merge as @hofmann-felix is busy today.

PR #265, opened 2021-12-13
chore: make checkbox use Icon Component

milljoniaer:
### What is included?

this PR makes the Checkbox use the Icon Component to show the checkmark


georgiee (approved):
thanks, much better!

PR #266, opened 2021-12-13
refactor(office): rename to about-us

### LINKED ISSUE #259: Office -> About us
We are missing some important content on the page: who are we and how do we work. I think that can be one page where we can show a little bit about the office but mainly focus on the team: `About us` instead of `Office`.

I would really like to try to make this happen. Let's talk.

milljoniaer:
### What is included?

- Rename Gatsby Page
- rename all the Page Building files
- rename translation keys
- rename rename translation values
- rename Navigation

---

closes #259


src/locales/de/translations.json
```diff
@@ -103,13 +103,15 @@
...
-  "office.leadbox.link": "Karriere",
+  "about-us.title": "Über Uns",
+  "about-us.office.title": "Office",
+  "about-us.office.heading": "Unser neues Office direkt am Sendlinger Tor",
+  "about-us.office.text": "Fühlen Sie sich bei uns wie zu Hause. Ein paar kleine Eindrücke unseres neuen Büros in der Sendlinger Straße im Herzen Münchens.",
+  "about-us.team.title": "Team",
+  "about-us.team.heading": "Das sind wir",
+  "about-us.team.text": "Unser erfahrenes Team aus Programmierern und Designern",
+  "about-us.leadbox.title": "Haben Sie Interesse bekommen?",
+  "about-us.leadbox.link": "Karriere",
```
georgiee: I'm not happy about the dash but I wonder if an underscore would help much here. Technically fine so thanks!

georgiee (approved):
Please rename the stories as well:

`src/page-building/about-us/office.stories.mdx` -> src/page-building/about-us/about_us.stories.mdx` or so?

Also the story name should be "About us" not "office"
Screen Shot 2021-12-13 at 18 26 25


I already approve, mayeb you can change that.

src/locales/en/translations.json
```diff
@@ -103,13 +103,15 @@
...
-  "office.leadbox.title": "Did you get interested?",
-  "office.leadbox.link": "Career",
+  "about-us.title": "About Us",
+  "about-us.office.title": "Office",
+  "about-us.office.heading": "Our new Office directly next to Sendlinger Tor",
+  "about-us.office.text": "Feel at home with us. A small impression of our office in Sendlinger Straße in the heart of Munich.",
+  "about-us.team.title": "Team",
+  "about-us.team.heading": "That is who we are",
+  "about-us.team.text": "Our experienced team of engineers and designers",
+  "about-us.leadbox.title": "Did you get interested?",
```
fischeversenker: I know this is not necessarily related to this PR but these translations seem fairly "holprig". I suggest "This is who we are" instead of "That is who we are" as well as "Are you interested?" instead of "Did you get interested?":
```suggestion
  "about-us.team.heading": "This is who we are",
  "about-us.team.text": "Our experienced team of engineers and designers",
  "about-us.leadbox.title": "Are you interested?",
```

Some background:
- "That" implies distance. "This is my body, that is your body". "This" is close, "that" is distant. In this case the images of the team are right below the text so there is no distance thus we should be using "this".
- "Did you get interested?" is Denglisch. "Are you interested?" is the proper translation in my opinion.

fischeversenker (approved):
Thanks for this! :pray:

Everything looks great but I noticed a few janky translations, please see my comments. These translations are not related to this PR so they don't block this PR. I'm approving this but I'd appreciate it if you considered my suggestions 👍

src/locales/en/translations.json
georgiee: thanks for the catches. right now DE is the primary language and we plan to send the texts to a translator once the german version is approved/ready. But I still would like to fix things like this so I very appreciate your suggestion 👍 The translations could be weird because of deepl and myself looking over every translation but only to catche absolute non-sense. I guess this slipped.

Issue #251, opened 2021-12-09
Text Atoms & Markdown Content


georgiee:
We have no way to create single headlines or paragraphs easily because everything is grouped in bigger concepts right now. Let's provide components. Those components are also the ones we need for the markdown ast approach anyway🙏

+ Use the i18n link for [generated anchors](https://github.com/satellytes/satellytes.com/blob/35f2ef45b16cb351696e5493f66bcf0c4101ab6d/src/components/legacy/markdown/custom-components.tsx#L44-L56)

georgiee:
Things to consider when handling markdown: 

+ The markdown must go through the Gatsby pipeline so images can be processed properly
+ We use the already processed markdown and forward the ast presentation to our rehype processor defined in `MarkdownAst` via `<MarkdownAst htmlAst={markdown.htmlAst} />`
+ I have gone the way to define our very own prism parser in the past but I have been crawling back to `gatsby-remark-prismjs` I think because of the complexity to combine the existing markdown ast which things I do in the code parser (but I can't remember exactly what failed 😬)
+ It would be awesome to have support for custom tags in the markdown, something like `<my-component>` and I checked a few bits about that in the past. By disabling the XSS relevant checks in rehype this is possible but still limited the moment you nested those components. It would be nice to have something like that to output custom content because a code block opener like ` ``` ` or a simple image can't have any additional data while we would like to pass in attribution or things like attribution and such.

georgiee:
* I think we still can't embed easily videos (youtube and friends). While there are plenty of community plugins like `gatsby-remark-embed-youtube` I think this would be better to implement ourselves to gain control over the privacy (pre-fetch preview image so users don't load directly from youtube etc). Embed iframe only after a click etc.
* We don't have an easy way to attribute images in the content. What's the markdown formtat for that we can use and map with rehype ?

georgiee:
This is obsolete. We have worked on delivering many new components already.

PR #267, opened 2021-12-13
refactor(let paragraph in career inherit font)

milljoniaer:
### Why?

I recognized, that the Text of the Section Header on the Career Page was to small in Desktop Mode compared to Figma and the other Headers,

### What is included?

The local Paragraph Component inherits the font from its parent the Section Header, in that way it is responsive and has in each breakpoint the correct size.


src/page-building/career/career-page.tsx
```diff
@@ -14,7 +14,6 @@ import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { SyPersonioJob } from '../../types';
 
 const Paragraph = styled.p`
-  ${TextStyles.textR}
```
fischeversenker, line 17: Thanks for removing this. The following styles (space between adjacent paragraphs) is not needed as well. The template for this component only contains a single `<Paragraph>` so there is no point in keeping these styles. I suggest to completely remove this `Paragraph` and instead use a simple `<p>` in the template.

milljoniaer: I am not sure, but I think the Plan was to add multiple paragraphs.
For now there is only one, so I left it out. If we add more, we have to change it again, but I think it is less enough to remove it for now.

milljoniaer:
I changed it just like Felix suggested it, please have another look at the PR.

src/page-building/career/career-page.tsx
georgiee: I have removed those paragraphs in some PR before. While this change is okay I'm worrying about text paragraphs in general. We are not capable of freely writing texts right now 😕 but I think we can talk about that in https://github.com/satellytes/satellytes.com/issues/251

georgiee (approved):
let's merge, thanks

fischeversenker (approved):
Thanks for the discussion and the changes 🙌

PR #268, opened 2021-12-13
refactor(landingpage translation & content)

milljoniaer:
### Why?

- I found a german text on the english version of the landingpage.
- The wrong teasers are used on the landingpage for the services

### What is included?

- The german text on the english translation json is replaced with an english one.
- The correct translation keys for the teasers are used


src/locales/en/translations.json
```diff
@@ -11,7 +11,7 @@
   "main.services.teasers.third.text": "Fast feedback, short decision-making processes and open mindedness are important to us - whether you call it Scrum, Kanban or XP.",
   "main.career.title": "Work with us",
   "main.career.kicker": "Career at Satellytes",
-  "main.career.text": "Wir suchen Entwickler:innen aus Leidenschaft! Du hast noch nicht viel Berufserfahrung? Kein Problem. Denn alles, was du wissen musst, lernst Du bei uns. Du kannst schon alles? Dann findest Du bei Satellytes neue Herausforderungen und erfahrene Kollegen, mit denen Du weiter wachsen kannst. Schaue Dir unsere offenen Stellen an. Wir freuen uns auf Deine Bewerbung.",
+  "main.career.text": "We are searching for passionated Developers! You do not have much Experience? No Problem. You will learn everything you need to know. You are already an Expert? You will get opportunities and challenges to grow further with your experienced Colleagues. Have a look at our open situations. We are looking forward to receive your application.",
```
georgiee, line 14: thanks for the catch. here the deepl translation (https://www.deepl.com/translator) which sounds more natural and with the correct case (you don't write `Developers` with a cap `D`).

```suggestion
  "main.career.text": "We are looking for developers with passion! You don't have much work experience yet? No problem. Because you will learn everything you need to know with us. You can already do everything? Then you will find new challenges and experienced colleagues at Satellytes with whom you can grow. Take a look at our open positions. We look forward to receiving your application.",
```

georgiee (approved):
thanks for the changes. please take in my suggestion then we can merge

PR #271, opened 2021-12-14
refactoring(Icons): replace remaining icons with new Icon component

hofmann-felix:
Resolves issue from https://github.com/satellytes/satellytes.com/issues/254
Storybook: https://deploy-preview-271--satellytes-website-storybook.netlify.app
Preview URL: https://satellytescommain21751-choreicons.gtsb.io/

# Description
The goal of this PR is to replace all `.tsx` icons from `components/icons/...` with the new Icon component. Following icons changed: 
- `next-arrow.tsx` and `prev-arrow.tsx`: Only used in the Pagination component. The two `tsx` files were deleted and replaced with new Icon component in the pagination
- `chevron.tsx`: Only used in the LanguageSwitch component (Deleted and replaced too)
- `swoosh.tsx`: Used in the header and in the old typography -> Since the old typography is still used in some places I decided to keep this file

Remaining icons in `components/icons/...` 
- all social icons: I'm not sure how to deal with them. 
- `arrow.tsx`: will be removed in https://github.com/satellytes/satellytes.com/pull/261

# Jest Problems
Jest doesn't know how to load svg files. That's why the header unit test failed. I found a solution for this issue here: https://stackoverflow.com/questions/58603201/jest-cannot-load-svg-file. 
It works but I'm not sure if there is any better solution. 

Log of the failed unit test: https://github.com/satellytes/satellytes.com/runs/4519284825?check_suite_focus=true
FAIL src/components/header/header.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/runner/work/satellytes.com/satellytes.com/src/icons/theme-switch-24.svg:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){<svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
                                                                                      ^

    SyntaxError: Unexpected token '<'

    > 1 | export { ReactComponent as theme_switch } from '../../icons/theme-switch-24.svg';
        | ^
      2 | export { ReactComponent as checkmark_bold } from '../../icons/checkmark-bold-24.svg';
      3 | export { ReactComponent as chevron_down } from '../../icons/chevron-down-24.svg';
      4 | export { ReactComponent as chevron_left } from '../../icons/chevron-left-24.svg';

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (src/new-components/icon/icon-files.tsx:1:1)


georgiee:
For Jest: Delete the tests. We don't need them right now and  they have a low quality anyway.

jest.config.js
```diff
@@ -26,4 +26,8 @@ module.exports = {
   moduleNameMapper: {
     '\\.(css|less|woff|woff2|ttf|eot)$': 'identity-obj-proxy',
   },
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.svg$': '<rootDir>/svgTransform.js',
+  },
```
georgiee, line 32: I think that's a good solution for jest 👍

src/components/language-switch/language-switch.tsx
```diff
@@ -88,7 +87,7 @@ export const LanguageSwitch = ({
       $lightTheme={$lightTheme}
       fromNavigation={fromNavigation}
     >
-      <StyledChevron />
+      <StyledIcon show="caret_squared_down" />
```
georgiee, line 90: good to see our icons getting more and more aligned 👍

georgiee (approved):
Let's merge this. Looks good in storybook (checked the icons & pagination) and also checked the page (starring at the language select). Thank you🙌

PR #272, opened 2021-12-14
refactor: heroes, images & layout

georgiee:
This is a bigger PR due to the heavy refactoring. Whenever I create so large PRs I would love to use [Git stack](https://stacked-git.github.io/) on GitHub to provide two PRs that are clearly stacked on each other. I have created two main commits (one for heroes the other for the images/layout changes) but then decided to ship everything with this PR.

A shallow review is good enough. I appreciate any review that goes a little in details too. Thank you 🌟

+ [Storybook Preview: Heroes](https://deploy-preview-272--satellytes-website-storybook.netlify.app/?path=/docs/layout-heroes--page)
+ [Preview Website](https://satellytescommain21751-gkrefactorimages2.gtsb.io/). Visually almost nothing (ex career details hero) should have changed. ideal contribution when being in a rush to get this thing online 😬 But I think this change is worth it to make the foundation much more reliable 💪

---

I have walked through all pages to craft a pleasant experience for us developers to work with images & heroes. Here is what I have done. Apologize for the big PR, that's the nature of such a refactoring across many pages. I'm delighted about the result and I hope you share the same feeling.

1. Dedicated hero components
We have now two dedicated hero components: `ImageHero` & `AuroraHero`. Both accept the same text properties (kicker, headline, text) plus the background is either made of the aurora effect or a Gatsby Image (to benefit from all the nice image features like srcset and SSR). The heroes are described within a story `Layout/Heroes`. I opened that section as I would like to describe how to build proper page layouts within that part of the storybook, because right now it's implicit & learn from the existing stuff.

2. Updated all pages
I have updated all pages that need a hero (excluding the blogo).

3. New hero layout
I have asked the design team if we can put text into image heroes, because otherwise we are wasting a lot of space. It was well received and I got one example of text with a dimmed background (for readability). I have updated the career details to use that layout as a first example.

4. Better images for landingpage
The spacing image were direct imports (no image processing at all). I used my learnings from the aurora images to query for all office images and use a special layout for the full width + cover.

5. Refactor layoutV2
While working on the aurora I tinkered around with the layoutV2 as the hero depends on the given environment. I moved some functionalities into separate files (like `useAnchorTagScrolling`) to clear up the clutter and then I focused on the actual layout. The most important thing I changed here is the `grid-template-columns` configuration. I use a more versatile statements now and got rid of the involved `calc`.

## Gatsby Image Insights
The core of the problem with the heroes and images: How can I spread a `GatsbyImage` to the container. There is no simple configuration and I googled the entire internet with no obvious solutions.

I tried out a lot with GatsbyImages and went very different routes. My intuitive try to just update the css with `position: absolute` and friends did not work. I then looked into [gatsby-background-image](https://www.gatsbyjs.com/plugins/gatsby-background-image/) which is heavily outdated so now insights. I then tried to work with the raw srcSets given by the graphql query but I didn't dare to work to the end mainly because of concerns with SSR. There is a reason why [GatsbyImage](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-image/src/components/gatsby-image.browser.tsx) is super complex.

I then went back to the drawing board aka the first intuition with css and found this remarkable small note:

> This is an example of a hero image component with text overlaying an image background. It uses CSS grid to stack the elements on top of each other.

on this [page](https://www.gatsbyjs.com/docs/how-to/images-and-media/using-gatsby-plugin-image/).

The example showed this code (simplified) and I extracted the idea of creating a grid and  then assigning the children to the one and only grid area: `gridArea: 1/1`. That way it will spread across the available space. This is basically `position: absolute; left:0; right:0 and so on`. The only difference: This works while the position absolute did not work before, probably because some css statements are overlapping or so? Anyway this is much more elegant and comes straight into my toolbox of awesome css stuff 🙌

```

export function Hero() {
  return (
    <div style={{ display: "grid" }}>
      <GatsbyImage
        style={{
          gridArea: "1/1"
        }}
        layout="fullWidth"
        ...
      />
      <div
        style={{
          // By using the same grid area for both, they are stacked on top of each other
          gridArea: "1/1",
          display: "grid",
        }}
      >
        {/* Any content here will be centered in the component */}
        <h1>Hero text</h1>
      </div>
    </div>
  )
}

```


src/components/layout/layout-v2.tsx
```diff
@@ -35,68 +41,32 @@ const Main = styled.main`
    */
   display: grid;
   grid-template-columns:
-    1fr
-    min(821px, calc(100% - 2 * 24px))
-    1fr;
+    [main-start] minmax(24px, 1fr)
+    [content-start] minmax(0, 820px) [content-end]
+    minmax(24px, 1fr) [main-end];
```
georgiee, line 46: @feedm3 I have refactored this one to let the grid do the calculations. I developed that within a hero and though it's worth to replace in the layout (admittedly not much easier though)

In addition created named tracks to assign the content to names and not numbers (`grid-column: content`)

src/components/layout/layout-v2.tsx
```diff
@@ -35,68 +41,32 @@ const Main = styled.main`
...
-
-  ${up('md')} {
-    height: 640px;
-  }
-`;
-
 const BreadcrumbContainer = styled.div<{ hero: boolean }>`
   margin: ${(props) => !props.hero && `calc(${HEADER_HEIGHT} + 16px)`} 24px 16px;
 `;
-const useAnchorTagScrolling = (): void => {
```
georgiee, line 79: separate file because it's cluttering the view and deserves its own file for sure

src/components/layout/layout-v2.tsx
```diff
@@ -111,31 +81,6 @@ interface LayoutProps {
...
-
-  if (isLight) {
-    document.body.classList.remove(classNameDark);
-    document.body.classList.add(classNameLight);
-  } else {
-    document.body.classList.remove(classNameLight);
-    document.body.classList.add(classNameDark);
-  }
-}
-
```
georgiee, line 138: separate file because it's cluttering the view and deserves its own file for sure

src/components/layout/layout-v2.tsx
```diff
@@ -156,7 +101,6 @@ function overrideDarkFromQuery() {
 
 export const LayoutV2 = ({
   transparentHeader,
-  heroImage,
```
georgiee, line 159: good bye. We have not used this for quite some and it's much better (for us) to pass in a dedicated component anyway.

src/components/layout/layout-v2.tsx
```diff
@@ -180,12 +123,12 @@ export const LayoutV2 = ({
           siteTitle="Satellytes"
           siteTitleUrl={siteTitleUrl}
           $lightTheme={isLight}
-          transparent={transparentHeader || Boolean(heroImage)}
+          transparent={transparentHeader}
```
georgiee, line 126: We have been very explicit with that before so this was a no brainer once I decided to remove heroImage

src/new-components/heroes/aurora-hero.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const AuroraFullSize = styled(Aurora)`
+  grid-area: 1/1;
+`;
+
+interface AuroraHeroProps extends HeroWithText {
+  auroraType?: AuroraType;
+}
+
+export const AuroraHero = ({
```
georgiee, line 15: Welcome `AuroraHero` 🌟

src/new-components/heroes/hero-text.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+`;
+
+const Text = styled.div`
+  ${TextStyles.textR}
+
+  ${up('md')} {
+    ${TextStyles.textL}
+  }
+`;
```
georgiee, line 55: This is the text component used in both heroes. Very similar to section headline but I want a `h1` as the headline and the text needs to be white all the time. So I copied most content.

src/new-components/heroes/heroes.stories.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+## Different Aurora type
+<Canvas withSource="open">
+  <AuroraHero title="Satellytes" auroraType={AuroraType.Pink}>
+    Welcome folks!
+  </AuroraHero>
+</Canvas>
+
+# Image Hero
+Due to the gatsby image dependency no preview is available.
```
georgiee, line 25: That is so sad. The storybook is half useful when we are not able to preview images that rely on gatsby image 😞

src/new-components/heroes/heroes.stories.mdx
```diff
@@ -0,0 +1,59 @@
+import { Canvas, Meta } from '@storybook/addon-docs';
+import { AuroraHero } from './aurora-hero'; import { AuroraType } from '../../components/aurora/aurora';
+
+<Meta
+  title="Layout/Heroes" />
+
+# Aurora Hero
```
georgiee, line 7: I like to showcase the heroes in a distinct page. Hence this page without any stories but only some examples. We would have two stories (aurora + images) but the image story is useless so I spared the stories at all.

src/new-components/heroes/image-hero.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const ImageHero = ({
+  image,
+  title,
+  children,
+  kicker,
+}: ImageHeroProps) => {
+  const gatsbyImageData = getImage(image);
+
+  return (
+    <HeroContainer>
```
georgiee, line 31: Welcome `ImageHero` 🌟 You will save so much brain power in the future because you should just work

src/new-components/heroes/index.tsx
```diff
@@ -0,0 +1,2 @@
+export { ImageHero } from './image-hero';
+export { AuroraHero } from './aurora-hero';
```
georgiee, line 2: I had the feeling this makes sense for the heroes to define what is meant to be used on the outside. I'm not sure this will trigger someone of my fellow reviewers because we never used barrel files before 😬

src/new-components/heroes/support.tsx
```diff
@@ -0,0 +1,58 @@
...
+    grid-column: content;
+  }
+
+  align-items: end;
+
+  ${({ dimmed }) =>
+    dimmed &&
+    css`
+      background-color: ${rgba('#000000', 0.2)};
+    `}
```
georgiee, line 57: Based on a first experiment in figma, the text background can be dimmed for the text being readable:

Screen Shot 2021-12-14 at 13 57 05

src/new-components/heroes/support.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * within the hero aligned to the main content (by using the same grid template)
+ * The background can be optionally dimmed/
+ */
+export const TextContainer = styled.div<TextContainerProps>`
+  grid-area: 1/1;
+  display: grid;
+  grid-template-columns:
+    [main-start] minmax(24px, 1fr)
+    [content-start] minmax(0, 820px) [content-end]
+    minmax(24px, 1fr) [main-end];
```
georgiee, line 45: this is the same as the layout to provide the same alignment with the content

.storybook/preview.js
```diff
@@ -13,7 +14,25 @@ addDecorator((storyFn) => (
...
+const DocsPageContainer = ({ children, context }) => {
+  return (
+    <DocsContainer context={context}>
+      <ThemeProvider theme={theme}>
+        <GlobalStyle $lightTheme={true} />
+        {children}
+      </ThemeProvider>
+    </DocsContainer>
+  );
+};
```
georgiee, line 31: plain mdx files without a story won't have our global styles applied. This is the fix👍

src/components/header/header.tsx
```diff
@@ -8,8 +8,9 @@ import { NavigationFlyout } from './menu-flyout';
 import { Link } from '../links/links';
 import { Swoosh } from '../icons/swoosh';
 import { LanguageSwitch } from '../language-switch/language-switch';
+import { HEADER_HEIGHT_VALUE } from '../layout/theme';
 
-export const HEADER_HEIGHT = '65px';
+export const HEADER_HEIGHT = `${HEADER_HEIGHT_VALUE}px`;
```
georgiee, line 13: I placed some core numbers into the theme. One is the header height as a number which is exposed here. I didn't want to refactor `HEADER_HEIGHT` so I just left it as is but made it dependent on the new integer from the theme.

src/components/layout/layout-v2.tsx
```diff
@@ -23,6 +24,11 @@ const FullHeightContainer = styled.div`
   flex-direction: column;
 
   min-height: 100vh;
+
+  padding-top: ${HEADER_HEIGHT_VALUE}px;
+  ${up('md')} {
+    padding-top: ${HEADER_HEIGHT_VALUE}px;
+  }
```
georgiee, line 31: I have the idea that the layout MUST work without any container. For this to work we have to provide counter paddings for the fixed navigation/header. They both belong together so we can assume they will always be on a page. Without the padding I'm not able to use the layout with some container with is a big red flag for me. Hence the idea of making the layout independent with this. This brings other problems: The container we use to wrap the content & sections will always apply the top spacing and add up. I didn't want to rely on collapsing margin's (too hard to follow and some don't even know about that). That's why I introduced the `first-child` statement to make the container behave differently when it's the very first element in the layout (assuming it's always placed in the layout directly). See the [container component](https://github.com/satellytes/satellytes.com/pull/272/files#diff-74b31a29bc40787564bd5df1ddd53394b1830233566d0a4790158fa57c36bafd) for the first-child thing.

src/page-building/landingpage/image-spacer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const gatsbyImageData = getImage(image);
+
+  return (
+    <ImageContainer>
+      {gatsbyImageData && (
+        <GatsbyImage style={coverContainerCss} alt="" image={gatsbyImageData} />
+      )}
+    </ImageContainer>
+  );
+};
```
georgiee, line 45: I think you can see how I came from the heroes for this. It's working great and we have now properly loaded images on the landingpage actings as spacers.
Screen Shot 2021-12-14 at 14 08 27

src/page-building/landingpage/landingpage.tsx
```diff
@@ -1,111 +1,57 @@
...
+
+type OfficeImages = { [relativePath: string]: OfficeImage };
 
 interface LandingpageProps {
   positions: SyPersonioJob[];
   posts: BlogPostTeaser[];
+  officeImages: OfficeImages;
 }
 
-const HeroContainer = styled.div`
```
georgiee, line 22: good by you clutter

src/pages/career.tsx
```diff
@@ -44,6 +49,12 @@ export default Career;
 
 export const CareerPageQuery = graphql`
   query ($language: String!) {
+    hero: file(relativePath: { eq: "office/sy-office-01.jpg" }) {
+      childImageSharp {
+        gatsbyImageData(layout: FULL_WIDTH)
```
georgiee, line 54: `FULL_WIDTH` is the only thing we need. The default props include a set of breakpoints which means we get the image in different srcSets while being set to display as 100%.

src/pages/imprint.tsx
```diff
@@ -35,11 +35,11 @@ const ImprintPage = ({ data, location }: ImprintPageProps) => {
         location={location}
         noIndex={true}
       />
-      <ContentBlockContainerWithoutHero>
+      <ContentBlockContainer>
         <SectionHeader headline={t('navigation.imprint')} />
         <MarkdownAst htmlAst={data.markdownRemark.htmlAst} />
         <BottomNote>{t('imprint.updated')}</BottomNote>
-      </ContentBlockContainerWithoutHero>
+      </ContentBlockContainer>
```
georgiee, line 42: This is the reason I included the layout/container refactoring in this PR. We expected the users to tell if there is a hero or not because the layout did not allow the content to be displayed independent of the hero. I changed that with some default spacing and a proper content block container that counts in the fact, when it's the first child of the layout please don't add to much of spacing.

src/pages/index.tsx
```diff
@@ -14,18 +22,29 @@ interface IndexPageProps {
...
   location: Location;
 }
 
 const IndexPage = (props: IndexPageProps) => {
   const jobPositions = props.data.allSyPersonioJob.nodes;
   const blogPosts = props.data.allMarkdownRemark.nodes;
+  const officeImages = props.data.officeImages.nodes.reduce((memo, image) => {
+    memo[image.relativePath] = image;
+    return memo;
+  }, {});
```
georgiee, line 38: I want to work with a map of images so I can access them in the actual landingpage with `officeImages["office/01.png"]` and such. We need that because we pick selectors and we don't just show every office image.

src/page-building/landingpage/image-spacer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  /**
+    We enforce a higher specificity with \`&&\` (repeat the classname two times) to overrule the layout main rule of assigning any content to a specific column \`> *{  grid-column: content; }\` in the grid
+    which we want to escape to spread over the full width.
+    
+    Do not remove easily. The actual order of CSS rules is not stable
+    and the two rules (from the layout and this) will have the same specificity. 
+   */
+  && {
+    grid-column: -1/1;
+  }
```
georgiee, line 20: > We enforce a higher specificity with `&&` (repeat the classname two times, `.my-class.my-class`) to overrule the layout main rule of assigning any content to a specific column `> *{  grid-column: content; }` in the grid which we want to override to spread over the full width.
    
> Do not remove easily. The actual order of CSS rules is not stable and the two rules (from the layout and this) will have the same specificity. Sometimes it works, sometimes now if removed.

src/components/layout/layout-v2.tsx
```diff
@@ -156,7 +101,6 @@ function overrideDarkFromQuery() {
 
 export const LayoutV2 = ({
```
feedm3, line 102: Is the legacy layout still used in some page? If not, I guess it's a good time to delete the old layout and  rename this file and component to `layout.tsx`

feedm3 (approved):
I opened the preview and tested all pages and the new teaser grid. Everything still works as expected and has the proper size/margins. The new structure and the separate components and util functions are new better than before. Like it! Also appreciate the documentation/comments you wrote, it was pleasant to check this PR 🙂

src/components/layout/layout-v2.tsx
georgiee: the blog is pending. once migrate we can remove the old layout and rename this one finally.

georgiee:
Thank you for your review @feedm3 ! very much appreciated 🙏

PR #273, opened 2021-12-15
chore: update blog

georgiee:
I have migrated the blog and blog post pages. There are no legacy layouts left so I removed it too.
The blog overview design changed. It's more or less correct according to the design but I think we waste a lot of screen estate. I spend too much time solving this adhoc with css grid so I stopped myself to provide this first update only. It looks good enough and it's a pleasure as usually to split the gatsby clutter from the actual page.

I also split the blog post template into a proper page component. It's much better to handle now. I did only change the most urgent things. In addition I added a `BlogHero` which can handle two images (square and original) which we had before. A pretty nice idea I forgot we have implemented. Hence this third hero. I deliberately duplicated code instead of marrying those features with the image hero. There is room for some refactoring for sure but it's good enough 👍

I also removed some obvious components related to the blog. I will do a clean up following this PR, check other components and I try to reorganize things if possible.


georgiee:
I actually liked to highlight the latest two blog posts and mayeb we can restore that behavior with a local grid easily.

Screen Shot 2021-12-15 at 14 54 21

src/new-components/heroes/blog-hero.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <HeroText title={title} kicker={kicker}>
+            {children}
+          </HeroText>
+        </TextContainer>
+      )}
+
+      {attribution && <Attribution attribution={attribution} />}
+    </HeroContainer>
+  );
+};
```
fischeversenker, line 83: It would be nice to have a story for this new hero. While looking for it I noticed the current `Hero` stories live at `Layout/Heros`... aren't they regular components? I also noticed that the `Code` ("Canvas") view in the Hero story is not working... not sure what's up with that.

Long story short: Could you add a short section about this new hero to the existing hero stories?

src/page-building/blog/blog-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <Layout light showLanguageSwitch={false} breadcrumb={BREADCRUMB}>
+      <ExtraWideContainer>
+        <BlogHeader headline={t('navigation.blog')}>
+          {t('blog.info')}
+        </BlogHeader>
+      </ExtraWideContainer>
+
+      <ExtraWideContainer>
+        {language != 'en' && <OnlyEnglishNotice />}
```
fischeversenker: This `<OnlyEnglishNotice>` looks out of place. Can you take a look and fix this, please?

image

src/pages/blog.tsx
```diff
@@ -45,73 +38,19 @@ interface BlogPageProps {
...
-          <BlogPageTitle>{t('navigation.blog')}</BlogPageTitle>
-        </GridItem>
-        <GridItem md={8}>
-          <BlogPageSubTitle>{t('blog.info')}</BlogPageSubTitle>
-          {language != 'en' && <OnlyEnglishInfo onClick={onClick} />}
-        </GridItem>
-      </Grid>
-      {language == 'en' && <BlogPostOverview blogPosts={blogPosts} />}
-    </Layout>
+      <BlogPage posts={blogPosts} />
```
fischeversenker, line 48: Nice :rocket: :sparkles:

fischeversenker (changes requested):
Nice! A lot of deleted code. Me gusto!

I left a few comments. The one regarding the `OnlyEnglishNotice` is a blocker to me. Can you take care of that before we merge this, please?

src/page-building/blog/blog-page.tsx
georgiee: oh I did not check this for a while.

src/new-components/heroes/blog-hero.tsx
georgiee: `Layout/Hero` because it's made for page building. I would like to group the thing we need to create pages (heroes, containers, sections or so). That's why I just try things like that. This will change for sure. I will provide the missing story, which I forgot! Thanks for the ctahc.

src/page-building/blog/blog-page.tsx
georgiee: fixed, see https://satellytescommain21751-gkblog.gtsb.io/de/blog/

Screen Shot 2021-12-15 at 16 33 58

src/new-components/heroes/blog-hero.tsx
georgiee: https://deploy-preview-273--satellytes-website-storybook.netlify.app/?path=/docs/layout-heroes--page
👍
Screen Shot 2021-12-15 at 16 34 46

georgiee:
@fischeversenker 

I have created the missing story plus one for the blog page itself while being there.
Screen Shot 2021-12-15 at 16 35 13

In addition I fixed a build error as the page relies on `location` which I didn't pass in for the breadcrumb. Instead of passing it in I decided to build the breadcrumb data on the outside. I think that manual approach is okay but we should try to move out data and component step by step and this is a first step.

src/new-components/heroes/heroes.stories.mdx
```diff
@@ -57,3 +57,44 @@ optionally with headline and text.
...
 ```
+
+# Blog Hero
+> Due to the gatsby image dependency no preview is available.
+
+This hero behaves very similar to the `ImageHero` but you can pass in two additional things: `attribution` and a squared image
+that will be used once the website is very narrow (higher than wide) to present a more pleasant part from the feature image.
+
+
+```
```
fischeversenker:
```suggestion
```tsx
```

src/new-components/heroes/heroes.stories.mdx
```diff
@@ -57,3 +57,44 @@ optionally with headline and text.
...
+  }
+  image={featuredImage}
+  imageSquare={featuredImageSquared}
+/>
+```
+
+
+You can query for both images in a single graphql query. Pass `aspectRatio: 1` to get a square image.
+
+```
```
fischeversenker:
```suggestion
```graphql
```

fischeversenker (approved):
Thanks for the updates 👍 

I saw that in the markdown in the stories we missed out on the opportunity to provide some beautiful syntax highlighting. I left two suggestions for `tsx` and `graphql` but there are others in that same file where I wasn't able to provide suggestions. You might want to change these others locally or ignore the suggestions so we can take care of this for all stories at once some time in the future.

Apart from that there are a lot of warnings for unused imports or `forbidden non-null assertions` in this PR. I'm not sure which ones got introduced in this PR and which ones already existed before. They don't block this PR but you might want to have a look.

No blockers. Approving 👌

src/new-components/heroes/heroes.stories.mdx
georgiee: I did not expect to get colored code here, that's much better 😅 Thanks

PR #274, opened 2021-12-16
refactoring: integrate new Button component

hofmann-felix:
Resolves Issue from https://github.com/satellytes/satellytes.com/issues/254
Storybook: https://deploy-preview-274--satellytes-website-storybook.netlify.app/
Preview URL: https://satellytescommain21751-choreintegratebutton.gtsb.io/

### What's included?
- `<LinkButton />` is removed and replaced with the general `<Button />` component
- The `<Button />` component can now also receive a `to` prop. If this prop is passed, the button will be rendered as `a`, otherwise as `button`.

### DIFF
--- a/src/components/ui/buttons/button.stories.mdx
+++ b/src/components/ui/buttons/button.stories.mdx
@@ -16,7 +16,21 @@ This is the Regular Button, which should be used in all the standard use cases. 
 <Canvas withSource="open">
   <Story name="Regular">
     <Button onClick={() => console.log('Regular Button clicked')}>
-      Alle Services
+      Submit
+    </Button>
+  </Story>
+</Canvas>
+
+<ArgsTable of={Button}/>
+
+## Button vs. Anchors
+
+This component can be used both as a button and as an anchor. If the `to` prop is provided, the component is rendered as an anchor. Otherwise, the component is rendered as a button. Here you can read why this distinction is useful: https://bitsofco.de/anchors-vs-buttons/
+
+<Canvas withSource="open">
+  <Story name="Button as Link">
+    <Button to={'https://www.example.com/'}>
+      Link
     </Button>
   </Story>
 </Canvas>
--- a/src/components/content/leadbox/leadbox.tsx
+++ b/src/components/content/leadbox/leadbox.tsx
@@ -2,13 +2,14 @@ import React from 'react';
 import styled from 'styled-components';
 import { TextStyles } from '../../typography';
 import { theme } from '../../layout/theme';
-import { Link, LinkButton } from '../../legacy/links/links';
+import { Link } from '../../legacy/links/links';
 import { up } from '../../support/breakpoint';
 import {
   Illustration,
   IllustrationSize,
 } from '../../ui/illustration/illustration';
 import { IllustrationType } from '../../ui/illustration/illustration-set';
+import { Button } from '../buttons/button';
 
 export interface LeadContact {
   headline: string;
@@ -128,7 +129,7 @@ interface LeadLinkProps {
 }
 
 export const LeadLink = ({ link }: LeadLinkProps) => {
-  return <LinkButton to={link.href}>{link.title}</LinkButton>;
+  return <Button to={link.href}>{link.title}</Button>;
 };
 
 export const Leadbox = (props: LeadboxProps) => {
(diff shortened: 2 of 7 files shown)


src/new-components/buttons/button.tsx
```diff
@@ -25,29 +27,32 @@ const StyledButton = styled.button`
...
+        <StyledIcon show={'chevron_right'} size={IconSize.NORMAL} />
+      </StyledButton>
+    ) : (
+      <StyledButton as="button" {...props}>
+        <ButtonText>{props.children}</ButtonText>
+        <StyledIcon show={'chevron_right'} size={IconSize.NORMAL} />
+      </StyledButton>
+    )}
+  </>
+);
```
georgiee: I think this would be much more readable with an explicit if-else condition. The ternary operator always makes reading code harder.

```suggestion
export const Button = (props: RegularButtonProps) => {
  if (props.to) {
    return (
      <StyledButton to={props.to}>
        <ButtonText>{props.children}</ButtonText>
        <StyledIcon show={'chevron_right'} size={IconSize.NORMAL}/>
      </StyledButton>
    )
  } else {
    return (
      <StyledButton as="button" {...props}>
        <ButtonText>{props.children}</ButtonText>
        <StyledIcon show={'chevron_right'} size={IconSize.NORMAL}/>
      </StyledButton>
    )
  }
};

```

src/components/links/links.tsx
```diff
@@ -73,45 +72,3 @@ const ExternalLink = styled.a`
...
-    margin-left: 20px;
-  }
-  :hover {
-    svg {
-      transform: translateX(2px);
-    }
-  }
-`;
-
-export const LinkButton = (props: LinkProps): JSX.Element => {
```
georgiee: good bye 👋 much better to deal only with a Link or Button and let the components pick the right semantic i.e. `button` or `a` internally👍

src/new-components/buttons/button.tsx
```diff
@@ -25,29 +27,32 @@ const StyledButton = styled.button`
...
-      <StyledIcon show={'chevron_right'} size={IconSize.NORMAL} />
-    </StyledButton>
-  );
+type RegularButtonProps = React.ComponentPropsWithoutRef<'button'> & {
+  to?: string;
 };
+
+export const Button = (props: RegularButtonProps) => (
+  <>
+    {props.to ? (
```
georgiee: I'm not reading anything about that behavior in the story. Could you explain the fact in a section "Button vs. Anchors`. In addition to a few words by you, you can also link to some article explaining the difference. I didn't know the following article before but it looks a great explanation of this topic https://bitsofco.de/anchors-vs-buttons/

georgiee: I usually remember "anchor for links, button to trigger interactions" and that button support enter key presses by default which is important for a11y but also UX.

georgiee (changes requested):
Great change. Thank you. I have a few things to update. Can you check and provide a little addition to the button story and check my ternary suggestion ? Thank you!

georgiee:
by the way `refactoring` is not a correct keyword in your commit. This should be `refactor`. Absolutely nitpicking because we don't parse the commit history anyway. Just that you know 👍

src/components/ui/buttons/button.stories.mdx
```diff
@@ -16,7 +16,21 @@ This is the Regular Button, which should be used in all the standard use cases.
...
+
+<ArgsTable of={Button}/>
+
+## Button vs. Anchors
+
+This component can be used both as a button and as an anchor. If the `to` prop is provided, the component is rendered as an anchor. Otherwise, the component is rendered as a button. Here you can read why this distinction is useful: https://bitsofco.de/anchors-vs-buttons/
+
+<Canvas withSource="open">
+  <Story name="Button as Link">
+    <Button to={'https://www.example.com/'}>
```
georgiee, line 32: Thanks for using example.com 😎

src/components/ui/buttons/button.stories.mdx
```diff
@@ -16,7 +16,21 @@ This is the Regular Button, which should be used in all the standard use cases.
...
+      Submit
+    </Button>
+  </Story>
+</Canvas>
+
+<ArgsTable of={Button}/>
+
+## Button vs. Anchors
+
+This component can be used both as a button and as an anchor. If the `to` prop is provided, the component is rendered as an anchor. Otherwise, the component is rendered as a button. Here you can read why this distinction is useful: https://bitsofco.de/anchors-vs-buttons/
```
georgiee, line 28: That's good, thanks!

georgiee (commented):
let's merge this beauty

georgiee (approved):

PR #275, opened 2021-12-16
chore: add missing translation keys

hofmann-felix:
URL: https://satellytescommain21751-choremissingtranslation.gtsb.io/

In some buttons from the landingpage the text was hardcoded. I added translation and the corresponding keys to it. 

Bildschirmfoto 2021-12-16 um 16 21 01
Bildschirmfoto 2021-12-16 um 16 21 09


georgiee (approved):
thank you very much appreciated

Issue #278, opened 2021-12-16
SEO Issues 


georgiee:
1. Sharing
Some pages don't have any text or image, like `Services`. We should ensure that every page provides a distinct share image and description wherever it's possible. This involves generating a shared card for every page and providing a distinct short description.

Examples:
+ https://satellytes.com/about-us/
+ https://satellytes.com/services
+ https://satellytes.com/contact/

those pages don't have a dedicated card nor a dedicated sharing text and we should change that once the text are better.

2. SEO & Translations
I think we need dedicated translations keys for seo otherwise titles or such will change and will have side effects on seo without our knowledge.

3. Cards
We have cards for every page we can create. The pages that are created automatically don't have any cards. I'm not sure how to build the cards though. Probably `onCreatePage`, detect the pages and then create the card and add a node field with the card image.

georgiee:
SEO is monitored pretty good by Fabian and this issue doesn't have any actionable items left.

PR #279, opened 2021-12-16
fix: change services page meta title

hofmann-felix:
URL:  

The services page used the wrong page title. I changed the translation key in the component, the title is correct now: 
Bildschirmfoto 2021-12-16 um 15 51 51

### DIFF
--- a/src/pages/services.tsx
+++ b/src/pages/services.tsx
@@ -13,7 +13,7 @@ const ServicesPage = ({ location }: ServicesPageProps) => {
 
   return (
     <>
-      <SEO title={`${t('services.title')} | Satellytes`} location={location} />
+      <SEO title={`${t('services.hero')} | Satellytes`} location={location} />
       <Service />
     </>
   );


src/pages/services.tsx
```diff
@@ -13,7 +13,7 @@ const ServicesPage = ({ location }: ServicesPageProps) => {
 
   return (
     <>
-      <SEO title={`${t('services.title')} | Satellytes`} location={location} />
+      <SEO title={`${t('services.hero')} | Satellytes`} location={location} />
```
georgiee, line 16: I'm always torn between defining extra keys or sharing keys. Feels wrong to use the `hero` key here but it's good enough and better than `What we offer`. But the moment we decide to change the hero to something else the seo will silently change so we need to handle this. I have updated https://github.com/satellytes/satellytes.com/issues/278 to work on this later.

georgiee (approved):
thanks let's merge

PR #285, opened 2021-12-16
add new text

abdolg:


src/assets/locales/de/translations.json
```diff
@@ -1,8 +1,8 @@
 {
-  "main.description": "Wir sind pragmatische Profis, die zuverlässige Software für das Web entwickeln.",
-  "main.services.title": "Plattformen, Beratung & Produkte",
+  "main.description": "Wir machen die digitale Welt einfacher. Mit maßgeschneiderter Web-Software für Unternehmen. Von Fontend bis Backend. Vom Webauftritt bis zur digitalen Sicherheitsinfrastruktur.",
+  "main.services.title": "Wir verbinden. Mit Technologie.",
```
georgiee, line 3: 🙌

src/assets/locales/de/translations.json
```diff
@@ -114,9 +114,9 @@
   "about-us.leadbox.title": "Haben Sie Interesse bekommen?",
   "about-us.leadbox.link": "Karriere",
   "services.hero": "Leistungen",
-  "services.title": "Unser Angebot an Sie",
+  "services.title": "Wir verbinden Development & Design",
```
georgiee, line 120: This is different to `Wir verbinden. Mit Technologie.` (via `main.services.title"`) which we could just reuse here as on the start page?

src/assets/locales/de/translations.json
```diff
@@ -135,7 +135,7 @@
   "services.products_services.list.prototyping": "Prototyping",
   "services.products_services.list.ideation": "Ideation & Proof of Concepts",
   "services.products_services.list.development": "Product Design & Development",
-  "services.industries.text": "Derzeit bieten wir unsere Dienstleistungen in den folgenden Branchen an",
+  "services.industries.text": "Wir streben stets hochwertige und zeitgemäße Lösungen für verschiedene Branchen an – die Wahl der Technologie ist für uns dabei sekundär.",
```
georgiee, line 141: Awesome, thanks!
Screen Shot 2021-12-17 at 08 53 10

georgiee (approved):
I approve but I have one note that you might want to change?

I love this 💛
Screen Shot 2021-12-17 at 08 48 35




The landingpage service section and the service page itself now starts with different headers. Seems to be intentionally while I would just use the same phrase? See here:

Screen Shot 2021-12-17 at 08 54 55




----

FYI `no description provided` is the worst PR 🤓 

This PR is about the service content and the main hero by looking at the text changed. You could tell "I have updated the services textes" and in addition provide a deep link to the deployed page https://satellytescommain21751-ganewtext.gtsb.io/services/ or include a screenshot of content if it's a small area.

It's also only in Germany which is fine as we want to translate the content anyway. But should we keep the outdated text. Is that text okay compared to the new text in German?

This helps tremendously to make the review efficient 🙏

src/assets/locales/de/translations.json
```diff
@@ -35,7 +35,7 @@
   "career.privacy-policy": "<0>Hiermit bestätige ich, dass ich die <1>Datenschutzerklärung</1> zur Kenntnis genommen habe. <4>*</4></0>",
   "career.introduction.headline": "Arbeite mit uns",
   "career.introduction.kicker": "Karriere bei Satellytes",
-  "career.introduction.paragraphs.0": "Wir sind Satellytes, ein Team aus sehr erfahrenen Entwicklern und einigen Designern. Wir bringen großen Unternehmen nachhaltige, barrierefreie, verlässliche und moderne Lösungen rund ums Web. Wenn du Interesse hast, bei uns mitzumachen, dann schau dir gleich unserer offenen Positionen an.",
+  "career.introduction.paragraphs.0": "Wir sind Satellytes, ein Team aus sehr erfahrenen Software Ingenieuren und einigen Designern. Wir bringen großen Unternehmen nachhaltige, barrierefreie, verlässliche und moderne Lösungen rund ums Web. Wenn du Interesse hast, bei uns mitzumachen, dann schau dir gleich unserer offenen Positionen an.",
```
milljoniaer, line 41: I am not quite sure, but I think this sounds a bit weird, doesn't it? It seems that we have "(sehr erfahrene Software Ingenieure)" and on the other hand "(einige Designer)". What do you think about leaving out the "einige" to have "(sehr erfahrene Software Ingenieure und Designer)"?

src/assets/locales/de/translations.json
```diff
@@ -1,8 +1,8 @@
 {
-  "main.description": "Wir sind pragmatische Profis, die zuverlässige Software für das Web entwickeln.",
-  "main.services.title": "Plattformen, Beratung & Produkte",
+  "main.description": "Wir machen die digitale Welt einfacher. Mit maßgeschneiderter Web-Software für Unternehmen. Von Fontend bis Backend. Vom Webauftritt bis zur digitalen Sicherheitsinfrastruktur.",
```
milljoniaer, line 2: Seems to be a typo
```suggestion
  "main.description": "Wir machen die digitale Welt einfacher. Mit maßgeschneiderter Web-Software für Unternehmen. Von Frontend bis Backend. Vom Webauftritt bis zur digitalen Sicherheitsinfrastruktur.",
```

PR #287, opened 2021-12-17
fix: wrong url generation

georgiee:
Preview (the two DE pages with links to career positions)

https://satellytescommain21751-gkfixlinks.gtsb.io/de/
https://satellytescommain21751-gkfixlinks.gtsb.io/de/career

# Summary

This will fix the wrong urls on the landingpage by introducing the following changes:
+ Removal of the career path slug generation without any replacement
+ Update the slug
+ Use the slug directly for the linls

# Details
+ We will not create a path `field` and insert it into the sy job item. That value is only required when the career page is created.
+ Instead we need a proper slug. Instead of defining an additional field (like with path before) I decided to update the slug coming from personio to a value that matches our installation. `frontend-engineer` gets transformed to `/career/frontend-engineer/`. I include the leading and trailing slashes as this value can then be used together with the Link component from the i18n plugin.
+ I updated the landingpage and the career overview page to use the slug
+ We can configure the generated prefix in the plugin (to keep the plugin a little bit agnostic)

```
{
  resolve: `sy-personio-jobs-source-plugin`,
  options: {
    slugPrefix: `career`,
  },
},
```

+ I also fixed the blog post as discussed: Hard code the language. I still think that's the best solution. There is no other or simpler way to tell the i18n plugin that there is no translation.


georgiee:
auto merge enabled for the responsible person who can review this (also shallow by checking the result is fine I think)

feedm3 (approved):
I checked all links and pages, everything works as expected. Well done 👍

PR #288, opened 2021-12-17
fix: use TeaserGrid Component in the career/openings

milljoniaer:
The TeaserGrid was not used in Career/Openings. With the Teaser Grid four or more Teasers are always displayed in two 
columns. Thats why it looks now like this:

image

and not like this:
 
image


src/components/pages/career/openings.tsx
```diff
@@ -5,18 +5,7 @@ import { up } from '../../support/breakpoint';
...
-  gap: 24px;
-
-  justify-items: stretch;
-  grid-template-columns: repeat(auto-fit, 250px);
-  ${up('md')} {
-    gap: 70px;
-    grid-template-columns: 1fr 1fr 1fr;
-  }
-`;
+import { TeaserGrid } from '../../content/teaser/teaser-grid';
```
feedm3, line 7: Can you please remove the unused `up` import?

feedm3 (approved):
Nice fix! There's an unused import now, but I already approve in advance so you can merge once removed this import.

src/components/pages/career/openings.tsx
milljoniaer: Oh yes, thank you, I removed it 👍

PR #289, opened 2021-12-17
chore: change service teasers on main

milljoniaer:
This changes the Teaser textes on the main page. It uses the `service.platform.title` tag and the `service.platform.teaser` tag. 

The old main teaser tags in the translation files are removed.


feedm3 (approved):

fischeversenker (approved):
Looks a lot cleaner now, thanks! Maybe even a little too clean 😂 
Let's merge and improve if necessary :sparkles:

PR #296, opened 2021-12-17
refactor: make meta tags more solid

georgiee:
Initially I wanted to write a little blog post but then I entered the rabbit hole of meta tags and here we are 😅
This PR brings:
+ Removed duplicate meta tags. We can safely remove the twitter image tags as we correctly implement the open graph tags already
+ We had some wrong meta tags like `property="name"` which don't exist (property is almost exclusively used for open graph and there is no `name`
+ I replaced the hard to read `listLocalizedVersions` with a more readable version plus this fixes the fact, that we have never output any alternate tags for our default pages as far as I tested. The new versions fully rely on the i18n list of available languages. 
+ The i18n available languages do not work for our jobs. That's where a lot of technical debt poured into our codebase and I could not fix all of them with the change. I solely focused on the SEO component. I removed the `translatedPath` and `noTranslation` properties entirely and provide a `overrideLanguages` instead to be able to override the automatic list of languages that we want to offer as alternatives. This is currently only relevant for the career page. And that's where this property will be coming from: I set this during the page creation by determining the "other translations". Right now only the UX/UI position suffers from the fact that there is no translations but it's a valid edge case.

+ Added new meta tags for: author, publish_date and copyright and I fill author and publish date for blog posts already.

+ I fixed some random linting errors I saw in github.

## How to review
+ Check the existence of the alternate tags with the correct urls (do not rely on the language switch)
+ Check the meta tags. 
+ Ensure our sharing still works (use metatags.io & such)

## Next
Let's work on the language switch because it's a technical debt drama when I see how many things we pass through. I think we should introduce a context and rely on the same logic I created for the SEO component. 

I still have to figure out why linkedin shows the wrong image from the body of a blog post. Maybe we need an additional meta tags as they don't properly read the open graph image tags?

### DIFF
--- a/gatsby/create-pages/create-career-pages.js
+++ b/gatsby/create-pages/create-career-pages.js
@@ -58,8 +58,9 @@ const createCareerPages = async ({ actions, graphql }) => {
   positions.forEach((position) => {
     const complementPosition = complementFinder.get(position);
     const path = getTranslationPath(position.slug, position.lang);
+    // create an array of available languages, otherwise empty
     const overrideLanguages = [complementPosition?.lang ?? null].filter(
-      (n) => n !== null,
+      (language) => language !== null,
     );
 
     createPage({
--- a/gatsby/create-pages/create-career-pages.js
+++ b/gatsby/create-pages/create-career-pages.js
@@ -58,6 +58,9 @@ const createCareerPages = async ({ actions, graphql }) => {
   positions.forEach((position) => {
     const complementPosition = complementFinder.get(position);
     const path = getTranslationPath(position.slug, position.lang);
+    const overrideLanguages = [complementPosition?.lang ?? null].filter(
+      (n) => n !== null,
+    );
 
     createPage({
       path: path,
@@ -66,6 +69,7 @@ const createCareerPages = async ({ actions, graphql }) => {
         id: position.id, // internal object id which is unique across all jobs
         language: position.lang,
         translation: complementPosition?.slug ?? null,
+        overrideLanguages: overrideLanguages,
       },
     });
   });
--- a/src/components/layout/seo.tsx
+++ b/src/components/layout/seo.tsx
@@ -3,31 +3,74 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import { useI18next } from 'gatsby-plugin-react-i18next';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
+import { I18nNextData } from '../../types';
 
 const DEFAULT_META_IMAGE_URL_PATH = '/sy-share-image.jpg';
-const LANGUAGES = ['en', 'de'];
 
 interface SeoProps {
   title: string;
+  author?: string;
+  publishDate?: string;
   description?: string;
   imageUrl?: string;
   siteType?: string;
   noIndex?: boolean;
-  noTranslation?: boolean;
-  translatedPath?: string;
+  overrideLanguages?: string[] | null;
   location: Location;
 }
 
-const SEO: React.FC<SeoProps> = ({
+/**
+ * Help Google & friends to show the most appropriate version by language,
+ * for more information: https://developers.google.com/search/docs/advanced/crawling/localized-versions#html
+ *
+ * You can force a specific set of languages by passing in `overrideLanguages`
+ * which is currently used to determine of a translation is available for a given job position detail
+ * which we can query from personio upfront.
+ *
+ * Otherwise i18n-next just concludes that there is a translation as it creates a translated page
+ * for every english page by default.
+ *
+ * The function calculates a list of meta tags from a given set of i18n-next data
+ * that is available for every page.
+ */
+const buildAlternateMetaTags = (
+  { languages, language, path, originalPath, defaultLanguage }: I18nNextData,
+  host,
+  overrideLanguages,
+) => {
+  const otherLanguages =
+    overrideLanguages ?? languages.filter((item) => item !== language);
+
+  const getUrl = (path, language) => {
+    if (language === defaultLanguage) {
+      return `${host}${originalPath}`;
+    }
+    return `${host}/${language}${originalPath}`;
+  };
+
+  const alternateLanguages = otherLanguages.map((language) => {
+    const url = getUrl(originalPath, language);
+    return { url, language };
+  });
+
+  return alternateLanguages.map(({ url, language }) => {
+    return (
+      <link rel="alternate" href={url} hrefLang={language} key={language} />
+    );
+  });
+};
+
+const SEO = ({
   description = '',
   title,
+  author,
+  publishDate,
   imageUrl,
   siteType,
   noIndex,
-  translatedPath,
-  noTranslation,
+  overrideLanguages,
   location,
-}) => {
+}: SeoProps) => {
   const { site } = useStaticQuery(
     graphql`
       query {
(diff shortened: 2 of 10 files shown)


gatsby/create-pages/create-career-pages.js
```diff
@@ -66,6 +69,7 @@ const createCareerPages = async ({ actions, graphql }) => {
         id: position.id, // internal object id which is unique across all jobs
         language: position.lang,
         translation: complementPosition?.slug ?? null,
+        overrideLanguages: overrideLanguages,
```
georgiee, line 73: Ideally this replaces the `translation` above which I keep until we can refactor the language switch.

gatsby/create-pages/create-career-pages.js
```diff
@@ -58,6 +58,9 @@ const createCareerPages = async ({ actions, graphql }) => {
   positions.forEach((position) => {
     const complementPosition = complementFinder.get(position);
     const path = getTranslationPath(position.slug, position.lang);
+    const overrideLanguages = [complementPosition?.lang ?? null].filter(
+      (n) => n !== null,
+    );
```
georgiee: I want a list of other languages or an empty array.

src/components/layout/seo.tsx
```diff
@@ -3,31 +3,74 @@ import React from 'react';
...
+    const url = getUrl(originalPath, language);
+    return { url, language };
+  });
+
+  return alternateLanguages.map(({ url, language }) => {
+    return (
+      <link rel="alternate" href={url} hrefLang={language} key={language} />
+    );
+  });
+};
```
georgiee: much easier to understand then the one I replaced I think 🙏

src/components/layout/seo.tsx
```diff
@@ -3,31 +3,74 @@ import React from 'react';
...
+) => {
+  const otherLanguages =
+    overrideLanguages ?? languages.filter((item) => item !== language);
+
+  const getUrl = (path, language) => {
+    if (language === defaultLanguage) {
+      return `${host}${originalPath}`;
+    }
+    return `${host}/${language}${originalPath}`;
+  };
```
georgiee: I decided to keep this function inline otherwise I have more input parameters than code 🤓

src/components/layout/seo.tsx
```diff
@@ -42,78 +85,59 @@ const SEO: React.FC<SeoProps> = ({
...
-    return LANGUAGES.map((lang) => {
-      const languagePath = lang === 'en' ? '' : `/${lang}`;
-      const pathNameWithSlashes =
-        language !== lang && translatedPath
-          ? prependAndAppendTrailingSlash(translatedPath)
-          : pathName;
-      const href = `${location.origin}${languagePath}${pathNameWithSlashes}`;
-      return <link rel="alternate" href={href} hrefLang={lang} key={lang} />;
-    });
-  };
```
georgiee, line 68: while this worked I have HUGE problems reading it and it's too much path processing which is not needed with the data from i18n next.

src/components/layout/seo.tsx
```diff
@@ -42,78 +85,59 @@ const SEO: React.FC<SeoProps> = ({
...
-      {metaImageUrl && <meta property="og:image" content={metaImageUrl} />}
+      <meta property="og:site_name" content={title} />
       <meta property="og:type" content={typeOfSite} />
+      <meta name="copyright" content="Satellytes" />
 
-      {/* -- Twitter Meta Tags -- */}
-      <meta name="twitter:card" content="summary_large_image" />
-      <meta name="twitter:title" content={title} />
-      <meta name="twitter:description" content={metaDescription} />
-      <meta name="twitter:creator" content={site.siteMetadata.author} />
```
georgiee, line 93: all replaced by og tags and this is a valid approach, see here https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started

src/components/layout/seo.tsx
```diff
@@ -42,78 +85,59 @@ const SEO: React.FC<SeoProps> = ({
...
-      {metaImageUrl && <meta name="twitter:image" content={metaImageUrl} />}
-      {metaImageUrl && <meta property="twitter:image:alt" content={title} />}
+      {author && <meta property="article:author" content={author} />}
+      {publishDate && <meta property="og:publish_date" content={publishDate} />}
```
georgiee, line 120: two new fields we want to have for blog posts.

src/components/pages/about-us/team.tsx
```diff
@@ -38,11 +38,10 @@ export const Team = ({ team }: TeamProps) => {
 
       <TeamLayout>
         {team.map((member) => {
+          const imageData = getImage(member.image);
```
georgiee, line 41: linting stufff

src/components/pages/blog/posts.tsx
```diff
@@ -25,15 +25,15 @@ export const Posts = ({ posts }: PostsProps) => {
   return (
     <BlogTeaserGrid>
       {posts.map((item) => {
-        const image = getImage(item.frontmatter.featuredImage)!;
+        const imageData = getImage(item.frontmatter.featuredImage);
```
georgiee, line 28: linting stufff

src/components/pages/landingpage/blog.tsx
```diff
@@ -33,7 +33,7 @@ export const Blog = ({ posts }: BlogProps) => {
 
       <TeaserGrid>
         {posts.map((item) => {
-          const image = getImage(item.frontmatter.featuredImage)!;
+          const imageData = getImage(item.frontmatter.featuredImage);
```
georgiee, line 36: linting stufff

src/templates/blog-post.tsx
```diff
@@ -37,11 +37,12 @@ const BlogArticleTemplate: React.FC<BlogArticleTemplateProps> = ({
        */}
       <SEO
         title={`${markdown.frontmatter.title} | Satellytes`}
+        author={markdown.frontmatter.author}
+        publishDate={markdown.frontmatter.date}
```
georgiee, line 41: welcome meta fields 💫

src/types.ts
```diff
@@ -84,3 +84,11 @@ export interface BreadcrumbEntry {
...
   label: string;
 }
+
+export interface I18nNextData {
+  languages: string[];
+  language: string;
+  path: string;
+  originalPath: string;
+  defaultLanguage: string;
+}
```
georgiee, line 94: I could not find any types for i18n next

georgiee (commented):
some comments on my PR changes

georgiee:
@satellytes/website-crew hey crew if this is an intimidating change (because of the complexity or the meta tags topic) I want to offer a review session in gather tomorrow to interested folks 🙏

gatsby/create-pages/create-career-pages.js
```diff
@@ -58,6 +58,9 @@ const createCareerPages = async ({ actions, graphql }) => {
   positions.forEach((position) => {
     const complementPosition = complementFinder.get(position);
     const path = getTranslationPath(position.slug, position.lang);
+    const overrideLanguages = [complementPosition?.lang ?? null].filter(
+      (n) => n !== null,
```
feedm3: What about `language` instead of `n`?

src/components/layout/seo.tsx
```diff
@@ -3,31 +3,74 @@ import React from 'react';
...
+ * which we can query from personio upfront.
+ *
+ * Otherwise i18n-next just concludes that there is a translation as it creates a translated page
+ * for every english page by default.
+ *
+ * The function calculates a list of meta tags from a given set of i18n-next data
+ * that is available for every page.
+ */
+const buildAlternateMetaTags = (
+  { languages, language, path, originalPath, defaultLanguage }: I18nNextData,
```
feedm3: `path` is defined but never used

src/components/layout/seo.tsx
feedm3: Yes it is 🙏

feedm3 (approved):
Check the SEO tags on the deployed page, seems to work. Well done 👍 

Not related to this PR: The homepage preview image was not created and only returns a `500` (blog posts work). Checked the logs but couldn't find any suspicious stuff. 
https://satellytescommain-gkmetatags.gtsb.io/sy-share-image.jpg

gatsby/create-pages/create-career-pages.js
georgiee: 👍

PR #292, opened 2021-12-17
Some text changes

abdolg:
Added and changed some content. Only in german because we will translate the german content all in once

PR #298, opened 2021-12-20
chore: remove illegal unicode character from translation file

fischeversenker:
We introduced the `\u2028` character in our german translation file. This completely and silently broke storybook.

This PR removes that character.

Since you [introduced this character](https://github.com/satellytes/satellytes.com/commit/6a30fb3c4052bcaa27945cff980ccc3cadaf4ce1), @abdolg, and you did not yet react to [my question in your most recent PR](https://github.com/satellytes/satellytes.com/pull/292#discussion_r771388091) about what this character does I'm assuming we can simply remove it.


hofmann-felix (approved):
Storybook works again 👍 Thank you for debugging it

georgiee (approved):
oh wow. I would not even now how to create such a character and insert it 😅
storybook works again. thanks for the fix.

PR #299, opened 2021-12-20
feat: add TextArea component

hofmann-felix:
Storybook URL: https://deploy-preview-299--satellytes-website-storybook.netlify.app/?path=/docs/components-forms-textinput--as-input&globals=locale:en;locales.en:English;locales.de:Deutsch

## What's included?
- A TextArea component, which can be used later in the Forms, was added to `textinput.tsx`
- With the `type` prop you can decide whether the TextInput is a normal `input` or a `textarea`. I decided to put both components in one file because they use nearly the same logic and style. 
- The API of `TextInput` is based on the legacy input component and should work with `react-hook-form`

## Preview
textarea: 
Bildschirmfoto 2021-12-20 um 13 25 55

Normal input field (was already implemented):
Bildschirmfoto 2021-12-20 um 13 25 01


src/components/forms/textinput/textinput.stories.mdx
```diff
@@ -5,14 +5,48 @@ import { TextInput } from './textinput';
...
+      errorMessage=""
+      required={true}
+      type='textarea'
+    />
+    <TextInput
+      label="Message"
+      name="message"
+      errorMessage="With Error"
+      required={true}
+      type='textarea'
```
georgiee, line 47: The textfield somehow is too wide? I tried with a lot of text but realized it also happens when empty. Can you check that?
Screen Shot 2021-12-20 at 16 53 51

src/components/forms/textinput/textinput.stories.mdx
```diff
@@ -5,14 +5,48 @@ import { TextInput } from './textinput';
...
+      errorMessage="With Error"
+      required={true}
+    />
   </Story>
 </Canvas>
 
-<ArgsTable of={TextInput}/>
+## As Textarea
+<Canvas withSource='open'>
+  <Story name='Text Area'>
```
georgiee: `As Input` is the other story. Please choose the same naming concept.

src/components/forms/textinput/textinput.tsx
```diff
@@ -1,53 +1,115 @@
...
+  resize: vertical;
+
+  ${TextStyles.textR}
+
+  background: ${theme.palette.background.leadbox};
+  border: 2px solid
+    ${(props) =>
+      props.hasError
+        ? theme.palette.text.errorMessage
+        : theme.palette.background.leadbox};
```
georgiee: I don't know how others think about that, but I would rather output the default 
and only override the special case instead of creating a condition between both.

Little bit easier to grasp maybe? Something like that?

```
  border: 2px solid ${theme.palette.background.leadbox}
  ${({hasError) => hasError && css`border: 2px solid ${theme.palette.text.errorMessage }` 
```

src/components/forms/textinput/textinput.tsx
```diff
@@ -1,53 +1,115 @@
...
-  font-weight: 700;
-  color: #ff0d35;
+  color: ${theme.palette.text.errorMessage};
 `;
 
-interface TextInputProps {
-  errorMessage: string;
-  label: string;
+export interface TextInputProps {
+  inputRef: any;
```
georgiee, line 59: That's the legacy part among others so you can replace it within the current forms? I think that approach is okay but I would have done it much different. Let the legacy form stay as is and create a fresh table of components and forms. Much more fun to work within that environment and you can still  cherry pick whatever you want to reuse.

What's your plan? Slowly refactoring what we have into a new shape?

georgiee (changes requested):
I have never seen input & textarea components combined like that. I fear this will hunt us pretty soon, but I want to acknowledge your efforts here.

I also have noted some blocker with the design (by looking at the storybook only) and I have the suggestion that you don't update the legacy forms and the legacy components but instead start with an empty set of forms and components. That way you can focus on the wiring of the react form hooks and components (to get comfortable with it), create a few components without caring for the desigh yet and once you are satisfied you can do the final polish and deliver two shiny new forms for career and contact.

What's your plan?

src/components/forms/textinput/textinput.stories.mdx
hofmann-felix: This is a pure storybook problem, which we also have with the header. On the actual website, the problem will not exist.

src/components/forms/textinput/textinput.tsx
hofmann-felix: Yes, my original plan was to refactor the existing components step by step. Therefore `input` and `textarea` are combined in one component (like in the legacy project). I like your suggestion to develop the new form independently from the existing implementation.

PR #300, opened 2021-12-20
chore: add missing Translation on service page

milljoniaer:
I found a german translation in the english translation file and changed it


georgiee (approved):
thanks.

PR #301, opened 2021-12-20
chore: add image of Felix Hamann

fischeversenker:
This adds me to the "Teams" section in the "About us" page 🎉


georgiee (approved):
Oh good point with the last names. Should we include them or not? This also reminds me of creating a PR for me too 😅

fischeversenker:
I like the personal touch that we have with using only first names... unfortunately with two Felixes this is not possible anymore. For all the other members I suggest to keep the first-name-only strategy 🙌

PR #304, opened 2021-12-20
chore: change leadbox text on the career page

milljoniaer:
My language skills come to an end, but I stumbled across this passage:
image

"Dein Job ist nicht dabei?" sounds better, doesn't it?

I was also unsure with the english translation.

Please have a look at the phrases. I am looking forward to hear your thoughts on this.


fischeversenker (approved):
Thanks! I like it a lot more than the previous versions 🙌 🙏

georgiee (approved):
Thanks for finding and thinking about this 🙏 Sounds much better!

PR #307, opened 2021-12-22
feat: better blog sharing

georgiee:
Preview: https://satellytescommain21751-gkfixblogsharing.gtsb.io/

+ Create the sharing cards with `createFileNodeFromBuffer` instead of writing into public 
+ Updated our `buildGatsbyCloudPreviewUrl` as it used the wrong prefix
+ Create share images for blog from the teaser instead of using the card
+ Node set to v16 *engine and gatsby env

Blog with Image Preview:
Screen Shot 2021-12-22 at 10 21 39

Cards are still working:
Screen Shot 2021-12-22 at 10 21 46

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@master
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: '14'
+          node-version: '16'
       - name: Install
         run: yarn
       - name: Unit tests
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -3,7 +3,7 @@ const {
 } = require('./gatsby/util/build-gatsby-cloud-preview-url');
 const siteMapTransformers = require('./gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap');
 
-const GATSBY_SITE_PREFIX = process.env.GATSBY_SITE_PREFIX ?? '';
+const GATSBY_SITE_PREFIX = process.env.GATSBY_SITE_PREFIX || '';
 const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
   prefix: GATSBY_SITE_PREFIX,
   branch: process.env.BRANCH,
@@ -12,8 +12,6 @@ const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
 // either use a branch preview url if any
 const BASE_URL =
   BRANCH_PREVIEW_URL || process.env.GATBSY_BASE_UR || 'http://localhost:8000';
-
-console.info('BASE_URL', BASE_URL);
 
 const LANGUAGES = ['en', 'de'];
 const DEFAULT_LANGUAGE = 'en';
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -9,6 +9,7 @@ const BASE_URL = buildGatsbyCloudPreviewUrl({
 });
 
 console.info(`Using BASE_URL ${BASE_URL}`);
+console.info(`env`, process.env);
 const LANGUAGES = ['en', 'de'];
 const DEFAULT_LANGUAGE = 'en';
 
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -4,10 +4,10 @@ const {
 const siteMapTransformers = require('./gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap');
 
 const GATSBY_SITE_PREFIX = process.env.GATSBY_SITE_PREFIX ?? '';
-const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl(
-  GATSBY_SITE_PREFIX,
-  process.env.BRANCH,
-);
+const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
+  prefix: GATSBY_SITE_PREFIX,
+  branch: process.env.BRANCH,
+});
 const BASE_URL =
   process.env.GATBSY_BASE_URL || BRANCH_PREVIEW_URL || 'http://localhost:8000';
 
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -8,8 +8,10 @@ const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
   prefix: GATSBY_SITE_PREFIX,
   branch: process.env.BRANCH,
 });
+
+// either use a branch preview url if any
 const BASE_URL =
-  process.env.GATBSY_BASE_URL || BRANCH_PREVIEW_URL || 'http://localhost:8000';
+  BRANCH_PREVIEW_URL || process.env.GATBSY_BASE_UR || 'http://localhost:8000';
 
 console.info({ GATSBY_SITE_PREFIX, BRANCH_PREVIEW_URL, BASE_URL });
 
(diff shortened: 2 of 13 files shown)


georgiee:
back in game, everything works, including the breaking incremental builds.
This was because of the gatsby foreign key issue. We relied on the deprecated way (see https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v3-to-v4/#___node-convention-is-deprecated) where an postfix of `___NODE` signals the foreign key relationship. This causes problems in gatsby 4 it seems. After I switched to the modern way through `createSchemaCustomization` (see https://www.gatsbyjs.com/docs/how-to/plugins-and-themes/creating-a-source-plugin/#create-foreign-key-relationships-between-data) it works flawless 🙌

We now have correct cards and the blog shows the correct teaser image instead of the card (we still create the card so we don't have to filter the blog md documents, okay for now).

I used two ways to define the new schema. The syntactic sugar from gatsby (or is it graphql?) with `@link`:
```
socialCard: File @link(from: "fields.socialCard")
```

and the more explicit way of using an resolver:

```
      fields: {
        socialCard: {
          type: 'File',
          resolve: (source, args, context, info) => {
            return context.nodeModel.getNodeById({
              id: source.fields.socialCard,
              type: 'File',
            });
          },
        },
      },
```

both have the same effect that they create a field `socialCard` in the root of the node with the actual node that is referenced through `fields.socialCard`. I wonder if we could put the actual card creation in the resolve or if it's better to leave it separated as is🧐

There the proof it's working now 🙌

Screen Shot 2021-12-22 at 15 05 10

Screen Shot 2021-12-22 at 15 05 02

Screen Shot 2021-12-22 at 15 03 12

.github/workflows/continous-integration.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@master
       - uses: actions/setup-node@v2-beta
         with:
-          node-version: '14'
+          node-version: '16'
```
georgiee, line 20: While debugging, but I think this not wrong at all to bump for our project.

gatsby-config.js
```diff
@@ -3,10 +3,16 @@ const {
...
+const GATSBY_SITE_PREFIX = process.env.GATSBY_SITE_PREFIX || '';
+const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
+  prefix: GATSBY_SITE_PREFIX,
+  branch: process.env.BRANCH,
 });
+
+// either use a branch preview url if any
+const BASE_URL =
+  BRANCH_PREVIEW_URL || process.env.GATBSY_BASE_UR || 'http://localhost:8000';
+
```
georgiee, line 15: I untangled the fallback to focus on the branch url creation. I think that's easier to follow now

gatsby-node.js
```diff
@@ -27,6 +27,58 @@ exports.onCreateNode = (gatsbyCreateNodeArgs) => {
...
+          },
+        },
+      },
+      interfaces: ['Node'],
+    }),
+  ];
+
+  createTypes(typeDefs);
+};
+
```
georgiee, line 81: yeah that solved my hour long trouble getting the file properly connected to the career and blog posts. It always worked on a fresh build but incremental builds failed. Probably because the incremental build hat some issues restoring through the implicit fk `___NODE` assignment we used before. much better because it's explicit now.

gatsby-node.js
```diff
@@ -27,6 +27,58 @@ exports.onCreateNode = (gatsbyCreateNodeArgs) => {
...
+            return context.nodeModel.getNodeById({
+              id: source.fields.socialCard,
+              type: 'File',
+            });
+          },
+        },
+      },
+      interfaces: ['Node'],
+    }),
+  ];
```
georgiee, line 77: This could be the same one liner as above for `MarkdownRemark`, but I wanted to keep it for now to demonstrate the alternative because I wonder if I should resolve and generated the card in here. Fine to keep it around for now or better put it into an issue?

gatsby/create-node/create-preview-cards.js
```diff
@@ -1,65 +1,44 @@
...
-  const publicUrl = `${siteMetadata.siteUrl}/${fileName}`;
-
-  generateCard({ title: post.title }, outputFile).then(() => {
-    createNodeField({
-      node,
-      name: `socialCard`,
-      value: publicUrl,
-    });
+  const buffer = await generateCardToBuffer({ title });
+  const fileNode = await createFileNodeFromBuffer({
```
georgiee, line 18: finally I had time to create the file from buffer to create a proper file node instead of working around gatsby and output the file directly in static. That approach was never wrong but the 500 errors (which were caused by plain wrong urls we generated) brought me on this so. I like this much more. It's more complicated but less implicit.

gatsby/create-node/create-preview-cards.js
```diff
@@ -1,65 +1,44 @@
...
+    name: 'social-card',
```
georgiee, line 19: extension is inferred from the buffer where I set jpg.

gatsby/create-node/create-preview-cards.js
```diff
@@ -1,65 +1,44 @@
...
-const createPreviewCards = ({ node, _, actions }) => {
+const createPreviewCards = ({ node, ...rest }) => {
   if (node.internal.type === 'SyPersonioJob') {
-    createPreviewJob({ node, _, actions });
+    return createPreviewCard(node.name, { node, ...rest });
   }
 
   if (node.internal.type === 'MarkdownRemark') {
-    createPreviewMarkdown({ node, _, actions });
+    return createPreviewCard(node.frontmatter.title, { node, ...rest });
```
georgiee, line 41: I only need to pass in the actual string to generate. Before it was two methods while it's actually a single one 👍

src/components/layout/seo.tsx
```diff
@@ -89,9 +89,9 @@ const SEO = ({
 
   const metaDescription = description || t('main.description');
   const typeOfSite = siteType || 'website';
-  const defaultImageUrl =
-    site.siteMetadata.siteUrl + DEFAULT_META_IMAGE_URL_PATH;
-  const metaImageUrl = imageUrl ?? defaultImageUrl;
+
+  const metaImageUrl =
```
georgiee, line 93: every image we pass in need to be path. never an url as seo should resolve that. which we do now.

src/templates/blog-post.tsx
```diff
@@ -20,6 +20,9 @@ const BlogArticleTemplate: React.FC<BlogArticleTemplateProps> = ({
   const { t } = useTranslation();
   const markdown = data.markdownRemark;
 
+  const shareImagePath =
+    markdown.frontmatter.shareImage.childImageSharp.fixed.src;
```
georgiee, line 24: we switch from card to the image for blog posts here. The shareImage is just another variant of the teaser cropped to the expected format.

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "git@github.com:satellytes/satellytes.com.git"
   },
   "engines": {
-    "node": ">= 14.0.0",
+    "node": ">= 16.0.0",
```
georgiee, line 12: this will cause yarn to complain when a local node version is lower. let's see who will complain 😁

georgiee (commented):
some comments

georgiee:
damn. this is nuts. is this gatsby 4 killing us? 
Screen Shot 2021-12-22 at 15 33 30

georgiee:
back to draft. this might be best handled in isolation while I can bring the parts that work through #308

georgiee:
closed in favor of #309

PR #309, opened 2021-12-23
refactor: create share card from buffer

georgiee:
I have created a Gatsby PR to fix this situation for the future and other lost souls:
https://github.com/gatsbyjs/gatsby/pull/34313

---
This will change the social cards generation. Instead of writing directly into `public/` which happily works we move closer to the gatsby way of doing things by creating an actual file node. We can do this with `createFileNodeFromBuffer` from [gatsby-source-filesystem](https://www.gatsbyjs.com/plugins/gatsby-source-filesystem/). 

In order to save that node to our original node we can't just save the ID of the file in some fields. We need to tell gatsby about the foreign key relationship. This was usually done with the `myField___NODE` notation which is now deprecated.

That's why we have to create a custom graphql schema addition which is also included in there.

Preview:
https://satellytescommain21751-gkcardsbuffer.gtsb.io/

## Still works
We have swapped from file to buffer and the files are still work. You can see the switch after looking the the generated path `https://satellytescommain21751-gkcardsbuffer.gtsb.io/static/93ec6b3d62c7099eaea57bcea3aad6a6/468fb/social-card.jpg` 👍

via metatags.io
Screen Shot 2021-12-23 at 11 23 06

Screen Shot 2021-12-23 at 11 22 40

## Insights
Another thing that we struggled a lot with: The created files are fine, but only on the initial build. Any incremental build (repeated starts of gatsby) throw away the files and the value were null then. We tried so many things because we didn't know why this was happening while using standard gatsby functions.

In the end the solution is a single line: `parentNodeId: node.id` which we pass in addition to `createFileNodeFromBuffer`. This solves the problem because now gatsby leaves the file as is. It seems they have been deleted before because gatsby though they are dangling files without an owner.

The problematic thing here is `createRemoteFileNode` requires the parent id while the `createFileNodeFromBuffer` happily set null as a default. I think that's a bug worth to submit with this feedback.

Here the referenced source for createFileNodeFromBuffer
https://github.com/gatsbyjs/gatsby/blob/b9f7749f36ace526bf0ba05b80731091cc0f2ceb/packages/gatsby-source-filesystem/src/create-file-node-from-buffer.js#L125

vs createRemoteFileNode
https://github.com/gatsbyjs/gatsby/blob/b9f7749f36ace526bf0ba05b80731091cc0f2ceb/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L100

I also found this issue talking about a similar thing:
https://github.com/gatsbyjs/gatsby/issues/4537

referencing this PR but it's not really explaining thinks even though Kyle asked for that:
https://github.com/gatsbyjs/gatsby/pull/6793/files#diff-b746772e1ad7ebd3e0f347c527f10886ffd1f7e74eb29ee957f1d7946241781aR108-R121

The actual foreign key documentation can be found here:
https://www.gatsbyjs.com/docs/node-creation/

### DIFF
--- a/gatsby/create-node/create-preview-cards.js
+++ b/gatsby/create-node/create-preview-cards.js
@@ -20,7 +20,14 @@ const createPreviewCard = async (
   }
 
   const buffer = await generateCardToBuffer({ title });
-
+  /**
+   * The util function `createFileNodeFromBuffer` from the official gatsby source plugin `gatsby-source-filesystem`
+   * creates a file node from a given file buffer. The value of `parentNodeId` creates the necessary relationship
+   * between the original node and the actual file node so it's not garbage collected.
+   *
+   * The actual foreign key relationship is resolved through `createSchemaCustomization`
+   * in gatsby-node.js for all node types the `createPreviewCard` is invoked for.
+   */
   const fileNode = await createFileNodeFromBuffer({
     name: 'social-card',
     buffer,
@@ -33,7 +40,7 @@ const createPreviewCard = async (
   if (fileNode) {
     createNodeField({
       node,
-      name: `socialCard`,
+      name: `socialCardFileId`,
       value: fileNode.id,
     });
   }
--- a/gatsby/create-node/create-preview-cards.js
+++ b/gatsby/create-node/create-preview-cards.js
@@ -1,65 +1,51 @@
-const slugify = require('slugify');
-const path = require('path');
-const { siteMetadata } = require('../../gatsby-config');
 const {
-  generateCard,
+  generateCardToBuffer,
 } = require('../util/preview-card-generator/generate-card');
+
+const { createFileNodeFromBuffer } = require('gatsby-source-filesystem');
 
 /**
  * Create a preview card file and put the url to the Gatsby store to be able
  * to link it in the page header.
  */
-const createPreviewMarkdown = ({ node, _, actions }) => {
-  const { createNodeField } = actions;
-  const post = node.frontmatter;
+const createPreviewCard = async (
+  title,
+  { node, _, actions, getCache, createNodeId },
+) => {
+  const { createNode, createNodeField } = actions;
 
   // we need the title in order to generate anything
-  if (!post.title) {
+  if (!title) {
     return;
   }
 
-  // we create a file name from the path (which is the page slug and more unique) but if none is available take the title
-  const fileName = `social-card---${slugify(post.path || post.title, {
-    lower: true,
-  })}.jpg`;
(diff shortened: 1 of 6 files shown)


gatsby/create-node/create-preview-cards.js
```diff
@@ -1,65 +1,51 @@
...
-  const outputFile = path.join('public', fileName);
-  const imagePath = `/${fileName}`;
-
-  generateCard({ title: node.name }, outputFile).then(() => {
+  if (fileNode) {
     createNodeField({
       node,
-      name: 'socialCard',
-      value: imagePath,
+      name: `socialCard`,
```
feedm3: I think the name `socialCardId` would be a better fit, as this field represents the `id` and not the whole object.

gatsby/create-node/create-preview-cards.js
```diff
@@ -1,65 +1,51 @@
...
-  const imagePath = `/${fileName}`;
+  const buffer = await generateCardToBuffer({ title });
 
-  generateCard({ title: post.title }, outputFile).then(() => {
-    createNodeField({
-      node,
-      name: `socialCard`,
-      value: imagePath,
-    });
+  const fileNode = await createFileNodeFromBuffer({
```
feedm3, line 31: This function creates AND links the node to the parent, right? A comment above would make this more clear, as I was just wondering at which place the actual connection between the Markdown/SyJob and the file is happening.

feedm3 (approved):
Well done 👍  I just have 2 small suggestions, will approve right away to not unblock this PR from merging.

gatsby/create-node/create-preview-cards.js
georgiee: yeah I bounced that idea forth and back and now you convinced me ✨

gatsby/create-node/create-preview-cards.js
georgiee: I'll  add the following comment to make things clear:

> The util function `createFileNodeFromBuffer` from the official gatsby source plugin `gatsby-source-filesystem`
> creates a file node from a given file buffer. The value of `parentNodeId` creates the necessary relationship
> between the original node and the actual file node so it's not garbage collected.
>  
> The actual foreign key relationship is resolved through `createSchemaCustomization`
> in gatsby-node.js for all node types the `createPreviewCard` is invoked for.


Ideally  we create a plugin "create social cards" or such that would bundle this invocation (tied to `onCreateNodde`) and the fk lookup which is tied to `createSchemaCustomization`. Feels like a random split while it's technically required. That's our next small local plugin we can extract (and configure) 🙏

PR #310, opened 2021-12-23
feat: update contact form and text field components

hofmann-felix:
Preview URL: https://satellytescommain21751-feattextarea98960.gtsb.io/contact/
Storybook: https://deploy-preview-310--satellytes-website-storybook.netlify.app/?path=/docs/components-forms-textinput--regular&globals=locale:en;locales.en:English;locales.de:Deutsch

# What's included?
This PR includes an updated version of the `TextInput` and `TextArea` components. To show the changes the contact form was also updated. The [controlled approach of react-hook-form](https://react-hook-form.com/api/usecontroller) is used:
### TextInput and TextArea
- Both components use the `useController` hook to get all the necessary information. This includes for example the fieldstate with attributes like required and error messages. 
- In the legacy version, `input` and `textarea` were a single component. Now they are split to make the difference in the parent more visible
### Contact Form
- The contact form uses the new components. That's why the logic is adjusted a little bit. The main difference is the `controller` prop which is given to all text fields.
- When the user successfully submits a form a success message is shown 

# Preview
_Contact Form:_
Bildschirmfoto 2021-12-23 um 12 05 18
_Success Message:_
Bildschirmfoto 2021-12-23 um 12 06 16


src/assets/locales/de/translations.json
```diff
@@ -167,10 +167,13 @@
...
   "contact.error.email": "Deine E-Mail fehlt",
   "contact.error.email-undefined": "Irgendwas stimmt an dieser E-Mail nicht",
-  "contact.action.again-text": "Leider gab es einen Fehler. Bitte versuche es noch einmal. Klappt das nicht, schicke deine Nachricht bitte direkt an <2>beep@satellytes.com</2>",
+  "contact.action.again-text": "Leider gab es einen Fehler. Bitte versuche es noch einmal. Klappt das nicht, schicke deine Nachricht bitte direkt an",
   "contact.action.missing": "Bitte fülle alle benötigten Felder aus",
   "contact.action.send": "Senden",
-  "contact.action.sent": "Gesendet",
+  "contact.action.sent.headline": "E-Mail wurde gesendet",
+  "contact.action.sent.kicker": "Vielen Dank!",
+  "contact.action.sent.info": "Wir schauen uns alles an und melden uns blablabla. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Etiam porta sem malesuada magna mollis euismod. Fusce dapibus, tellus ac cursus commodo, tortor mauris condimentum nibh, ut fermentum massa justo sit amet risus. Morbi leo risus, porta ac consectetur ac, vestibulum at eros. Cras justo odio, dapibus ac facilisis in, egestas eget quam.",
```
hofmann-felix: Here is currently still the Lorem Ipsum text from Figma. It would be nice if someone has a suggestion for this text.

georgiee: Oh I see, we only had the success button before right? I would prefer never check in text like that. I recognize that this comes straight from the design and that they should never do that. Either `lorem ipsum` or a real text. Everything else is not good. Next time you see this you can happil replace it with lorem entirely.

Here my suggestion:

```suggestion
  "contact.action.sent.info": "Vielen Dank für Ihre Nachricht. Wir bearbeiten diese und melden uns schnellstmöglich bei Ihnen.",
```

src/assets/locales/de/translations.json
```diff
@@ -157,7 +157,7 @@
   "services.leadbox.mail": "info@satellytes.com",
   "contact.title": "Kontakt",
   "contact.address": "Adresse",
-  "contact.info-link": "<0>Nutzen Sie unser Kontaktformular oder schreiben Sie uns eine E-Mail an <1>beep@satellytes.com</1></0>",
+  "contact.info-link": "<0>Nutzen Sie unser Kontaktformular oder schreiben Sie uns eine E-Mail an <1>info@satellytes.com</1></0>",
```
georgiee, line 161: 👍

src/components/forms/text-field/text-area.stories.mdx
```diff
@@ -0,0 +1,24 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { TextArea } from './text-area';
+import { getSampleController } from './text-field-story-helper';
+
+<Meta
+  title="Components/Forms/TextArea"
+  component={{ TextArea }} />
+
+# TextArea
+The component is used to get input from the user. Besides the `label` it receives props of the type `UseControllerProps`. The API of this component is based on the [controlled approach of react-hook-form](https://react-hook-form.com/api/usecontroller).
```
georgiee: Thanks for taking care of the description. That will help people in the future for sure.

src/components/forms/text-field/text-area.stories.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+<Canvas withSource='open'>
+  <Story name='Regular'>
+    <TextArea
+      label="Message"
+      name="message"
+      control={getSampleController()}
+      rules={{ required: "Your message is missing" }}
+      style={{width: 'calc(100% - 30px)'}}
```
georgiee: that's a quick fix for storybook for the story only ?

georgiee: I like the API. How is your feeling?

src/components/forms/text-field/text-area.tsx
```diff
@@ -0,0 +1,50 @@
...
+  background: ${theme.palette.background.leadbox};
+  border: 2px solid ${theme.palette.background.leadbox};
+  ${(props) =>
+    props.hasError &&
+    css`
+      border: 2px solid ${theme.palette.text.errorMessage};
+    `};
+`;
+
+export const TextArea = (
```
georgiee: Feels much better to have a distinct component and not to mix them. Thank you

src/components/forms/text-field/text-area.tsx
```diff
@@ -0,0 +1,50 @@
...
+      )}
+      <StyledTextArea
+        aria-required={true}
+        hasError={Boolean(fieldState?.error?.message)}
+        {...field}
+        {...props}
+      />
+
+      {fieldState?.error?.message && (
+        <StyledErrorMessage>{fieldState?.error?.message}</StyledErrorMessage>
```
georgiee: you have already checked that those values are present in the boolean check so make this explicit without the optional chaining 💪

src/components/forms/text-field/text-area.tsx
```diff
@@ -0,0 +1,50 @@
...
+  props: UseControllerProps<any> & { label: string },
+) => {
+  const { field, fieldState } = useController(props);
+
+  return (
+    <>
+      {props.label && (
+        <Label htmlFor={props.name}>
+          {props.label}{' '}
+          {props?.rules?.required && <span aria-hidden={true}>*</span>}
```
georgiee: That's looks nice

src/components/forms/text-field/text-area.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const TextArea = (
+  props: UseControllerProps<any> & { label: string },
+) => {
+  const { field, fieldState } = useController(props);
+
+  return (
+    <>
+      {props.label && (
+        <Label htmlFor={props.name}>
+          {props.label}{' '}
```
georgiee: What's happening here?

src/components/forms/text-field/text-area.stories.mdx
georgiee: > The component is used to get input from the user.

but it's a special format being an text area isn't it? But good enough for now, I just say that the two components have the same text.

src/components/forms/text-field/text-input.tsx
```diff
@@ -0,0 +1,68 @@
...
+  display: block;
+  margin-top: 4px;
+
+  color: ${theme.palette.text.errorMessage};
+`;
+
+export const TextInput = (
+  props: UseControllerProps<any> & { label: string },
+) => {
+  const { field, fieldState } = useController(props);
```
georgiee: that controller makes things so easy regarding error handling and such.

georgiee: You don't use the `formState` but it contains `formState` and I think this could help disabling the input fields while the submission is running. It's not in the design but we know better that this is a good addition, isn't it ?

src/components/pages/contact/contact-form.tsx
```diff
@@ -56,101 +73,62 @@ export const ContactForm: React.FC = () => {
...
-        </GridItem>
-      </Grid>
-    </form>
+    <>
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <StyledGrid nested>
+          <GridItem xs={12} md={6}>
+            <TextInput
+              name={'name'}
+              label={t('contact.name')}
```
georgiee: Did you create the contact form again or did you adapt the legacy one? I thought you build from scratch (for fun) ? This looks like a legacy adaption. I would like to get rid of the grid and we should rather define a local set of divs with the appropriate css grid. We don't need a shared grid component. This looks super noisy and I think there is a lot of refactor potential. Maybe you can just leave the contact form as is and instead create a second contact form we can improve slowly (via the storybook) and later just replace it on the contact page.

It's not wrong what you are doing here. If you feel comfortable we can move forward with that approach too.

src/components/pages/contact/contact-form.tsx
```diff
@@ -56,101 +73,62 @@ export const ContactForm: React.FC = () => {
...
+              control={control}
+              rules={{ required: t<string>('contact.error.name') }}
```
georgiee: Why did you type `t` here ? We never did before?

src/components/pages/contact/contact-form.tsx
```diff
@@ -56,101 +73,62 @@ export const ContactForm: React.FC = () => {
...
+              label={t('contact.email')}
+              control={control}
+              rules={{
+                required: t<string>('contact.error.email'),
+                pattern: {
+                  value: SIMPLE_EMAIL_PATTERN,
+                  message: t<string>('contact.error.email-undefined'),
+                },
+              }}
+            />
```
georgiee: I think you could create local components at the top of the file like `Email`, `FirstName` and later the form would look super tidy and easy to navigate:

````
<form>
  <Row>
     <FirstName control={control}/>
     <LastName control={control}/>
   </Row>
  <Row>
     <Email control={control}/>
     ...
   </Row>
</form>
```

georgiee (changes requested):
hey thanks,
that goes absolutely in the right direction 👍
Was it a clear approach for you? Looks good and I can see us quickly refactoring our two forms with that and getting rid of the technical debt 💪

I have some questions regarding the next steps. That's why I would like to block it for now.


---

I can't send any messages as the url from netlify is missing of course. I think you can add them manually, read https://docs.netlify.com/forms/setup/.

Ideally we use another service. Fabian had some ideas with Gatsby Lambda to send the mails ([here some example](https://dev.to/char_bone/using-netlify-lambda-functions-to-send-emails-from-a-gatsbyjs-site-3pnb)). 

And here some idea to think out of the box: Currently netlify sends a mail and a slack notification. Actually, we are good for the beginning with a Slack message. Maybe we can focus on creating a gatsby lambda function to make that happen? It's a nice excursion for you I bet. Let's talk about that in Slack 🙌

src/components/forms/text-field/text-area.stories.mdx
hofmann-felix: Yes, it's a storybook-only fix. We have the same problem with the header component

src/components/forms/text-field/text-area.tsx
hofmann-felix: I wanted to make the label clickable to put the field in focus, but I forgot to set the id for the input element. Now it works

src/components/forms/text-field/text-area.stories.mdx
hofmann-felix: Yes I think it is easier to understand than in the legacy version 👍

src/components/forms/text-field/text-input.tsx
hofmann-felix: Nice idea 👌

src/components/pages/contact/contact-form.tsx
hofmann-felix: I build everything from scratch except the grid and the handleSubmit method. But I agree with you, it would be nice if we get rid of the gird

src/components/pages/contact/contact-form.tsx
hofmann-felix: Good suggestion, thank you 🙏

src/components/forms/text-field/text-area.stories.mdx
georgiee: 👍

src/components/forms/text-field/text-input.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import { TextStyles } from '../../typography';
+import { theme } from '../../layout/theme';
+import { useController, UseControllerProps } from 'react-hook-form';
+
+const Input = styled.input<{ hasError?: boolean }>`
```
georgiee: that file could should placed in a separate folder `text-input`, shouldn't it ?

src/components/forms/text-field/text-input.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {...field}
+        {...props}
+      />
+
+      {fieldState?.error?.message && (
+        <StyledErrorMessage>{fieldState.error.message}</StyledErrorMessage>
+      )}
+    </>
+  );
+};
```
georgiee: I think you could extract the message as a local variable, then the code would be easier to read as you don't have so long chains in there.

```suggestion
export const TextInput = (
  props: UseControllerProps<any> & { label: string },
) => {
  const { field, fieldState, formState } = useController(props);
  const errorMessage = fieldState?.error?.message;
  
  return (
    <>
      {props.label && (
        <Label htmlFor={props.name}>
          {props.label}{' '}
          {props?.rules?.required && <span aria-hidden={true}>*</span>}
        </Label>
      )}
      <Input
        type={'text'}
        aria-required={true}
        hasError={Boolean(errorMessage)}
        disabled={formState.isSubmitting}
        id={props.name}
        {...field}
        {...props}
      />

      {errorMessage && (
        <StyledErrorMessage>{errorMessage}</StyledErrorMessage>
      )}
    </>
  );
};
```

src/components/forms/text-field/text-input.tsx
georgiee: I have seen the usage, thanks🙌

src/components/pages/contact/contact-form.tsx
```diff
@@ -1,148 +1,165 @@
...
-            name="message"
-            label={t('contact.message')}
-            type={'text-area'}
-          />
-        </GridItem>
-
-        <HoneypotField label="First Name" control={register('firstName')} />
-        <HoneypotField label="Phone" control={register('phone')} />
-
+    <form name="xontact" onSubmit={handleSubmit(onSubmit)}>
```
georgiee: something is off here with `xontact`

src/components/pages/contact/contact-form.tsx
```diff
@@ -1,148 +1,165 @@
...
-type RequestStatus = 'pending' | 'submitting' | 'success' | 'error';
 const API_ENDPOINT = '/api/contact-form';
 
-const StyledCaptionText = styled(CaptionText)`
-  color: inherit;
+const Grid = styled.div`
+  margin-bottom: 24px;
+  gap: 24px;
+  display: grid;
+  grid-template-columns: repeat(12, 1fr);
```
georgiee: I see your shortcut here but please don't use a 12 column based grid for that simple case. I have imagined a markup like that.


```
<FormLayout>
  <Name/>
  <Email/>
  <Message/>
</FormLayout>

<HoneyPot/>
<MandatoryNotes/>
<Submit/>

<Errors/>

```

where `FormLayout` defines the two column grid and `Message` stretch itself across the two columns with grid-column set to -1/1. This removes a lot of noise in the markup.

I love to plan and design my code with this kind of API or markup driven approach. I can ask myself what would I love to interact with in the future. I write down a neat structure and see if I can map the reality to it. Maybe this helps you here to refactor the current markup which doesn't look very friendly to any future developer touching the contact form.

src/components/pages/contact/contact-form.tsx
```diff
@@ -1,148 +1,165 @@
...
+  return (
+    <TextArea
+      name={'message'}
+      label={t('contact.message')}
+      control={control}
+      rules={{ required: t<string>('contact.error.message') }}
+    />
+  );
+};
+
```
georgiee: I think you can collect those fields in a separate file `form-fields.tsx` to clean up the contact form file itself from the clutter.

src/components/pages/contact/contact-form.tsx
```diff
@@ -1,148 +1,165 @@
...
+    control,
+    formState: { isValid, isSubmitted },
+  } = useForm<FormData>({
+    defaultValues: {
+      name: '',
+      email: '',
+      message: '',
+      firstName: '',
+      phone: '',
+    },
```
georgiee: 1. I think you can remove all those defaults as they are empty which is the default by itself, isn't it?
2. Even if you want to set the default please make sure to exclude the honeypot fields. Ideally those fields are not visible to the developer describing the form to prevent the noise coming from their declaration.

src/components/pages/contact/form.tsx
```diff
@@ -14,18 +15,46 @@ const StyledLink = styled(Link)`
...
-      </Trans>
-    </SectionHeader>
-    <ContactForm />
-  </>
-);
+const StyledSectionHeader = styled(SectionHeader)`
+  margin-bottom: 48px;
+`;
+
+export const Form = () => {
```
georgiee: This is kind of upside down.
You export this component `Form` and internally you refer to a `ContactForm`. I would have expected to have an internal `Form` and you export `ContactForm`. Maybe you can swap the names ?

georgiee (changes requested):
Hey that looks much better. I would like that you invest a little time in refactoring the markup in the contact-form.tsx file as it's hard to read and I feat future developers will have problems reading it.

Apart from that and a few little comments we can almost merge 🙌 Thank you.

src/components/pages/contact/contact-form.tsx
hofmann-felix: > Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-> 
> components

The `defaultValues` were set to get rid of this warning. 
I removed the `defaultValues` now and added `value={field?.value || ''}` to `text-area.tsx` and `text-input.tsx`, which also prevents this warning

georgiee:
I'm unsure. Is this ready to be reviewed?

hofmann-felix:
@georgiee Yes, it's ready. Thanks for your feedback/review so far

src/assets/locales/de/translations.json
```diff
@@ -167,10 +167,13 @@
...
   "contact.error.email": "Deine E-Mail fehlt",
   "contact.error.email-undefined": "Irgendwas stimmt an dieser E-Mail nicht",
-  "contact.action.again-text": "Leider gab es einen Fehler. Bitte versuche es noch einmal. Klappt das nicht, schicke deine Nachricht bitte direkt an <2>beep@satellytes.com</2>",
+  "contact.action.again-text": "Leider gab es einen Fehler. Bitte versuche es noch einmal. Klappt das nicht, schicke deine Nachricht bitte direkt an",
   "contact.action.missing": "Bitte fülle alle benötigten Felder aus",
   "contact.action.send": "Senden",
-  "contact.action.sent": "Gesendet",
+  "contact.action.sent.headline": "E-Mail wurde gesendet",
+  "contact.action.sent.kicker": "Vielen Dank!",
+  "contact.action.sent.info": "Vielen Dank für Ihre Nachricht. Wir bearbeiten diese und melden uns schnellstmöglich bei Ihnen.",
```
georgiee, line 176: 👍

src/components/forms/text-area/text-area.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import { TextStyles } from '../../typography';
+import { theme } from '../../layout/theme';
+import { useController, UseControllerProps } from 'react-hook-form';
+import { StyledErrorMessage, Label } from '../text-input/text-input';
+
+const StyledTextArea = styled.textarea<{ hasError?: boolean }>`
```
georgiee, line 8: Thanks for the separate file

src/components/pages/contact/form.tsx
```diff
@@ -1,31 +1,103 @@
...
+      {isSubmitted && !isValid && (
+        <ErrorMessage>{t('contact.action.missing')}</ErrorMessage>
+      )}
+      {apiError && (
+        <ErrorMessage>
+          {t('contact.action.again-text')}{' '}
+          <Link to="mailto:info@satellytes.com">info@satellytes.com</Link>
+        </ErrorMessage>
+      )}
+    </form>
```
georgiee: looks great ✨ Thank you for your effort!

src/components/forms/text-field/text-input.tsx
georgiee: Thank you I spotted it

src/components/pages/contact/contact-form.tsx
georgiee: I think that helped a lot for readability and focus thanks

src/components/pages/contact/form.tsx
```diff
@@ -1,31 +1,103 @@
...
+import { Email, FirstName, MessageArea } from './form-fields';
+
+const API_ENDPOINT = '/api/contact-form';
+
+const FormLayout = styled.div`
+  margin-bottom: 24px;
+  gap: 24px;
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+`;
```
georgiee, line 20: Perfect without the 12 column grid

src/components/pages/contact/contact-form.tsx
georgiee: Greta new structure and I'm happy that you picked up my recommendations so it made sense I guess 🙏

src/components/pages/contact/contact-form.tsx
georgiee: Thank you!

src/components/pages/contact/form.tsx
georgiee: Reading much better know seeing this in code

src/components/pages/contact/contact-form.tsx
```diff
@@ -1,148 +1,60 @@
...
-  const { t } = useTranslation();
-  const [requestStatus, setRequestStatus] = useState<RequestStatus>('pending');
-  const {
-    register,
-    handleSubmit,
-    formState: { errors },
-  } = useForm();
-
-  const onSubmit = (formData: FormData) => {
-    setRequestStatus('submitting');
```
georgiee, line 43: @hofmann-felix this is crucial. Please bring back that combined with the disabled button to prevent multiple submissions while the request is loading

georgiee (changes requested):
Thank you for your efforts. We are done. Once crucial bug to fix and then we can merge finally. See my comment about the state `submitting` and the channel with the latest test submission.

src/components/pages/contact/form.tsx
```diff
@@ -1,31 +1,109 @@
...
+        <FirstName control={control} />
+        <Email control={control} />
+        <MessageArea control={control} />
+      </FormLayout>
+
+      <HoneypotField name="firstName" label="First Name" control={control} />
+      <HoneypotField name="phone" label="Phone" control={control} />
+
+      <MandatoryNotes>* {t('career.mandatory-field')}</MandatoryNotes>
+      <Submit disabled={requestStatus === 'submitting'} type={'submit'}>
```
georgiee, line 94: thanks for the disable 👍

src/components/ui/buttons/button.tsx
```diff
@@ -24,6 +24,11 @@ const StyledButton = styled(Link)`
   &:active {
     background: ${lighten(0.1, '#3e61ee')};
   }
+
+  &:disabled {
+    background: #3e61ee;
+    cursor: default;
+  }
```
georgiee, line 31: I thought the cursor should be handled correctly by the disabled state but I guess you had reasons for that?

georgiee (approved):
I see validation errors while submitting, also the state of the button seems to change once the submission is done. Something is off. Maybe you tests that while working on the submit button (and disabling the actual API).

https://user-images.githubusercontent.com/1701755/150531107-aac4c0e2-537a-4a97-8c73-99b3e4cb83c3.mp4



Anyway we are good for the current scope, let's merge finally 👍

src/components/ui/buttons/button.tsx
hofmann-felix: Yes it did not work without setting it explicitly

PR #316, opened 2022-01-05
chore(landingpage): slogan text update

georgiee:
here a quick update to remove our obsolete slogan.


abdolg (approved):
sounds passionate :D

PR #319, opened 2022-01-05
feat: contact form via gatsby function

### LINKED ISSUE #317: Submit Forms without Netlify
In order to get rid of netlify we need to use a different form service. Let's use [Gatsby Functions](https://www.gatsbyjs.com/products/cloud/functions/) (as suggested by @feedm3). I suggest to submit only to slack for the beginning. We can always think about sending mails afterwards.

georgiee:
This will remove the netlify dependency for our form submission and closes #317 so we can move over to gatsby deployment entirely.

✅ The functionality is confirmed by submitting the form. Will be received in our website channel.

## Description
Added a "gatsby function" a small serverless function to forward our contact form submissions to our Slack with the help of `@slack/web-api`. I created a new slack bot for this, registered the Slack API token in the gatsby cloud and tested the message format in some hidden channel.

The API endpoint is available here:
https://satellytescommain21751-gkgatsbyfunctionsslack.gtsb.io/api/contact-form

And the form makes use of it:
https://satellytescommain21751-gkgatsbyfunctionsslack.gtsb.io/contact/

To prevent too many spam bots I wanted to deliver a small honeypot with this PR that shows some hidden fields to bots, that are hopefully filled with data so we can filter them. I didn't want to introduce a captcha with this change (Possible hcaptcha.com which is used by cloudflare)

I also had to introduce a new status for the form as we never blocked multiple submissions (double click and such).

### DIFF
--- a/src/api/contact-form.ts
+++ b/src/api/contact-form.ts
@@ -1,0 +1,4 @@
+export default function handler(req, res) {
+  console.log(`submitted form`, req.body);
+  res.status(200).json({ hello: `world` });
+}
--- a/src/api/contact-form.ts
+++ b/src/api/contact-form.ts
@@ -1,17 +1,95 @@
 import { WebClient, LogLevel } from '@slack/web-api';
 
-export default async function handler(req, res) {
-  const channelId = 'website';
+const TOKEN = process.env.SLACK_BOT_SY_TOKEN;
 
-  const client = new WebClient(process.env.SLACK_BOT_SY_TOKEN, {
+/**
+ * Easily build this block structure via https://app.slack.com/block-kit-builder
+ */
+const createSlackMessage = ({ name, email, message }) => {
+  return [
+    {
+      type: 'header',
+      text: {
+        type: 'plain_text',
+        text: 'New form submission',
+      },
+    },
+    {
+      type: 'section',
+      text: {
+        type: 'mrkdwn',
+        text: 'We have received a new message.',
+      },
+    },
+    {
+      type: 'section',
+      fields: [
+        {
+          type: 'mrkdwn',
+          text: `*Name*:\n ${name}`,
+        },
+        {
+          type: 'mrkdwn',
+          text: `*E-mail*:\n ${email}`,
+        },
+      ],
+    },
+    {
+      type: 'divider',
+    },
+    {
+      type: 'section',
+      fields: [
+        {
+          type: 'plain_text',
+          text: message,
+        },
+      ],
+    },
+    {
+      type: 'context',
+      elements: [
+        {
+          type: 'mrkdwn',
+          text: 'Sent via _Satellytes Contact Form_',
+        },
+      ],
+    },
+  ];
+};
+
+export default async function handler(req, res) {
+  const { name, email, message, ...honey } = req.body;
+  console.log('receive form data', { name, email, message }, honey);
+
+  if (honey.firstName || honey.phone) {
+    console.log(`honeypot trap filled, aborting (received "${honey}")`);
+
+    // this is probably a bot as our honeypot fields have been filled
+    // let's silently ignore that bot
+    return res.status(200).json({ ok: true });
+  }
+
+  if (req.method !== `POST`) {
+    return res.status(500).json({ error: `not allowed` });
+  }
+
+  if (!TOKEN) {
+    return res.status(500).json({ error: `no slack token given` });
+  }
+
+  const channelId = 'website';
+  const client = new WebClient(TOKEN, {
     logLevel: LogLevel.DEBUG,
   });
 
   try {
-    // Call the chat.postMessage method using the WebClient
     const result = await client.chat.postMessage({
       channel: channelId,
-      text: 'Hello world',
+      // raw text as the fallback content for notifications.
+      text: `${name} (${email}) sent the following message:\n ${message}`,
+      blocks: createSlackMessage({ name, email, message }),
+      icon_emoji: ':inbox_tray:',
     });
 
     console.log(result);
@@ -19,6 +97,5 @@ export default async function handler(req, res) {
     console.error(error);
   }
 
-  console.log(`submitted form`, req.body);
-  res.status(200).json({ hello: `world` });
+  return res.status(200).json({ ok: true });
 }
--- a/src/api/contact-form.ts
+++ b/src/api/contact-form.ts
@@ -1,4 +1,5 @@
 import { WebClient, LogLevel } from '@slack/web-api';
+import { GatsbyFunctionRequest, GatsbyFunctionResponse } from 'gatsby';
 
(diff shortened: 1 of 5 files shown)


src/api/contact-form.ts
```diff
@@ -0,0 +1,101 @@
+import { WebClient, LogLevel } from '@slack/web-api';
+
+const TOKEN = process.env.SLACK_BOT_SY_TOKEN;
+
+/**
+ * Easily build this block structure via https://app.slack.com/block-kit-builder
```
georgiee, line 7: really nice tool I found after I wasted too much time in the documentation
Screen Shot 2022-01-06 at 07 20 44

src/api/contact-form.ts
```diff
@@ -0,0 +1,101 @@
...
+    logLevel: LogLevel.DEBUG,
+  });
+
+  try {
+    const result = await client.chat.postMessage({
+      channel: channelId,
+      // raw text as the fallback content for notifications.
+      text: `${name} (${email}) sent the following message:\n ${message}`,
+      blocks: createSlackMessage({ name, email, message }),
+      icon_emoji: ':inbox_tray:',
```
georgiee, line 96: this requires additional permission `chat:write.customize`.
Speaking of permissions: `chat:write` is the base permission to send messages. I also gave `channels:join` as I wasn't able to join the channel before.

src/components/pages/contact/contact-form.tsx
```diff
@@ -38,21 +60,12 @@ export const ContactForm: React.FC = () => {
...
-    // partialy taken from the Netlify Blog:
-    // - https://www.netlify.com/blog/2017/07/20/how-to-integrate-netlifys-form-handling-in-a-react-app/
-    const encodeForNetlify = (data: any): string => {
-      return Object.keys(data)
-        .map(
-          (key) =>
-            encodeURIComponent(key) + '=' + encodeURIComponent(data[key]),
-        )
-        .join('&');
-    };
```
georgiee, line 50: obsolete now as we can send json to the endpoint. I wonder why netlify did not allow that

src/api/contact-form.ts
```diff
@@ -0,0 +1,101 @@
...
+        {
+          type: 'mrkdwn',
+          text: 'Sent via _Satellytes Contact Form_',
+        },
+      ],
+    },
+  ];
+};
+
+export default async function handler(req, res) {
```
feedm3: You can declare the `req` and `res` argument with the Gatsby types to make it simpler to work with:

```
(req: GatsbyFunctionRequest, res: GatsbyFunctionResponse)
```

(not sure about the correct import here in Github, but you will figure it out 😉 )
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/index.d.ts#L1620-L1655

src/api/contact-form.ts
```diff
@@ -0,0 +1,101 @@
...
+  const { name, email, message, ...honey } = req.body;
+  console.log('receive form data', { name, email, message }, honey);
+
+  if (honey.firstName || honey.phone) {
```
feedm3, line 69: Very cool 👍 

Would be interesting to have a log service to see if and how often it works.

src/api/contact-form.ts
hofmann-felix: Looks very practical 💫 Nice that there is such a tool

src/components/pages/contact/contact-form.tsx
```diff
@@ -28,6 +30,26 @@ interface FormData {
   message: string;
 }
 
+/**
+ * This is a honeypot field but we want to disguise it including the wording
+ * and hope the bots can't filter them by any heuristic
+ * (hopefully ignoring this comment
+ **/
+
+const MagicFieldContainer = styled.div`
```
feedm3: `Magic` sounds funny but also a bit confusing. What about `HoneypotFieldContainer`?

feedm3 (approved):
Very straightforward and without any surprises 👍 

I only added some small suggestions, can be merged even without them.

src/components/pages/contact/contact-form.tsx
```diff
@@ -28,6 +30,26 @@ interface FormData {
...
+  ${hideVisually()}
+`;
+const MagicField = ({ label, control }) => {
```
hofmann-felix: Good idea 👍 I am curious whether this spam is intercepted

hofmann-felix (commented):
Nice changes. I haven't tested the contact form because you did this already. I think https://github.com/satellytes/satellytes.com/pull/310 should also work after a few little adjustments to it

hofmann-felix (approved):

src/components/pages/contact/contact-form.tsx
georgiee: I got kind of esoteric because I thought advanced spammers would read the source code variable names. That's not backed with any information so I revert and make this instead more understandable. We can always optimize afterwards. Thanks for the nudge 👍

src/api/contact-form.ts
georgiee: yeah, do you have an idea for that ?

src/api/contact-form.ts
georgiee: I bet I copied this from some example but that's a good idea indeed.

src/components/pages/contact/contact-form.tsx
feedm3: But after the code is compiled there is no readable variable name and comments anymore?

src/api/contact-form.ts
georgiee: I fixed that 👍

src/components/pages/contact/contact-form.tsx
georgiee: I changed the name to honeypot again and put it into a separate file. Thanks.

src/components/pages/contact/contact-form.tsx
georgiee: Yeah some logging would help as suggested by Fabian. Let's see.

src/api/contact-form.ts
feedm3: maybe panelbear works on server side? https://panelbear.com/docs/custom-events/

src/api/contact-form.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (req.method !== `POST`) {
+    return res.status(500).json({ error: `not allowed` });
+  }
+
+  if (!TOKEN) {
+    return res.status(500).json({ error: `no slack token given` });
+  }
+
+  const channelId = 'website';
```
georgiee, line 85: We could also use the more anonymous channel id but I think we are not harming anyone here by using the plain name 🙏

georgiee:
Thanks for your reviews. I just merged it.

PR #333, opened 2022-01-19
fix: make redirects work again

feedm3:
To make redirects work on Gatsby Cloud we need to add [`gatsby-plugin-gatsby-cloud`](https://www.gatsbyjs.com/plugins/gatsby-plugin-gatsby-cloud/#redirects). That's it. 

I also figured out that we still had the Netlify plugin installed. It's now removed.

**Test**
- `/blog/scoped-registry/ ` needs to redirect to `/blog/enterprises-benefit-from-scoped-npm-registries/`
- https://satellytescommain21751-fixredirects.gtsb.io/blog/scoped-registry/ 

Edit: It turns out that redirects already work on production. It's only a problem in Preview environments where Gatsby Cloud is not enabled. 
Nevertheless, instead of relying on that Gatsby Cloud injecting the Gatsby Cloud Plugin, we now have it defined ourselves.


fischeversenker:
I can confirm that it is working in the prod environment.

In the PR description you added the wrong test URL, @feedm3. You are using the netlify storybook preview deployment. Unfortunately it doesn't work with the [proper gatsby cloud URL](https://satellytescommain21751-fixredirects.gtsb.io/) (and [the URL that we expect to be redirected](https://satellytescommain21751-fixredirects.gtsb.io/blog/scoped-registry/)) neither. Since this does not fix the preview deployment and it is already working in the production deployments I don't see why we would need to add this change tbh. It makes perfect sense to remove the netlify plugin... but apparently there is no benefit of including the gatsby-cloud plugin.

In my opinion "the best code is no code at all" I suggest to remove the Netlify plugin but not to add the gatsby-cloud plugin - unless you can make it work with the preview deployments.

feedm3:
I changed the link in my description, it's now correct.

For the redirect on the Preview: I also fixed it. As Gatsby Cloud Hosting is only available on the production hosting, `redirectInBrowser` needs to be enabled for the preview environment. It's still a `301` redirect on production.

For the plugin itself: Not adding the plugin only works, because Gatsby Cloud is injecting it in our `gatsby-config.js` during buildtime. It's more of a fix when people forget to add it to `gatsby-config.js` on their own, [than the recommended approach](https://support.gatsbyjs.com/hc/en-us/articles/360061641114-Deploying-to-Gatsby-Cloud-Hosting). We even had the problem on a different project the version of the injected `gatsby-cloud-plugin` didn't work with our Gatsby version, as Gatsby Cloud is always injecting the latest version.

fischeversenker:
Thanks for the changes and the explanation. I agree that this is a meaningful change now.

I can confirm that the redirect works in the preview deployment now. The 404 page is visible for a split second before the redirect happens... but I guess there is no way around this? In the production env it happens instantly. Is that the difference between the status code redirect and the in-browser redirect?

fischeversenker (approved):

feedm3:
Yes. The Preview environment has the browser redirect as a fallback.

PR #334, opened 2022-01-19
chore: delete bash history

Wompipomp:
### Description
When entering the password into the bash for encoding the password is saved in the bash history which could be a security risk. This PR changes the text to delete the bash history afterwards or use JS REPL.

Thanks a lot to @fischeversenker for bringing this up :raised_hands:


Wompipomp:
I changed the text but I am not 100% happy with it. Lets see what you guys say :pray:

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
```diff
@@ -51,9 +51,11 @@ Even if the registry is hosted internally, it is good practice to add authentica
 
 `username:password` ⇒ base64 encoded ⇒ `dXNlcm5hbWU6cGFzc3dvcmQ=`
 
-For encoding you can use the command line `echo -n 'my-string' | base64` (Linux) or an IDE (vscode with plugin) or search for an online base64 encoder.
+For encoding you could use the command line `echo -n 'my-string' | base64` (bash) or an IDE (e.g. vscode with plugin). Or alternatively you could even use a REPL (e.g. [JS REPL](https://nodejs.dev/learn/how-to-use-the-nodejs-repl) with the following command in JavaScript `Buffer.from('username:password', 'binary').toString('base64')`).
```
fischeversenker, line 54: There is a "shorthand" for that `Buffer.from(...)` command: `btoa()` (https://developer.mozilla.org/de/docs/Web/API/btoa).

image


```suggestion
For encoding you could use the command line `echo -n 'my-string' | base64` (bash) or an IDE (e.g. vscode with plugin). Or alternatively you could even use a REPL (e.g. [JS REPL](https://nodejs.dev/learn/how-to-use-the-nodejs-repl) with the following command in JavaScript `btoa('username:password')`).
```

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
```diff
@@ -51,9 +51,11 @@ Even if the registry is hosted internally, it is good practice to add authentica
...
 
->💡 Use online encoder only for testing purposes. Never enter a productive username and/or password there.
+>💡 If you use the command-line, the password is saved in plaintext into the command-line history and can be retrieved from there. You should delete the password afterward (e.g. for bash you can delete the complete history with `history -c`)
```
fischeversenker: I've just learned that there is a way to bypass the history by adding a space in front of the command. On my machine that works but it seems to be configurable, so no guarantee that it works everywhere.

Generally I wouldn't advise people to clear their entire history with `history -c`. I would be absolutely lost and furious if that happened to me (over 10000 saved history items that make my daily workflow sooo much easier). I'd rather suggest to remove that single new entry that includes the password.

Here is a nice looking answer on stackexchange that I suggest to include. It also adds this trick to bypass the history by using a leading whitespace: https://unix.stackexchange.com/a/49216

georgiee: `history -C` is almost violent, please don't suggest that. When someone can extracts my history, then I already have other problems anyway. I think it's enough to inform the user about the attack vector and they will figure it out.

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
feedm3: Interesting, didn't know that!

Wompipomp: @fischeversenker maybe I did something wrong but I tried it with the btoa() command but it fails:
image

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
Wompipomp: @fischeversenker thank you for the effort :raised_hands:
Yes I agree, `history -c` isn't maybe a good thing :see_no_evil: I don't use bash history too much  (only the latest history) because it isn't shared in different bash terminals sometimes and so I almost never find the command I want.
I also found this or similar thread and tried it with history -d -1 but it gave me always 
image
So I think best would be to just leave it as @georgiee suggested.

Wompipomp: :see_no_evil: I agree, like I said I don't rely too much on the bash history as it is not consistent for me. I updated the text as I think that is the easiest and best solution :+1:

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
fischeversenker: Turns out this is not yet available in node 14. As you can see initially I'm in node 16 where it works, then I switch to node 14 (`nvm use 14`) and try again to see it fail:
image

I also had a look at the documentation. It turns out `atob` and `btoa` were added in Node 16 but they are already considered Legacy: https://nodejs.org/dist/latest-v16.x/docs/api/globals.html#btoadata
So ignore this and stick to the `Buffer.from(...)` approach 👌

fischeversenker (approved):
Looks and reads great now 🙌

blog-posts/enterprises-benefit-from-scoped-npm-registries.md
Wompipomp: :raised_hands: good to know :+1:

PR #337, opened 2022-01-24
fix: add image to blog description

### LINKED ISSUE #335: RSS feed is missing some images
Some images are either missing or have a broken URL in the RSS feed. They are not visible in Feedly:

Screenshot 2022-01-21 at 14 26 19

Haven't looked into it, but it has something to do with the rss.xml we create.

feedm3:
Feedly parses the description of a blog post to extract the image. Therefore we happen an image tag in every description. Let's see if it works with the deployed URL imported in Feedly.

**Update**
Appending it to the description doesn't work, will append it to the content

> If it works I will refactor the code

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -196,6 +196,7 @@ module.exports = {
             `,
             output: RSS_FEED_URL,
             title: 'Satellytes',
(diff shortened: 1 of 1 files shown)


feedm3:
Close this PR and create a new one to update the URL (gets cached by Feedly)

PR #339, opened 2022-01-25
feat: add redirect to our gather office

georgiee:
I would like to have a vanity url for our gather office. That way we can easily add the link to calendar invites and also invite other external folks. Our Gather space is protected by a password which means there is no risk to expose the url.

I have implemented a lambda function to redirect to the gather space and an additional redirect from the root of our website (which can't handle external urls).

Edit:
This will provide the vanity url http://satellytes.com/orion. I picked the name orion as it's an awesome [star constellation](https://en.wikipedia.org/wiki/Orion_(constellation)) and [nebula](https://en.wikipedia.org/wiki/Orion_Nebula) and sounds much more exicting than `virtual-office` or `gather`. I would like to give our gather space a name like `Orion` to so we are prepared for our hybrid times when we have to clarify where we are working (in the office or in orion).

### DIFF
--- a/src/api/redirect-gather-sy-office.ts
+++ b/src/api/redirect-gather-sy-office.ts
@@ -1,0 +1,13 @@
+import { GatsbyFunctionRequest, GatsbyFunctionResponse } from 'gatsby';
+
+const GATHER_URL = `https://gather.town/app/ea0xvXaHYWuWurME/satellytes`;
+
+/*
+Provide an endpoint to redirect to our gather space. 
+ */
+export default async function handler(
+  req: GatsbyFunctionRequest,
+  res: GatsbyFunctionResponse,
+) {
+  return res.redirect(GATHER_URL);
+}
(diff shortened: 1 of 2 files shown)


src/api/redirect-gather-sy-office.ts
```diff
@@ -0,0 +1,13 @@
+import { GatsbyFunctionRequest, GatsbyFunctionResponse } from 'gatsby';
+
+const GATHER_URL = `https://gather.town/app/ea0xvXaHYWuWurME/satellytes`;
+
+/*
+Provide an endpoint to redirect to our gather space. 
+ */
+export default async function handler(
```
georgiee, line 8: that works:
https://satellytescommain21751-gatherredirect.gtsb.io/api/redirect-gather-sy-office

georgiee:
The redirect does not work:
https://satellytescommain21751-gatherredirect.gtsb.io/orion/

Does gatsby prevent redirect to non-pages 🤨

feedm3 (approved):

georgiee:
Spoken with @feedm3 and decided to give it a try. It doesn't work in the branch preview but the prod deployment is handled differently by the gatsby cloud plugin (as we host on gatsby cloud). Let's see 🤞

PR #344, opened 2022-02-01
feat: add attribution to cover image in rss feed

feedm3:
Our cover image has currently no attribution in the RSS feed. Therefore I added it. Let's see how it looks in Feedly.

RSS URL: https://satellytescommain21751-featrssattribution.gtsb.io/blog/rss.xml

**Result**
Not sure if we want this. The caption gets used as content preview. Makes the preview less readable. In the detailed view it's ok though.
- Felix post has no caption
- Gatsby and NPM post have a caption

<details>
<summary>Screenshot overview</summary>
grafik
</details>
<details>
<summary>Screenshot detail</summary>
grafik

</details>

**Alternative**
We put the attribution at the end of the article, like `Cover image by XY - example.com`. Probably semantically not as correct as the `figcaption`, but still better than not attribution at all.


gatsby-config.js
```diff
@@ -135,6 +135,17 @@ module.exports = {
                     BASE_URL +
                     (blogPostMeta.shareImage.childImageSharp.fixed.src ||
                       DEFAULT_META_IMAGE_URL_PATH);
+                  const coverImage =
+                    blogPostMeta.attribution && blogPostMeta.attribution.creator
+                      ? `
+                    <figure>
+                      <img src='${imageUrl}' alt=''>
+                      <figcaption>${blogPostMeta.attribution.creator} - ${blogPostMeta.attribution.source}</figcaption>
```
georgiee:
```suggestion
                      <figcaption>Image by ${blogPostMeta.attribution.creator} - ${blogPostMeta.attribution.source}</figcaption>
```

This should clarify what it is. I'm okay with presenting the attributes here, we don't need to hide them.

georgiee (approved):
Thanks for your thoughts around this. I like your solution and I don't want to hide them, we use the image prominently so we can provide the attribution in the same way. Maybe smaller than the text would be good, but it's okay for now. It's a very specific case for feedly/rss folks anyway.

Maybe you can accept my addition which makes it clear that this is not an author but the image attribution.

gatsby-config.js
feedm3: Good idea. I will also try to combine the text and the link with an `a` tag inside `figcaption`. And add a separtor between the caption and the preview text. Ideal solution:

```
Image by Daniel Kudela · Before starting a React based project, the same question often arises: 
Which framework should be used? Of course, this question cannot be answered in a general way, 
since each framework has its advantages and disadvantages for specific use cases. Nevertheless, 
you can compare the frameworks, especially the concepts behind them.
```

feedm3:
Check this out, I think it's pretty good right now:

RSS: https://build-3df0e726-c161-46f7-bad1-e013fc23118e.gtsb.io/blog/rss.xml

<details>
<summary>Screenshot overview</summary>


</details>
<details>
<summary>Screenshot detail</summary>

</details>

The `·` looks a bit missplaced in the detailed view, but it makes the preview nicer. So I would vote to leave it in (you should read the detailed article on the Satellytes page anyway).

gatsby-config.js
```diff
@@ -135,6 +135,17 @@ module.exports = {
                     BASE_URL +
                     (blogPostMeta.shareImage.childImageSharp.fixed.src ||
                       DEFAULT_META_IMAGE_URL_PATH);
+                  const coverImage =
+                    blogPostMeta.attribution && blogPostMeta.attribution.creator
+                      ? `
+                    <figure>
+                      <img src='${imageUrl}' alt=''>
+                      <figcaption><a href='${blogPostMeta.attribution.source}' target='_blank' rel="nofollow noreferrer">Image by ${blogPostMeta.attribution.creator} · </figcaption>
```
georgiee, line 143: I looked for nofollow noreferrer 🙌 Thanks!

georgiee (approved):
Let's ship, thank you!

PR #347, opened 2022-02-03
fix(career-form): make upload form accept names with special chars

### LINKED ISSUE #346: Career Form Upload Issues
1. Can't handle files with `,` in the file name
Edit: The minimal file name is :`,.pdf` which is valid on osx.
I have not checked the code but it's weird. If you have a file with the name `a (b, c).pdf` and you select or drop it into the file upload it throws an exception. See me screen record. You can easily create such a file with `touch "a (b, c).pdf"` and upload that empty file as we don't check the file content here.

It works for other files as expected. I had a real file and I just condensed the error down to this format.
https://user-images.githubusercontent.com/1701755/152187171-5c05000a-ffa9-4461-bcf7-eb0a417c269a.mp4

See https://github.com/satellytes/satellytes.com/pull/347 for a working fix, although I don't know if that has side effect.

2. Missing error key `career.error.message`
3. Drop is working but the validation doesn't respect the dropped file so you can't proceed.

Feel free to test the form submission as much as you want to fix the form 🙏

georgiee:
close #346

I found the reason for the upload form to fail in the file type dropdown which relies on an implicit data structure that is build using the file names. If the file name contains a comma, this fails. 

The fix is to create a sanitized id from a name by replacing anything that is not an alphanumeric character. That's something internal and does not change the appearance.

### DIFF
--- a/src/components/pages/career-details/career-form/career-file-upload-type.tsx
+++ b/src/components/pages/career-details/career-form/career-file-upload-type.tsx
@@ -82,8 +82,13 @@ export const CareerFileUploadType = ({
   watch,
 }: CareerFileUploadTypeProps) => {
   const { t } = useTranslation();
-  const fileName = file.name.split('.');
-  const name = `category_select.${fileName[0]}`;
+  const fileName = file.name.split('.')[0];
+  /**
+   * Create a sanitized file id by replacing anything that's not a letter which can be safely used as an identifier.
+   * Without this, a valid filename like `a,b.pdf` (osx) will create an identifier to safely use as an internal name.
+   */
+  const fileId = fileName.replace(/[^a-z0-9]/gi, '_').toLowerCase();
+  const name = `category_select.${fileId}`;
   const selectedFileType = watch(name);
 
   const onChange = (event) => {
@@ -114,7 +119,7 @@ export const CareerFileUploadType = ({
         </StyledSelect>
         <SelectArrow />
       </SelectContainer>
-      {errors.category_select?.[fileName[0]] && (
+      {errors.category_select?.[fileId] && (
         <FormErrorWrapper>
           <FormError
             error={{ message: t<string>('career.error.selection') }}


feedm3 (approved):
Nice catch 👍

src/components/pages/career-details/career-form/career-file-upload-type.tsx
```diff
@@ -82,8 +82,13 @@ export const CareerFileUploadType = ({
   watch,
 }: CareerFileUploadTypeProps) => {
   const { t } = useTranslation();
-  const fileName = file.name.split('.');
-  const name = `category_select.${fileName[0]}`;
+  const fileName = file.name.split('.')[0];
```
fischeversenker, line 85: What does this do with file names that contain multiple dots like `submission.final.pdf`? It looks like it would discard the `final` in this example? Since this is only internal I guess it doesn't matter... but might lead to communication issues ("I've uploaded a file called abc.xzy.pdf, didn't you get it?"). I understand that it was like this before, but we might as well fix it now?

src/components/pages/career-details/career-form/career-file-upload-type.tsx
```diff
@@ -82,8 +82,13 @@ export const CareerFileUploadType = ({
...
+  /**
+   * Create a sanitized file id by replacing anything that's not a letter which can be safely used as an identifier.
```
fischeversenker, line 87: *"letter or digit"
Instead of this comment, why not call the variable `sanitizedFileId`, which would make parts of the comment obsolete?

src/components/pages/career-details/career-form/career-file-upload-type.tsx
```diff
@@ -82,8 +82,13 @@ export const CareerFileUploadType = ({
...
+   * Without this, a valid filename like `a,b.pdf` (osx) will create an identifier to safely use as an internal name.
```
fischeversenker, line 88: I don't understand this sentence...

georgiee: Early morning fix. I will update it. Sorry.

src/components/pages/career-details/career-form/career-file-upload-type.tsx
georgiee: Yeah the behavior was the same before. This is a hotfix, I just wanted to fix the fact that people can't upload files with comma in the name right now. It just fails and I have very limited time today. I would be happy if someone helps out to improve the PR in that direction or create a follow up PR 🙏

georgiee:
I have talked to @hofmann-felix as there are other small issues (see #346 ) that we should include in a fix. I will close this PR and the fix for the `,` issue can be cheery picked from my branch. Thank you 🙏

PR #350, opened 2022-02-04
chore: update dependencies partially

feedm3:
I updated all dependencies, except
- `react-hook-form` - too much effort to migrate)
- `storybook/addon-docs` needs to be added - looks like it's not a peer dependency anymore (we already use it)
- `storybook-addon-gatsby` can't be upgraded, as the latest version is broken ATM

**Update**
It seems like the build is stuck. I don't know why:

Screenshot 2022-02-04 at 10 50 25


fischeversenker:
Did you see the broken build and are you taking care of it, @feedm3?

georgiee:
>  It seems like the build is stuck. I don't know why:

I think this PR is obsolete when we are moving forward with dependabot and can be closed anyway? I hope being stuck will be solved then 😅

Issue #366, opened 2022-02-04
Status: dependabot


georgiee:
For the current status see https://github.com/satellytes/satellytes.com/network/updates
based on the file https://github.com/satellytes/satellytes.com/blob/b41170ddf5847cbf71c831604109af0ecd648034/.github/dependabot.yml

Issues:
1. We need to build through github actions as we don't know if it's works due to the missign gatsby build (expected as they don't build on automatic PRs, @feedm3  found out about that).
2. It looks super annoying right now with the many PRs. But I I think it's just because we started. I should be better over time with fewer dependencies. dependencies are grouped by peer dependencies so gatsby update should be a single PR in the best case
3. I think we should enable auto merge for dependabot if github actions are testing & building everything. Makes it much easier to handle the single PRs

georgiee:
we need to add [skip netlify] to the commit messages otherwise our netlify build minutes will run out I think

feedm3:
What we need to do:
- Add `[skip ci]` to the commit messages to skip netlify builds https://www.netlify.com/blog/2019/10/10/intelligent-deploy-skipping-an-automatic-optimisation/
- Run a manual build with Github Actions, to make sure the dependency update doesn't break anything
- Make Gatbsy Cloud Build optional. That's a bit unfortunate, as they are only optional for Dependabot PRs and not manual PRs

Check https://github.com/satellytes/satellytes.com/pull/367 for the implementation.

Optionally:
- Merge Dependabot PRs automatically after all checks are green

fischeversenker:
> Automatic merge request for Depnendabot PRs

What do you mean, @feedm3?

Generally, how do we proceed here? I guess after #367 got merged we can rebase all the open Dependabot PRs so they trigger the build GitHub Workflow? And then we merge all the green ones? 🎉

feedm3:
Ups, my brain was somewhere else when I wrote this :D 

I mean we can add a workflow that will merge the Dependabot PRs automatically once all checks are green.

feedm3:
I just found the Feature Request for Dependabot grouped / batched updates. The thread is locked, but it looks like it is on their roadmap. Unfortunately without an ETA: https://github.com/dependabot/dependabot-core/issues/1190

georgiee:
[...]

PR #367, opened 2022-02-07
feat: skip netlify on dependabot pull requests

feedm3:
Improvements needed for #366 

We have some problems with Dependabot that we need to solve with this PR:
- Dependabot creates a lot of PRs. For every PR, Netlfiy starts a build. The free tier is quite limited, so within a few days we would run out of it. That's why we disable Netlify builds in Dependabot PRs
- Gatsby Cloud doesn't run on automatically created PRs. Even worse: They idle in the PR checks and don't do anything. That's why we need to make them optional _for all PRs_. To know if the dependency update still breaks our build, we run the build with Github Actions (only for Dependabot PRs). The free tier should be large enough for that.

With this setup:
- Depndabot PRs will trigger automatic tests and build with Github Actions
- Manual PRs will trigger Gatsby Cloud builds, Netlify Builds for Storybook and tests with Github Actions

If we see this setup is working, we can also add automatic merges of Dependabot PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,6 @@
 version: 2
 
 updates:
-  # npm should be used for npm and yarn
   - package-ecosystem: npm
     open-pull-requests-limit: 10
     directory: "/"
@@ -11,7 +10,7 @@ updates:
       time: "06:00"
       timezone: "Europe/Berlin"
     commit-message:
-      # we need to disable netlify as we don't have enough builds for all dependency updates
+      # we need to disable netlify as we don't have enough build minutes for all dependency updates
       # https://docs.netlify.com/site-deploys/manage-deploys/#skip-a-deploy
       prefix: "[skip netlify]"
     labels:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-      day: "sunday"
+      day: "monday"
+      time: "06:00"
+      timezone: "Europe/Berlin"
     commit-message:
       # we need to disable netlify as we don't have enough builds for all dependency updates
       # https://docs.netlify.com/site-deploys/manage-deploys/#skip-a-deploy
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,15 @@
 version: 2
 
 updates:
+  # npm should be used for npm and yarn
   - package-ecosystem: npm
     open-pull-requests-limit: 10
     directory: "/"
     schedule:
       interval: "weekly"
       day: "sunday"
+    commit-message:
+      # we need to disable netlify as we don't have enough builds for all dependency updates
+      prefix: "[skip ci]"
     labels:
       - dependencies
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,7 @@ updates:
       day: "sunday"
     commit-message:
       # we need to disable netlify as we don't have enough builds for all dependency updates
-      prefix: "[skip ci]"
+      # https://docs.netlify.com/site-deploys/manage-deploys/#skip-a-deploy
+      prefix: "[skip netlify]"
     labels:
       - dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,27 @@
+name: Build
+
+on:
+  pull_request:
+    types: [ labeled ]
+
+jobs:
+  build:
+    # only run a manual build when the PR comes from gatsby
+    # as gatsby doesn't run on automated pull requests
+    if: ${{ github.event.label.name == 'dependencies' }}
+    name: Build
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+
+      - name: Install
+        run: yarn
+
+      - name: Build
+        run: yarn build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,13 +2,13 @@ name: Build
 
 on:
   pull_request:
-    types: [ labeled ]
+    types: [ labeled, opened, synchronize, reopened ]
 
 jobs:
   build:
-    # only run a manual build when the PR comes from gatsby
+    # only run a manual build when the PR comes from Dependabot
     # as gatsby doesn't run on automated pull requests
-    if: ${{ github.event.label.name == 'dependencies' }}
+    if: contains(github.event.pull_request.labels.*.name, '<label_name>')
     name: Build
 
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -1,11 +1,15 @@
 version: 2
 
 updates:
+  # npm should be used for npm and yarn
```
fischeversenker: Dependabot updates the `yarn.lock` file (i.e. uses `yarn`), so I'm confused enough by this `package-ecosystem` property being set to `npm`... I'm afraid your comment doesn't clarify anything for me. What do you mean with this comment? What does the property actually do?

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
+name: Build
+
+on:
+  pull_request:
+    types: [ labeled ]
```
fischeversenker: If I'm not mistaken, this will only trigger this workflow **once** when the PR is labeled (immediately after it got created). It won't trigger the workflow for future events like rebases, etc. I suggest we add other event types and make the label check more robust. See this SO answer: https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label#comment122159108_62331521

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  build:
+    # only run a manual build when the PR comes from gatsby
```
fischeversenker: Did you mean to write dependabot?
```suggestion
    # only run a manual build when the PR comes from dependabot
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+
+      - name: Install
+        run: yarn
```
fischeversenker: I suggest we use `yarn install --frozen-lockfile` here (and in the other workflows). Since this workflow is especially sensible to dependency version changes, let's install the exact ones specified in the lock file.

.github/dependabot.yml
feedm3: I guess it tells Dependanbot what tools to use. I was also confused by it, as other tools (like lerna) specify `yarn` when working with the `yarn.lock` file and not `npm`.

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem

.github/workflows/build.yml
feedm3: Good hint, will add it 👍

.github/workflows/build.yml
feedm3: Yep, thanks.

.github/workflows/build.yml
feedm3: Yep, good idea, thanks 👍

.github/dependabot.yml
georgiee: I think this confused everyone of us now 😅 It's because of this entry on the linked page. They will use yarn but you still have to signal it with `npm` 
Screen Shot 2022-02-07 at 10 09 04

.github/dependabot.yml
```diff
@@ -1,11 +1,15 @@
...
   - package-ecosystem: npm
     open-pull-requests-limit: 10
     directory: "/"
     schedule:
       interval: "weekly"
       day: "sunday"
+    commit-message:
+      # we need to disable netlify as we don't have enough builds for all dependency updates
+      prefix: "[skip ci]"
```
georgiee: let's use the `[skip netlify]` which is a valid alternative and makes it much clearer that we only target netlify.

> Sometimes, you may want to push commits to your production branch without triggering a deploy on Netlify. 
> To do this, add [skip ci] or [skip netlify] anywhere in the Git commit message.

https://docs.netlify.com/site-deploys/manage-deploys/

.github/workflows/build.yml
georgiee: So it's  `[opened, synchronize, labeled, reopened]` I guess?

```suggestion
    types: [ opened, synchronize, labeled, reopened ]
```

.github/dependabot.yml
```diff
@@ -1,11 +1,15 @@
 version: 2
 
 updates:
+  # npm should be used for npm and yarn
   - package-ecosystem: npm
     open-pull-requests-limit: 10
     directory: "/"
     schedule:
       interval: "weekly"
       day: "sunday"
```
georgiee: Can we sneak this in? I think Monday morning save us a few thoughts about the website on the weekend.

```suggestion
    schedule:
      interval: 'weekly'
      day: "monday"
      time: '06:00'
      timezone: 'Europe/Berlin'
```

georgiee (approved):
Let's try this and thanks @fischeversenker for the catches!

.github/workflows/build.yml
fischeversenker: According to the SO answer these seem to be the most important ones, yes. But with these new event types we need to also change the `if` logic in the `build` job because most of these events won't have the `event.label.name` property so the workflow will throw an error. We need to use the
```
if: contains(github.event.pull_request.labels.*.name, 'dependabot')
```
logic.

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on:
+  pull_request:
+    types: [ labeled, opened, synchronize, reopened ]
+
+jobs:
+  build:
+    # only run a manual build when the PR comes from Dependabot
+    # as gatsby doesn't run on automated pull requests
+    if: contains(github.event.pull_request.labels.*.name, '<label_name>')
```
fischeversenker:
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'dependabot')
```

.github/dependabot.yml
feedm3: Makes sense, also added the link as comment

feedm3:
Thanks for the feedback, I added all your suggestions.

.github/dependabot.yml
fischeversenker: I claim this comment does more harm than good. Let's remove it. Whoever want's to know why it says `npm` should refer to the documentation. Right now I don't see any benefit in keeping it. It obviously causes confusion.
```suggestion
```

.github/dependabot.yml
```diff
@@ -1,11 +1,18 @@
...
     open-pull-requests-limit: 10
     directory: "/"
     schedule:
       interval: "weekly"
-      day: "sunday"
+      day: "monday"
+      time: "06:00"
+      timezone: "Europe/Berlin"
+    commit-message:
+      # we need to disable netlify as we don't have enough builds for all dependency updates
```
fischeversenker:
```suggestion
      # we need to disable netlify as we don't have enough build minutes for all dependency updates
```

fischeversenker (approved):
Since, this (probably) works perfectly fine, I'm approving.

You might still want to have a look at my comments. Most of the time I'm not a fan of comments in code. They are either wrong or outdated.

PR #368, opened 2022-02-07
feat: upgrade storybook

feedm3:
Manual PR to upgrade Storybook, as the new version needs a new package `@storybook/addon-docs`


package.json
```diff
@@ -99,12 +99,13 @@
...
   "devDependencies": {
     "@babel/core": "~7.16.7",
-    "@storybook/addon-actions": "^6.4.13",
-    "@storybook/addon-essentials": "^6.4.13",
-    "@storybook/addon-links": "^6.4.13",
-    "@storybook/builder-webpack5": "^6.4.13",
-    "@storybook/manager-webpack5": "^6.4.13",
-    "@storybook/react": "^6.4.13",
+    "@storybook/addon-actions": "^6.4.18",
+    "@storybook/addon-docs": "^6.4.18",
```
fischeversenker: On the documentation page for the `docs` addon, it says:
> we recommend that you install [@storybook/addon-essentials](https://www.npmjs.com/package/@storybook/addon-essentials), to include this and other great features into your project.

([source](https://storybook.js.org/docs/react/writing-docs/introduction))

The `docs` addon was already in our lock file so it seems like it is a dependency of some other dependency... are you sure we need to explicitly list it as a dependency?

From `main` (running `yarn why @storybook/addon-docs`):
image

georgiee: good point👍

feedm3: I'm 100% (ok, maybe 99% only) sure that the build didn't work last week because this addon was missing. I just checked it locally, it works without the extra dependency. So weird 😖 

Will remove it again, thanks for the double check.

package.json
```diff
@@ -130,8 +130,8 @@
     "lint-staged": "12.2.0",
     "prettier": "2.5.1",
     "serve": "^13.0.2",
-    "storybook-addon-gatsby": "^0.0.3",
-    "storybook-react-i18next": "^1.0.16",
+    "storybook-addon-gatsby": "0.0.3",
```
fischeversenker, line 133: Why did you decide to pin this version down? Is there something wrong with `0.0.4` or `0.1.0`?

fischeversenker (approved):
The changes look good, thanks for the updates.

Just one question. It doesn't block this PR, though. Approving 🚀

package.json
feedm3: yes, `0.0.4` (the latest) is broken

PR #382, opened 2022-02-08
feat: add caching to github actions and also build storybook

feedm3:
Some more housekeeping:

- add caching between Github Actions to speed up build time
- build Storybook within Dependabot PRs
- remove `colors` resolution, as it was updated by all of our `peerDependencies`
- remove outdated build command

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
-          cache: 'npm'
+          cache: 'yarn'
       - name: Install
         run: yarn install --frozen-lockfile
       - name: Build
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
-          cache: 'npm'
+          cache: 'yarn'
       - name: Install
         run: yarn install --frozen-lockfile
       - name: Build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,23 +5,36 @@ on:
     types: [ labeled, opened, synchronize, reopened ]
 
 jobs:
-  build:
+  build-website:
     # only run a manual build when the PR comes from Dependabot
     # as gatsby doesn't run on automated pull requests
     if: contains(github.event.pull_request.labels.*.name, 'dependencies')
-    name: Build
-
+    name: Build Website
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
-
+          cache: 'npm'
       - name: Install
         run: yarn install --frozen-lockfile
-
       - name: Build
         run: yarn build
+
+  build-storybook:
+    if: contains(github.event.pull_request.labels.*.name, 'dependencies')
+    name: Build Storybook
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -5,23 +5,36 @@ on:
...
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
-
+          cache: 'npm'
```
fischeversenker: I guess we should be using `yarn` here 👌 
https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies

Why it makes a difference: https://github.com/actions/cache/blob/main/examples.md#node---yarn

package.json
```diff
@@ -141,7 +140,6 @@
     ]
   },
   "resolutions": {
-    "webpack": "^5",
-    "colors": "1.4.0"
```
georgiee, line 145: bye & thanks!

package.json
```diff
@@ -27,8 +27,7 @@
     "precommit": "lint-staged",
     "storybook": "start-storybook --port 6006",
     "storybook:build": "build-storybook",
-    "storybook:serve": "serve storybook-static",
-    "build-storybook": "yarn storybook:build"
```
georgiee, line 31: why did we have that duplication here 😅

.github/workflows/test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
+          cache: 'npm'
```
georgiee: yeah, do you have numbers? That's caching the entire install which should save us a few minutes, right ?

.github/workflows/build.yml
```diff
@@ -5,23 +5,36 @@ on:
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+          cache: 'npm'
+      - name: Install
+        run: yarn install --frozen-lockfile
       - name: Build
-        run: yarn build
\ No newline at end of file
+        run: yarn storybook:build
```
georgiee: github warns about the line ending. let's fix that before some yaml parser fails because of that in the future. We have end of lines for a good reason although I never had any actual problems these days.
```suggestion
        run: yarn storybook:build

```

.github/workflows/build.yml
```diff
@@ -5,23 +5,36 @@ on:
...
         with:
           node-version: '16'
-
+          cache: 'npm'
       - name: Install
         run: yarn install --frozen-lockfile
+      - name: Build
+        run: yarn build
 
+  build-storybook:
```
georgiee, line 26: cool, thanks for including that build too

.github/workflows/build.yml
georgiee: ha, good catch (you catch many things, thanks for the open eyes 🤩).
I would have not seen this because of the discussion yesterday in the context of dependabot

georgiee (approved):
Good from my side. Some minor things and the important note from @fischeversenker regarding npm/yarn otherwise good!

package.json
feedm3: old open pull requests that were relying on it

.github/workflows/test.yml
feedm3: total time of the workflow right now is 7-8 minutes, install time is 1-2 minutes. so we probably save around 10% - 20% of time

PR #383, opened 2022-02-08
chore: bump react-form-hooks and update types

georgiee:
# Description
I went into the rabbit hole of updating the react form hooks package which refined the types at some point that revealed some issues in our current form code. I think it does not only reveal typing issues but the actual mess we have inside the career form component.

I managed to fix the types. I think the code should still work as I only focused on types and at some point I made the code more rigid by removing the flexibility to define a form name and instead hard code it  (`documents` in the upload component). This was was required as I didn't want to tinker around with typing programming (if this is feasible anyway in this context). I pinned the type to the string value `documents` because of that.

The most changes come from swapping the generic `FieldValues` type with our concrete `CareerFormValues` type. This creates some hard dependencies but to be honest, the components around the career from are one spaghetti plate anyway so this was a good way to fix the type issues while we still benefit from the types in the future

Speaking of the future: We should refactor that mess. This is not a failure of our juniors who mostly build that, but we missed the opportunity to refactor together with them from time to time I guess. Let's tackle this when we have some spare time. It's actually a nice challenge.

# How to review
Let's test the career submission at least once. As said technically nothing changed in the data structure but we have to make that sure.

### DIFF
--- a/src/components/pages/career-details/career-form/career-file-upload-type.tsx
+++ b/src/components/pages/career-details/career-form/career-file-upload-type.tsx
@@ -88,7 +88,7 @@ export const CareerFileUploadType = ({
    * Without this, a valid filename like `a,b.pdf` will cause an error.
    */
   const fileId = createFileId(file.name);
-  // TODO: CareerFormValues contains a list of keys in field `category_select` which we should reflect here
+  // TODO: CareerFormValue contains a list of keys in field `category_select` which we should reflect here
   const name: any = `category_select.${fileId}`;
   const selectedFileType = watch(name);
 
--- a/src/components/pages/career-details/career-form/career-file-upload-type.tsx
+++ b/src/components/pages/career-details/career-form/career-file-upload-type.tsx
@@ -3,7 +3,6 @@ import { FormError } from './career-components';
 import styled from 'styled-components';
 import {
   FieldErrors,
-  FieldValues,
   UseFormClearErrors,
   UseFormRegister,
   UseFormSetValue,
@@ -13,12 +12,13 @@ import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { up } from '../../../support/breakpoint';
 import { SelectArrow } from '../../../legacy/icons/form-icons/select-arrow';
 import { createFileId } from './utils';
+import { CareerFormValues } from './career-form';
 
 interface CareerFileUploadTypeProps {
-  setValue: UseFormSetValue<FieldValues>;
-  clearError: UseFormClearErrors<FieldValues>;
-  register: UseFormRegister<FieldValues>;
-  watch: UseFormWatch<FieldValues>;
+  setValue: UseFormSetValue<CareerFormValues>;
+  clearError: UseFormClearErrors<CareerFormValues>;
+  register: UseFormRegister<CareerFormValues>;
+  watch: UseFormWatch<CareerFormValues>;
   errors: FieldErrors;
   file: File;
 }
@@ -41,7 +41,7 @@ const StyledSelect = styled.select`
     width: unset;
     display: inline-block;
     min-width: 144px;
-    margin-top: 0px;
+    margin-top: 0;
   }
 `;
 
@@ -88,7 +88,8 @@ export const CareerFileUploadType = ({
    * Without this, a valid filename like `a,b.pdf` will cause an error.
    */
   const fileId = createFileId(file.name);
-  const name = `category_select.${fileId}`;
+  // TODO: CareerFormValues contains a list of keys in field `category_select` which we should reflect here
+  const name: any = `category_select.${fileId}`;
   const selectedFileType = watch(name);
 
   const onChange = (event) => {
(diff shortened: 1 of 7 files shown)


src/components/pages/career-details/career-form/career-form.tsx
```diff
@@ -240,19 +258,20 @@ export const CareerForm: React.FC<CareerFormProps> = (props) => {
...
-    <CareerFormStyled onSubmit={handleSubmit(onSubmit, onError)}>
+    <CareerFormStyled onSubmit={onSubmit}>
       <Headline>{t('career.headline')}</Headline>
       <Fieldset disabled={isSubmitting}>
         <Grid nested>
           <CareerTextFields register={register} errors={errors} />
 
           {/*File-Upload*/}
           <FileUpload
+            name={'documents'}
```
fischeversenker: I can't tell any difference between the old and the new way of passing this property (`name={'documents'}` vs. `name="documents"`). In addition, you hardcoded the value in the `FileUploadProps` interface... do we actually need to pass this at all? One of the places smells.

src/components/pages/career-details/career-form/career-form.tsx
```diff
@@ -32,15 +32,25 @@ interface CareerFormProps {
   scrollToStart?: () => void;
 }
 
-interface RawFormData {
+interface CategorySelect {
+  [key: string]: string;
+}
+
+export type CareerFormValues = {
```
fischeversenker: I don't particularly like the plural here. It's only a single value in my opinion.

src/components/pages/career-details/career-form/career-form.tsx
feedm3: `name={'documents'}` and `name="documents"` are the same. I prefer the latter but consistency within the project overrules that.

src/components/pages/career-details/career-form/career-form.tsx
feedm3: +1

Plural is an array, this is an interface/object.

src/components/pages/career-details/career-form/career-form.tsx
georgiee: Sorry folks. My IDE updated this. I removed it in a first step and inlined `documents` everywhere in the target component instead of using a name prop. In a second iteration I decided not to be so invasive and pinned the type to the string literal `documents` which fixed the typing issues. Without that TS could not infer that we refer to the `documents` key and I didn't want to find some type magic for that.

src/components/pages/career-details/career-form/career-form.tsx
georgiee: good point, I will fix that.

georgiee: Oh I see where my mind came from: `FieldValues` is the internal default. Anyway, I updated this.

src/components/pages/career-details/career-form/career-file-upload-type.tsx
```diff
@@ -88,7 +88,8 @@ export const CareerFileUploadType = ({
    * Without this, a valid filename like `a,b.pdf` will cause an error.
    */
   const fileId = createFileId(file.name);
-  const name = `category_select.${fileId}`;
+  // TODO: CareerFormValues contains a list of keys in field `category_select` which we should reflect here
```
fischeversenker:
```suggestion
  // TODO: CareerFormValue contains a list of keys in field `category_select` which we should reflect here
```

src/components/pages/career-details/career-form/career-form.tsx
fischeversenker: Not sure I understand this correctly. Does it still make a difference if you use the "unbound" way (`name="documents")`? Either way, fine with me. Thanks for the explanation 👌

fischeversenker (approved):
Looks good, thanks :+1:
Two more comments, no blockers.

I trust the pipeline on this one, didn't run it locally.

src/components/pages/career-details/career-form/career-form.tsx
georgiee: reverted the assignment.

@fischeversenker yes it smells a lot but it's the most pragmatic while functional approach to adhere to the typing given by the library. The `register` functions etc are all typed. This means if you create a sub components you have to relate that component somehow to the parent type (here the `CareerFormValue`).

I'm glad I found a way that is minimal invasive by pinning the type, hence the following change you mentioned:

```
interface FileUploadProps {
  name: 'documents';
```

A `name: string` will not work as the derived functions `register`, `setError` etc. all refer to the parent `CareerFormValue` type and an untyped name will not work.

I consider the entire form a technical debt. That's why I decided not to spend more time than useful on this topic. I hope this is okay for you, otherwise I'm happy to hear how you would approach this besides refactoring all involved components 🙏

src/components/pages/career-details/career-form/career-file-upload-type.tsx
georgiee: fixed.

feedm3 (approved):
I did a test on the Staff Engineering page and everything still works. Code looks also good to me. Well done!

georgiee:
Thanks for your reviews. Fabian did an AP test, I run second one and checked the data that arrived. Everything looked good. I feel secure enough to merge this now 🙌

PR #384, opened 2022-02-08
fix: adjust leaflet map zoom level

### LINKED ISSUE #352: Design: Viewport of map
Please change the default zoom level of the map on the contact page. It should look something like this:
https://cln.sh/iN3jb5 and https://cln.sh/jaRXa0.

If it's possible to set different levels for desktop and mobile, refer to the two screenshots, if not, any of the two levels are fine. In my screenshots the "Munich" wasn't visible on the zoom level I choose, which I like since it doesn't interfere with the pin. If you find a zoom level that does this good, if not, no biggie.

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/352
Preview URL: https://satellytescommain21751-fixleafletzoomlevel.gtsb.io/contact/

**What's included?**
- The zoom level of the leaflet map is changed to 14

Bildschirmfoto 2022-02-08 um 11 32 49


src/components/pages/contact/leaflet/leaflet.tsx
```diff
@@ -60,8 +61,17 @@ const MapWrapper = styled.div`
...
+
   useEffect(() => {
     setIsBrowser(true);
+
+    const { innerWidth: width } = window;
+    if (width < 960) {
+      setZoomLevel(MAP_VIEW_ZOOM_MOBILE);
+    } else {
+      setZoomLevel(MAP_VIEW_ZOOM_DESKTOP);
+    }
```
georgiee: 1. using window will break SSR. @feedm3 correct?
2. This will not update when the browser window changes. Which is actually not a big problem. But what about using `useMedia` which we already use to fetch the ratio. You can use it do determine the size like a breakpoint with max-width etc.

Reference: 
https://github.com/satellytes/satellytes.com/blob/eae2810c723badda85ad797899d27853cb36c02a/src/components/content/heroes/blog-hero.tsx#L56

georgiee (changes requested):
Thank you for handling the design changes! This is almost good and already works. I have two small comments. Please check them before we can proceed 🙌

src/components/pages/contact/leaflet/leaflet.tsx
hofmann-felix: I shared the Preview URL with Eric and he said that it is should have the same zoom level on mobile and desktop now. I will remove the logic. Thanks anyway for your feedback @georgiee 🙏

feedm3: You can use the window object within `useEffect`. It only gets executed on the client.

`useMedia` would be the better solution in general.

georgiee (approved):
let's merge 🙌

PR #388, opened 2022-02-08
fix: typography issues

### LINKED ISSUE #353: Design: Typography
- [x] Bulletpoints: Please use "-" as bulletpoints, not circles currently live. In blog articles, please do not indent bulletpoint sections, they should be aligned left with the rest of the text, see https://cln.sh/LJmLIZ. I wrote a "-" plus two spaces to achieve this design.
- [x] Captions: Image captions are currently italic and black and large. Please check Figma and make them non-italic, grey and smaller.
- [x] Italic: Is it possible to disable italic text? We don't need it, some authors used it and I would like to eliminate it. Replace it with regular.
- [x] Formfield labels and errors: They are too small in the design, they were 12px. I made them a tiny bit larger to 13px now.
- [x] Textsize: On this page https://satellytes.com/services/ the bullet point texts have the right textsize (16px), but the text above it is too large (20px). Please make all fontsizes in this block 16px. ATTENTION: See next issue.
- [ ] Introtexts: The first textblocks on every page should be 20px. The editor should be able to achieve that. It's already working e.g. on the service page, but I would like to make it available also in blog articles in markdown. The idea is that the first paragraph is an intro to the upcoming page/article so it should have larger (20px) text. Let's discuss (maybe also with Georgios) what idea he has to make larger floating text available in blog articles.

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/353

**Whats's included?**
- Use `-` instead of `•` as bullet points on all pages:
  - [Service Page](https://satellytescommain21751-fixtextstyle.gtsb.io/services/) 
  - [Career Details](https://satellytescommain21751-fixtextstyle.gtsb.io/career/staff-software-engineer/)
  - [Blog post](https://satellytescommain21751-fixtextstyle.gtsb.io/blog/getting-started-gatsby-next-remix/)
- Remove the indent of ordered and unordered lists in [blog posts](https://satellytescommain21751-fixtextstyle.gtsb.io/blog/how-consistency-helps-you-to-optimize-gatsby-urls/)
- Render italic text as normal text in blog posts
- Adjust the form field error and label font style
  - [Contact Page](https://satellytescommain21751-fixtextstyle.gtsb.io/contact/)
  - [Career Details](https://satellytescommain21751-fixtextstyle.gtsb.io/career/staff-software-engineer/)
- Adjust the font size of the [service page](https://satellytescommain21751-fixtextstyle.gtsb.io/services/)

Eric has already reviewed the changes ✔️


src/components/forms/text-input/text-input.tsx
```diff
@@ -24,6 +24,8 @@ const Input = styled.input<{ hasError?: boolean }>`
 
 export const Label = styled.label`
   ${TextStyles.textXS}
+  font-size: 13px;
```
feedm3: `font-size` shouldn't be set manually, but with the `TextStyles` object. We don't have a text style with 13px, so let's just create a new one (you can copy&paste the `textXS` if that fits).

I just add this comment here, but the new text style should be used everywhere in this PR.

src/components/pages/service/support.tsx
```diff
@@ -69,5 +65,19 @@ export const ExpandableStyled = styled(Expandable)`
 export const UnorderedList = styled.ul`
   line-height: 150%;
   margin: 0;
-  padding-left: 1em;
+  padding-left: 0;
+  list-style-type: none;
+`;
+
+export const DashedListItem = styled.li`
```
feedm3: Not totally sure about this but: All `li` children's of the unordered list should always have the `-` right? I think it makes sense to put the styled then inside the `UnorderdList` component. 

Should look something like this (not sure about the correct syntax, but you will figure it our ;-)):
```
export const UnorderedList = styled.ul`
  ...
  li > {
    ... your styles
  }
`;
```

This way, we don't have to use this custom component for the list items and we are sure that the style is used on all our unurderd lists.

feedm3 (changes requested):
Text looks way better now, I like it :) 

I found some smaller technical topics, functionality is already great.

feedm3 (approved):
Looks good now, well done!

PR #389, opened 2022-02-09
feat: migrate to native gatsby typescript support

feedm3:
Gatsby is working on supporting Typescript out-of-the-box not only for pages and components but also for gatsby files `gatsby-config`, `gatsby-node`, `gatbsy-ssr` and `gatsby-browser`. 

The RFC can be found here: https://github.com/gatsbyjs/gatsby/discussions/34613
Deployment Preview: https://satellytescommain21751-featgatsbyts.gtsb.io/

### What's in this PR
- upgraded gatsby version to the latest version
- replaced `__dirname` with `path.resolve`, as `__dirname` is not supported by parcel (typescript parser used by Gatsby)
- all `.js` files migrated to `.ts` files
  - switched from `module.exports` to `export const`
  - found a bug where we passed the wrong props into `createFileNodeFromBuffer` (that didn't appear to make any problems)
  - added Gatsby types to all page components, `createPages` functions and configs

Expect for the local Personio source plugin everything is typed now, which is pretty cool 🥳 

### How to test
I didn't change any functionality, a smoke test on the deployment preview should be enough in that regard.


georgiee:
Awesome and thanks for giving this a try🙌

gatsby-config.ts
```diff
@@ -274,7 +272,7 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
-        path: `${__dirname}/src/assets/locales`,
+        path: `./src/assets/locales`,
```
chunksnbits: I infer that is because you encountered path-resolution issues after updating to latest gatsby-version?
In that case you can also use `process.cwd()` instead

gatsby/create-node/create-preview-cards.ts
```diff
@@ -46,16 +45,12 @@ const createPreviewCard = async (
   }
 };
 
-const createPreviewCards = async ({ node, ...rest }) => {
+export const createPreviewCards = async ({ node, ...rest }) => {
   if (node.internal.type === 'SyPersonioJob') {
-    await createPreviewCard(node.name, { node, ...rest });
+    await createPreviewCard(node.name, { node, ...rest } as any);
```
chunksnbits: Could this be easily be resolve by adding proper types to `export const createPreviewCards = async ({ node, ...rest })`? Or too much effort?

chunksnbits (approved):
Builds seems to work. And as far as I am aware no business changes. 

So 🚀 ✨

src/pages/imprint.tsx
```diff
@@ -1,31 +1,26 @@
...
   margin-top: 80px;
   opacity: 0.8;
 `;
 
-interface ImprintPageProps {
-  data: {
-    locales: LocalesQuery;
-    markdownRemark: {
-      htmlAst: string;
-    };
```
georgiee, line 23: nice how that is now covered by `PageProps`

georgiee (approved):
Nice, that looks great. Did not check it out yet though.
⚠️Gatsby Cloud is currently timed out, is this some hiccup ? 
Screenshot 2022-03-07 at 07 16 57

Otherwise I'm super happy to finally have TypeScript in those files 🙌

gatsby/create-node/create-preview-cards.ts
feedm3: Yep, got it working 👍

Issue #392, opened 2022-02-10
Design: Update colors


esreverse:
In Safari we can colorize the header. Please use: #3E61EE
<img width="525" alt="EricsShot 2022-02-10 10 54 28@2x" src="https://user-images.githubusercontent.com/6918281/153382540-1df4965c-4fc4-468a-af77-b46e0e3a68a5.png">

Please update share section colors (facebook, whatsapp, ... subscribe) with the new color: #3E61EE, and the headlines "Share article" and "Subscribe to blog" shouldn't be blue because they are not clickable, but darkblue #202840 instead:
<img width="442" alt="EricsShot 2022-02-10 10 54 12@2x" src="https://user-images.githubusercontent.com/6918281/153382481-7432f3ab-aee8-41bc-a9d8-43622296ce0f.png">

Formfields have wrong colors. Should be: #F7F8FA, and hover/active don't make the field darker, but use a 1px inside stroke with color: #202840
<img width="432" alt="EricsShot 2022-02-10 10 49 53@2x" src="https://user-images.githubusercontent.com/6918281/153381747-a8585fee-8a13-4d01-8e34-7805698e3b27.png">

georgiee:
The scope of the issue being "color" is good from design perspective but useless for a developer as it's three different topics. I thing the link colors are done already. Please give feedback to the form fields @esreverse @milljoniaer if it's missing to create a new issue for it.

~I will create new issue for the header color and close this issue now.~
The theme color is present already 👍

milljoniaer:
The color of the form fields will be fixed with my career form pr #398.
Shall we although create an issue for it?

PR #397, opened 2022-02-11
chore: set correct theme color

milljoniaer:
#### What is included?
- This changes the `theme-color` meta tag using our gatsby-plugin-manifest
- In that way safari changes its header/tab bar color like mentioned in #392 

##### Design:
image

##### Solution
image


feedm3 (approved):
LGTM 👍 

The Gatsby Build is stuck. I will force merge this into main, as this is a trivial change. Thanks :)

PR #401, opened 2022-02-11
feat: improve header colors

### LINKED ISSUE #400: Design: Adjust Header colors
The color (and also the hover-color) of the text and the icons in the header should match to the state of the header, so that they are always perfectly visible readable.

We have three different states:
- transparent header (no hero) : color: lila -> hover: dark blue
- aurora : color: white
- image : color: white

milljoniaer:
#### What is included?
This should set the colors of the header text and icons according to #400 

It closes #400


feedm3 (approved):
Technically fine for me 👍

milljoniaer:
Eric approved it also from the Designer perspective.
I'll merge it.

Issue #343, opened 2022-01-31
SEO indexing issues or just a lag?


georgiee:
When I search for "[staff engineer](https://www.google.com/search?hl=en&q=staff%20engineer%20satellytes)", "[How enterprises benefit from scoped npm registries satellytes](https://www.google.com/search?hl=en&q=How%20enterprises%20benefit%20from%20scoped%20npm%20registries%20satellytes)", or "[metaverse satellytes](https://www.google.com/search?q=metaverse+satellytes&hl=en)" I expected to see our content as it's available for a few days or even weeks.

I can find the older content though. Is this just an indexing lag? Can we improve that?

feedm3:
Just checked the Google search console and it looks like some pages are not indexed:

<img width="898" alt="Screenshot 2022-01-31 at 08 02 01" src="https://user-images.githubusercontent.com/7814926/151751699-55753e0a-b614-42b0-b0b8-55c077a0fb9c.png">

<img width="1168" alt="Screenshot 2022-01-31 at 07 59 15" src="https://user-images.githubusercontent.com/7814926/151751409-9a180afd-c4f2-4413-8831-ff3e1fa8097c.png">

Usually, Google should crawl them automatically, they are even in the `sitemap.xml`: https://satellytes.com/sitemap/sitemap-0.xml

I now triggered a crawling action on all of those 4 pages manually within the Google Search Console.

georgiee:
SEO is good right now. @feedm3 looked at it and the last time we did not have any issues after improving the semantic data as seen in the linked PR. This issue is done.

PR #413, opened 2022-02-14
feat: improve SEO with static date and more sitemap data

feedm3:
Some pages are not indexed by Google. We are not sure why, as they are technically fine. But there are some small things we can improve. Those things are:

- Render the blog post publication date fixed. The date was previously rendered relative (at least if it's less than 1 month). This could be bad for SEO and it results in a flickering message because the string is different when the build is older than a day. 
Screenshot 2022-02-14 at 15 46 11

- Add some meta tags to the sitemap. `priority` and `changefreq` are a bit esoteric, I just adopted the same values we use for other customers. The `lastmod` date is only used on blog posts.

Let's hope those 2 fixes improve our SEO score and make some of the missing pages accessible on Google.

Test URLS:
- Blog Post: https://satellytescommain21751-featimproveseo.gtsb.io/blog/cloudfront-cache-efficiency/
- Sitemap: https://satellytescommain21751-featimproveseo.gtsb.io/sitemap/sitemap-0.xml

Relates to #343

### DIFF
--- a/gatsby/create-pages/create-blog-posts.js
+++ b/gatsby/create-pages/create-blog-posts.js
@@ -15,6 +15,7 @@ const createBlogPosts = async ({ actions, reporter, graphql }) => {
         nodes {
           frontmatter {
             path
+            date
           }
         }
       }
@@ -31,7 +32,9 @@ const createBlogPosts = async ({ actions, reporter, graphql }) => {
     createPage({
       path: appendTrailingSlash(node.frontmatter.path),
       component: BLOG_POST_TEMPLATE_PATH,
-      context: {},
+      context: {
+        publicationDate: node.frontmatter.date,
+      },
     });
   });
 };
--- a/gatsby/create-pages/create-blog-posts.js
+++ b/gatsby/create-pages/create-blog-posts.js
@@ -33,6 +33,7 @@ const createBlogPosts = async ({ actions, reporter, graphql }) => {
       path: appendTrailingSlash(node.frontmatter.path),
       component: BLOG_POST_TEMPLATE_PATH,
       context: {
+        // used for the sitemap
         publicationDate: node.frontmatter.date,
       },
     });
--- a/gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap.js
+++ b/gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap.js
@@ -118,7 +118,7 @@ function gatsbyPluginSitemap({ allSitePage: { nodes: allPages } }) {
   const paths = allPages.map(({ path }) => path);
   const translationLookup = createPageTranslationLookup(paths);
 
-  return paths.map((path) => {
+  return allPages.map(({ path, pageContext }) => {
     const normalizedPath = normalizePath(path);
     const languageKey = extractLangKey(path);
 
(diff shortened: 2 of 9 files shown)


gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap.js
```diff
@@ -144,7 +151,15 @@ function serialize(sitemapItem) {
...
-    links,
+
+    // there is a lot of fuzz about changefreq and priority, some even say it
+    // doesn't have any effect at all. so let's use a very simple definition
+    changefreq: 'daily',
+    priority: HIGH_PRIO ? 1 : 0.8,
+
+    // publicationDate is optional and won't be rendered if not present
+    lastmod: sitemapItem.publicationDate,
+    links: alternateLinks,
```
georgiee: thanks 👍

gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap.js
```diff
@@ -129,13 +129,20 @@ function gatsbyPluginSitemap({ allSitePage: { nodes: allPages } }) {
...
     };
   });
 }
 
 function serialize(sitemapItem) {
+  const HIGH_PRIO =
+    sitemapItem.path.includes('/blog') ||
+    sitemapItem.path.includes('/career') ||
+    sitemapItem.path === '/' ||
+    sitemapItem.path === '/de';
```
georgiee: this is kind of self explaining but I feel somehow surprised by that because I get the reasoning only by some thinking. It's the update frequency mostly? Maybe a comment could provide that link why those are high prio ? Or use a name "FREQUENTLY_UPDATED" and use that variable for the priority? That would explain it very much to me.

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -43,10 +46,7 @@ export const BlogPostPage = ({ markdown, breadcrumb }: BlogPostPageProps) => {
     },
   };
 
-  const dateInWords = formatDistanceToNow(new Date(fm.date), {
-    locale: enGB,
-    addSuffix: true,
-  });
+  const dateInWords = dateFormatter(fm.date);
```
georgiee: that naming is wrong now. It's not in words anymore.

src/components/pages/blog/posts.tsx
```diff
@@ -20,7 +20,7 @@ interface PostsProps {
 }
 
 export const Posts = ({ posts }: PostsProps) => {
-  const dateFormatter = useLocaleFormat('dd. MMMM yyyy');
+  const dateFormatter = useLocaleDateFormatter();
```
georgiee: I see you put that into the helper so you don't have to repeat the formatting? I think the name `useLocaleDateFormatter` is not the best as it doesn't give me any ideas how it's formatted.

Something like that would tell a better story ?
```
useLocaleFormat(LONG_DATE');
```

gatsby/create-pages/create-blog-posts.js
```diff
@@ -31,7 +32,9 @@ const createBlogPosts = async ({ actions, reporter, graphql }) => {
     createPage({
       path: appendTrailingSlash(node.frontmatter.path),
       component: BLOG_POST_TEMPLATE_PATH,
-      context: {},
+      context: {
+        publicationDate: node.frontmatter.date,
```
georgiee, line 37: That's only fore the sitemap. For a brief moment I thought, why should I pass in the data as the page itself queries for all of that 👍

georgiee (approved):
Thanks for those changes. I can confirm that the changes are as described in the preview links. I have left some notes, but nothing that blocks. Up to you.

gatsby/create-pages/create-blog-posts.js
feedm3: I added a comment to make it more clear

gatsby/gatsby-plugin-sitemap/gatsby-plugin-sitemap.js
feedm3: Yep, `FREQUENTLY_UPDATED` describes it better, thanks for that 👍  I was first using it for `prioity`, but then decided to only use it for the frequency, hence the "outdated" name.

src/components/pages/blog/posts.tsx
feedm3: Good idea, reads better now :) 

```js
const dateFormatter = useLocaleFormat(LONG_DATE_FORMAT);
```

PR #416, opened 2022-02-15
chore: remove language switch from header

### LINKED ISSUE #351: Design: Change language switch
Georgios and I agreed on that we want to remove the language switch in the header and only leave it inside the navigation. There however fully written as "English" and "Deutsch" so that we won't have a problem using native dropdowns anymore. So please remove the custom dropdown (rectangle with border) and call the native ones like we do on the fcbayern.com website language switch.

https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=3655%3A25488

hofmann-felix:
Resolves [#416 ](https://github.com/satellytes/satellytes.com/issues/351)
Preview: https://satellytescommain21751-choreheaderlanguageswitch.gtsb.io/

What's included?
- The language switch (and its header color logic) is removed from the header
- The language switch in the navigation is displayed in the flyout and footer version
- The language switch is a native `select` again
- Since I worked in the navigation anyway, I adjusted the margin of the social links by 1px

Bildschirmfoto 2022-02-15 um 10 43 26

### DIFF
--- a/src/components/layout/header/header.tsx
+++ b/src/components/layout/header/header.tsx
@@ -7,7 +7,6 @@ import { CloseBurgerMenuIcon } from '../../legacy/icons/header-icons/close-burge
 import { NavigationFlyout } from './menu-flyout';
 import { Link } from '../../legacy/links/links';
 import { Swoosh } from '../../legacy/icons/header-icons/swoosh';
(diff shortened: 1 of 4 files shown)


feedm3 (approved):
LGTM 👍

PR #418, opened 2022-02-15
chore: provide teaser text for older blog posts

georgiee:
I added some missing teaser texts to our older blog posts. 

before:
Screenshot 2022-02-15 at 11 28 32

after:
Screenshot 2022-02-15 at 11 28 49


feedm3 (approved):
Nice! Also had the idea to fix it at some time 👍

PR #433, opened 2022-02-21
fix: add some redirects that are still in the google index

feedm3:
Checking the Google Search Console, I found those 2 links that are still in the index but show a 404 page. I created a redirect to the correct page.

The change is trivial. Feel free to click on `merge` once approved.


georgiee (approved):
Thanks, let's do this 🙌

Works:
+ https://satellytescommain21751-fixredirects35500.gtsb.io/office
+ https://satellytescommain21751-fixredirects35500.gtsb.io/de/blog/monorepo-codeowner-github-enterprise/

PR #444, opened 2022-02-24
feat: close menu on background click

### LINKED ISSUE #393: Design: Close menu
When the hamburger menu is open, it should not only be possible to close it with the little x in the top right corner, but also when clicking somewhere outside the purple navigation that comes in from the bottom. Please make a 30% black overlay behind the navigation over the complete width/height of the page and make the pink part clickable so you can also close the menu when clicking there.
EricsShot 2022-02-10 11 30 32@2x

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/393
Preview: https://satellytescommain21751-featclosemenubackground.gtsb.io/

**What's included?**
* The background of the flyout menu has 30% black now
* The flyout menu closes on background click 
* The flyout menu closes on `ESC` 

**Preview**

https://user-images.githubusercontent.com/57712895/155494577-791b895c-fc1d-47de-9dbc-1cfd3a67e19b.mov


src/components/layout/header/menu-flyout.tsx
```diff
@@ -40,18 +54,25 @@ interface NavigationFlyoutProp {
...
+          <FullHeightNavigation
+            translation={props.translation}
+            showLanguageSwitch={props.showLanguageSwitch}
+          />
+        </ScrollContainer>
+        {props.visible && <NoScrollBody />}
+      </FullscreenOverlay>
+      <BackgroundOverlay
+        visible={props.visible}
+        onClick={() => props.setIsNavigationVisible(false)}
```
georgiee, line 75: Ad-hoc request: While working on that, maybe you can provide also the `ESC` key to close it for some additional convenience?

hofmann-felix: Good idea, I will add it 👍

src/components/layout/header/header.tsx
```diff
@@ -161,6 +162,7 @@ const Header: React.FC<HeaderProps> = (props) => {
     <StyledHeader
       $lightTheme={Boolean(props.$lightTheme)}
       $transparent={isHeaderTransparent}
+      onKeyDown={(e) => e.keyCode === 27 && setIsNavigationVisible(false)}
```
hofmann-felix: I tried to place the keyDown listener inside the `NavigationFlyout` component, but then the `CloseBurgerMenuIcon` focuses first on ESC keypress (even with `tabIndex`)

src/components/layout/header/menu-flyout.tsx
```diff
@@ -20,6 +20,20 @@ const FullscreenOverlay = styled.div<{ visible: boolean }>`
   transition: bottom 0.3s ease-out, visibility 0.3s ease-out;
 `;
 
+const BackgroundOverlay = styled.div<{ visible: boolean }>`
+  position: fixed;
+  z-index: 900;
```
georgiee: You introduced `FLYOUT_Z_INDEX` but here you create another magic number. Maybe you want to hold that in a variable too ?

src/components/layout/header/header.tsx
georgiee: This works, thank you!

You  can also attach and detach a global event listener for `keyup' or `keydown` on `document`. That way it only listens when it's open. I fear some side effects here. If you feel confident let's merge.

georgiee: Please store the value `27` in a local variable `ESC_KEY_CODE` so it's speaking for itself, right now you don't know what's happening.

src/components/layout/header/menu-flyout.tsx
georgiee: I see the added feature, let's discuss in the appropriate place. Thanks!

georgiee (changes requested):
Thanks for your functional changes! I have a few remarks after that I'm sure we are ready to merge. 🤩

hofmann-felix:
@georgiee Thanks for your feedback 🙏 I have implemented it

src/components/layout/header/menu-flyout.tsx
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
 import styled, { createGlobalStyle } from 'styled-components';
-import { HEADER_HEIGHT } from './header';
 import Navigation from '../navigation/navigation';
 
+export const FLYOUT_Z_INDEX = 999;
+const BACKGROUND_Z_INDEX = 900;
+
```
georgiee, line 7: thanks for variables 👍

src/components/layout/header/header.tsx
```diff
@@ -4,12 +4,13 @@ import { up } from '../../support/breakpoint';
...
 import { BurgerMenu } from '../../legacy/icons/header-icons/burger-menu';
 import { CloseBurgerMenuIcon } from '../../legacy/icons/header-icons/close-burger-menu';
-import { NavigationFlyout } from './menu-flyout';
+import { FLYOUT_Z_INDEX, NavigationFlyout } from './menu-flyout';
 import { Link } from '../../legacy/links/links';
 import { Swoosh } from '../../legacy/icons/header-icons/swoosh';
 import { HEADER_HEIGHT_VALUE } from '../theme';
 
 export const HEADER_HEIGHT = `${HEADER_HEIGHT_VALUE}px`;
+const ESC_KEY_CODE = 27;
```
georgiee, line 13: 🙌

src/components/layout/header/header.tsx
georgiee: did you try a global approach or checked side effects of your current solution? Will it hijack the ESC key when in a form (in the content) or such ? The PR is fine as is now but I would like to hear some feedback from you.

src/components/layout/header/menu-flyout.tsx
georgiee: I saw the fix

georgiee (approved):
Let's merge this, works pretty good 👍 I left one comment regarding the global event listener, maybe you can reply. Thanks.

src/components/layout/header/header.tsx
hofmann-felix: I compared the forms with those from the current version and the ESC key has no side effects. I have not tried a global approach, do you think this would be better?

PR #488, opened 2022-03-14
feat: generate anchors to be able to link to specific sections

### LINKED ISSUE #459: Generate anchors in blog posts to be able to link to specific sections
### Expected outcome
As a user I want to be able to link to a specific headline in one of our blog posts. 

Example: I want to be able to link to the section "Compiling and bundling an Action" in [this blog post](https://satellytes.com/blog/boosting-developer-productivity-with-github-actions/).

### Proposed solution
I suggest to add `id` attributes (or add nested `<a id="...">` elements) in headlines and make them clickable like in [the HTML that GitHub generates from Markdown](https://github.com/satellytes/satellytes.com#component-development-with-storybook): 
image

hofmann-felix:
Resolves: https://github.com/satellytes/satellytes.com/issues/459
Preview URLs: 
* Blog Article: https://satellytescommain21751-featblogpostanchor.gtsb.io/blog/four-ways-to-improve-collaboration-in-your-team/#in-practice
* Services page: https://satellytescommain21751-featblogpostanchor.gtsb.io/services/#digital-platforms

### What's included?
* This PR adds a higher-order component, which generates and adds anchor links to headlines
* The new component is currently included in all blog articles and in the services page

### To Do
- [x] wait for feedback from the design team

### Preview 
Bildschirmfoto 2022-04-05 um 13 48 14


src/components/legacy/markdown/custom-components.tsx
```diff
@@ -115,6 +114,25 @@ const Figcaption = styled.figcaption`
...
+    }
+  });
+
+  return anchorId
+    .replace(/[^\w\s]/g, '')
+    .split(' ')
+    .join('-')
+    .toLowerCase();
+};
+
```
georgiee: this deserves a own file with a test. I also don't understand what's happening here. I see the effect that this ends at the level of the text content. You could access the text much easier with `node.textContent` and then transform it into kebab case while removing the special characters (if needed).

How does github do it? Is the text transformed into an anchor id ?

src/components/legacy/markdown/custom-components.tsx
```diff
@@ -131,19 +149,69 @@ const customSatellytesComponents = {
     return <SubTitle as={'h1'}>{props.children}</SubTitle>;
   },
   h2(props) {
-    return <SubTitle as={'h2'}>{props.children}</SubTitle>;
+    return (
+      <SubTitle as={'h2'}>
+        <a
```
georgiee: cool that works. What would you do regarding our fixed header though? Because the headlines slips under it when you click it. Do we need js for that or do we have other possibilities?

https://user-images.githubusercontent.com/1701755/158153162-06749665-639e-4fb6-b000-b158b771d6eb.mp4

src/components/legacy/markdown/custom-components.tsx
```diff
@@ -131,19 +149,69 @@ const customSatellytesComponents = {
...
+    );
   },
   h4(props) {
-    return <SmallTitle as={'h4'}>{props.children}</SmallTitle>;
+    return (
+      <SmallTitle as={'h4'}>
+        <a
+          href={'#' + generateAnchorId(props.children)}
+          id={generateAnchorId(props.children)}
+        >
```
georgiee: seeing all those anchors nested I think you should not do that. We will not need it everywhere. 
I think it would be better to offer it for some specific cases where we want it (like in our blog or in sections).

Maybe you can create a higher order component and wrap any given header? The anchor doesn't need to stay inside the headline as it's solely an anchor. Something like this:

```
// AnchorizeHOC will prepend an anchor, read the content of the title component to generate the anchor id.
const AnchoredTitle = AnchorizeHOC(TitleComponent)

<AnchoredTitle/>

```

What do you think? Maybe @feedm3 has some thoughts on that ?

georgiee (commented):
Thanks for the draft. See my comments.

src/components/legacy/markdown/custom-components.tsx
georgiee: See me clicking on the headlines how it slips under the header.

hofmann-felix: Thanks for the hint, I fixed it with `scroll-margin-top`

src/components/legacy/markdown/custom-components.tsx
hofmann-felix: I converted the current version into a higher-order component and it looks much better now. In addition, it can now be reused more easily, e.g. on the service page (https://satellytescommain21751-featblogpostanchor.gtsb.io/services/#digital-platforms)

src/components/legacy/markdown/custom-components.tsx
hofmann-felix: Yes, GitHub transfers the text into an id too. The general sense of this function are edge cases with headings that contain more than plain text, such as here:
Bildschirmfoto 2022-03-15 um 15 06 18
(https://satellytescommain21751-featblogpostanchor.gtsb.io/blog/boosting-developer-productivity-with-github-actions/#the-actionyml-file)

Thanks for the hint with `node.textContent`. I tried several times to implement the function with it, but I always had the same problem: If the page is already loaded and you click on the anchor, everything works as usual. But if you open a link with an anchor in a new tab, it doesn't scroll to the right place. Therefore the function is currently still iterating over the children.

milljoniaer: I dont know if this hepls, but for the copy component I have used our [only-text.ts](https://github.com/satellytes/satellytes.com/blob/main/src/components/support/only-text.ts) method. Couldn't that be helpful for those edge cases where there is more than normal text?

milljoniaer:
Cool feature, I like it very much 👍 

Regarding your open point of integration into the design, I personally like the way Github shows the share Tag only when hovering:

https://user-images.githubusercontent.com/78978542/161261668-14f0afe8-7cf1-44d7-98d6-d4c41d1dc63f.mov

Could that be a way to integrate the feature?

src/components/legacy/markdown/custom-components.tsx
hofmann-felix: @milljoniaer Thanks 🙏 I didn't know that we already have such a function

feedm3 (approved):
Looks very good, well done! 👍

PR #504, opened 2022-03-21
feat: add copy component

### LINKED ISSUE #136: Show visual feedback after copying a code block

milljoniaer:
### What is included?
This adds a Copy component which shows visual feedback for a few seconds once clicked and copies the passed text.

I already used it in the legacy Code Block Component, so it is displayed in the Blogposts.

https://user-images.githubusercontent.com/78978542/159286255-794cabcd-cf8c-4980-b2c3-c72612e36bba.mov

closes #136


src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import styled, { css } from 'styled-components';
+import { IconButton } from '../buttons/icon-button';
+import { Icon } from '../icon/icon';
+
+const Default_Clicked_Duration = 3000;
```
feedm3: Constants always upper-case: `DEFAULT_CLICKED_DURATION`.

Also not 100% happy with the name. It doesn't clearly indicate what it does and the unit is missing. 
Better could be: `SHOW_SUCCESS_NOTIFICATION_MS`

src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface CopyProps {
+  /**
+   * Duration, that the Checkmark will be shown once the button is clicked
+   * @default 3000ms
```
feedm3: This is super hard to maintain. Once you change the variable, the comment doesn't get updated automatically. It's probably better to just remove it. As the variable is also on top of the file it's already easy to spot the value.

src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ${(props) =>
+    props.clicked &&
+    css`
+      padding: 8px;
+      border-radius: 3px;
+      background-color: white;
+    `}
+`;
+
+export const Copy = ({ clickedDuration, text, className }: CopyProps) => {
```
feedm3: Maybe `Copy` is a bit too general, it can mean multiple things (could be text that gets displayed or a more technical copy of a component, etc.). I think `CopyButton` narrows it down to what it actually is.

src/assets/locales/de/translations.json
```diff
@@ -196,6 +196,7 @@
   "navigation.contact": "Kontakt",
   "blog.info": "Unsere Softwareingenieur:innen und Designer:innen sprechen über die neuesten Trends und das Know-how rund um Tech & Design. Insbesondere über unsere neuesten Erkenntnisse und Einsichten zur Entwicklung guter Software für das Web.",
   "blog.share": "Artikel teilen",
+  "blog.copied": "Kopiert",
```
georgiee, line 199: Thanks for thinking of the translations 🙌

feedm3 (approved):
Good job! I liked that you put the component now into a separate file, very good to read :) 

I have some smaller recommendations. Will already approve so you can merge once you added them.

src/components/legacy/icons/markdown-icons/copy.tsx
```diff
@@ -1,43 +0,0 @@
-import React from 'react';
```
georgiee, line 1: thanks for removing this icon file 🙌

src/components/ui/copy/copy.stories.mdx
```diff
@@ -0,0 +1,32 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Copy } from './copy';
+import { CodeBlock } from '../../legacy/markdown/code-block';
+
+<Meta title="Components/UI/Copy" component={Copy} />
+
+# Copy
```
georgiee: Thanks for the story ✨

https://deploy-preview-504--satellytes-website-storybook.netlify.app/?path=/docs/components-ui-copy--code&globals=locale:en;locales.en:English;locales.de:Deutsch

src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
...
+    navigator.clipboard.writeText(text);
+
+    setTimeout(
+      () => setClicked(false),
+      clickedDuration ?? Default_Clicked_Duration,
+    );
+  };
+
+  return (
+    <CopyWrapper id="copy-container" clicked={clicked} className={className}>
```
georgiee: I think you don't need the `id` here? It's not referenced at all?

src/components/ui/copy/copy.tsx
georgiee: constants should always be in all caps: `DEFAULT_CLICK_DURATION`

src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Icon } from '../icon/icon';
+
+const Default_Clicked_Duration = 3000;
+
+interface CopyProps {
+  /**
+   * Duration, that the Checkmark will be shown once the button is clicked
+   * @default 3000ms
+   */
+  clickedDuration?: number;
```
georgiee: This name doesn't resonate with me, what about `feedbackMessageTime`? The `click` is of course the action someone did before but describing the thing, the "feedback message" makes it more readable. You can also omit `copy` because that's the context of the entire component anyway.

src/components/legacy/markdown/code-block.tsx
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
-import { CopyIcon } from '../icons/markdown-icons/copy';
-import copy from 'copy-to-clipboard';
 import styled, { CSSProperties } from 'styled-components';
 import { onlyText } from '../../support/only-text';
+import { Copy } from '../../ui/copy/copy';
```
georgiee: Even though code-block is legacy code it's okay to add it. Let's tackle and eliminate the legacy folder soon afterwards instead.

src/components/ui/copy/copy.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const handleClick = () => {
+    setClicked(true);
+
+    navigator.clipboard.writeText(text);
+
+    setTimeout(
+      () => setClicked(false),
+      clickedDuration ?? Default_Clicked_Duration,
+    );
```
georgiee: so that's the magic that shows the clicked message. While we don't have actual side effects here I wonder if involving `useEffect` would be the more correct approach. Maybe someone from the crew has some opinion on that. 

I'm happy with the given solution 👍

georgiee (changes requested):
Nice, this looks great. I have provided a few comments and a few things to change (constant related, variable names) but nothing with big issues.

I will lock this PR until handled and then we can merge. Thank you 🙌

milljoniaer:
@georgiee @feedm3 Thank you for your recommendations, I will implement them now.

But I have a remaining question:
With the legacy component we copied the text with a "copy-to-clipboard" dependency.
I have now used navigator.clipboard.writeText(). Is that fine? And should I remove this dependency?

feedm3:
The native function is supported by all major browsers. I'm not sure why the dependency was used, maybe you can check if it has some important features that we want to have. Your implementation seems fine 

https://caniuse.com/mdn-api_clipboard_writetext

georgiee:
> @georgiee @feedm3 Thank you for your recommendations, I will implement them now.
> 
> But I have a remaining question: With the legacy component we copied the text with a "copy-to-clipboard" dependency. I have now used navigator.clipboard.writeText(). Is that fine? And should I remove this dependency?

As said and linked by @feedm3 I think removing the dependency is the way to go. I checked the git history and found this commit https://github.com/satellytes/satellytes.com/commit/8983e403f49334e4dafd99020cc973769ec2a7b2 where it was introduced. I guess there is no particular reason, maybe we did not think of the native functionality back then.

src/components/ui/copy/copy.tsx
milljoniaer: Ohs 🧹

src/components/ui/copy/copy.tsx
milljoniaer: 🧹

src/components/ui/copy/copy.tsx
milljoniaer: 👍

src/components/ui/copy/copy.tsx
milljoniaer: Ohps, leftover of one of my previous experiments

src/components/ui/copy/copy.tsx
milljoniaer: @feedm3 suggested showSuccessNotificationMS, I have now used that

src/components/legacy/markdown/code-block.tsx
milljoniaer: 👍

src/components/ui/copy/copy.tsx
milljoniaer: so do we leave it as it is, or has anyone thoughts on that?

milljoniaer:
Thank you for your comments, I pushed a new version and left some comments, please have another look

src/components/ui/copy/copy.tsx
georgiee: @feedm3 maybe? otherwise leave it as is.

src/components/ui/copy/copy.tsx
georgiee: thanks! Mind the `MS` postfix noting the unit format of milliseconds. That's a good point too 👍 I agree with the name.

package.json
```diff
@@ -39,7 +39,6 @@
     "axios": "^0.26.1",
     "babel-plugin-styled-components": "^2.0.2",
     "canvas": "^2.9.0",
-    "copy-to-clipboard": "^3.3.1",
```
georgiee, line 42: byeee 👋

georgiee (approved):
Thanks for your changes. Let's merge it 🙏

PR #398, opened 2022-02-11
Refactor: career form

milljoniaer:
### What is included? 
- refactored career form with `useController` (expect FileUpload) -> just like the contact form
- use of the new design form fields

### Thoughts on the PR
I tried to develop the career form just like Felix Hofmann the contact form. That is why I used the controlled approach. But in Addition to the Textfields and Checkbox, I had to deal with the File Dropper, which uses "react-dropzone".
I was not able to get "react-dropzone" to work together with the short and clean controlled approach of "react-hook-form" and had to use the manual approach for the File Dropper which  requires more code and even some ugly workarounds in the Form component to make the File Upload required and to validate the File Categories.

Because of the size of the PR, I will provide some thoughts of mine about some of the changed components.

#### Checkbox
--- Update ---
handled in #511 

#### File Dropper
--- Update ---
handled in #512 

#### Upload
I copied most of the code from the previous upload, but I had to change the way documents are uploaded, since the categories are passed differently than before by the new form.

### Example
[Staff Engineer](https://satellytescommain21751-chorehoverstateformcontact.gtsb.io/career/staff-software-engineer/)

### Screenshots
Normal:
image

Error:
image

Success:
image

### Questions:
Do we need an Uploading State? e.g. a loading bar
There is nothing like this defined in the design.

as far as i can see, it closes #503

PR #510, opened 2022-03-23
feat: remove legacy in Blog component

milljoniaer:
### What is included?
- This will remove the legacy components in a Blog component.
- Those legacy components are defined in the old career form files
- This PR will make the Career form PR #398 smaller

### It looks still fine:
image


fischeversenker (approved):
Thanks! The changes look perfectly fine and I trust your screenshot 🙌

PR #512, opened 2022-03-23
feat: make FileDropper work

milljoniaer:
### What is included?
This PR makes it possible to use the File Dropper together with react-hook-form.
It reduces the size of the career refactor pr #398 

### My thoughts on the PR

Originally I wanted to use the controlled approach of react-hook-form but because of the use of react-dropzone and reaact-hook-form I was not able to get this done.

Thats why I took care of the registration manually and set the Value of documents manually onDrop and onRemove.


src/components/forms/file-dropper/file-dropper.tsx
```diff
@@ -15,21 +10,28 @@ import { IllustrationType } from '../../ui/illustration/illustration-set';
...
+ * This Interface is needed, since the Props are declared in the new Career-fomr.tsx
+ * I will remove it once the new career form replaces the old one.
+ */
+interface FormDataProps {
+  documents: FileDropperType[];
+}
+export interface FileDropperType {
   file: File;
-  fileCategory: string | null;
+  fileCategory?: { value: string; label: string };
```
feedm3, line 31: As this is used at several places, you can put it into it's own type: 

```
interface FileCategory {
  value: string;
  label: string;
}
```

src/components/forms/file-dropper/file-dropper.tsx
```diff
@@ -45,11 +47,24 @@ interface FileDropperProps {
...
    * (Optional) Categories that the user can add to each uploaded file (e.g. 'CV' or 'Cover Letter')
    * */
-  fileCategories?: string[];
+  fileCategories?: { value: string; label: string }[];
   /**
    * (Optional) Pass in a valid illustration keyword to show the according illustration.
    */
   illustration?: IllustrationType;
+  /**
+   * (Optional) label, which is displayed above
```
feedm3: `(Optional)` is not needed. This information is already in the type itself (`...?`)

src/components/forms/file-dropper/file-dropper.tsx
```diff
@@ -45,11 +47,24 @@ interface FileDropperProps {
...
+   */
+  label?: string;
+  /**
```
feedm3, line 65: Adding some whitespace above makes it a bit more clear where the comment belongs to:

```ts
  /**
   * (Optional) Pass in a valid illustration keyword to show the according illustration.
   */
  illustration?: IllustrationType;

  /**
   * (Optional) label, which is displayed above
   */
  label?: string;

  /**
   * Use Controller Props, to get the form to work
   */
  setValue: UseFormSetValue<FormDataProps>;
```

feedm3 (approved):
The `file-dropper` is quite complex, even feels too complex to me. But it works and your are refactoring the whole career forms, so let's just ignore that fact and wait until you are finished.

Nevertheless, well done 👍  Only some small comments.

src/components/forms/file-dropper/file-dropper.tsx
milljoniaer: Looks better, thanks 👍

src/components/forms/file-dropper/file-dropper.tsx
milljoniaer: I am still learning to stop doing this 😅 

image

src/components/forms/file-dropper/file-dropper.tsx
milljoniaer: 👍

PR #513, opened 2022-03-23
fix: blog hero an mobile devices

milljoniaer:
### What is inclueded?

This PR removes is about the mobile Version of Blog Heros.
It removes the static height and uses the full size image not the squared one also for mobile.

Since the squared one is not needed anymore on blogs, I removed it.


georgiee:
Hm, it's not deploying?

ghost:
**satellytes.com-main**

:tada: Your build was successful! [See the Deploy preview here.](https://satellytescommain21751-fixmobileblogheros.gtsb.io)

## Build Details

[View the build logs here.](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/674be368-adf0-4ac5-8fae-f9a4493adc51/builds/4a21bbe5-e5f1-4413-9083-40c6cfd42cee/details)

:clock1: Build time: **2m**

## Performance

### Lighthouse report

| Metric | Score |
| ----- | ---- |
| Performance |   :large_orange_diamond: 62   |
| SEO |   :large_orange_diamond: 79   |

:link: [View full report](https://build-4a21bbe5-e5f1-4413-9083-40c6cfd42cee.gtsb.io/reports/lighthouse/index.html)

src/components/content/heroes/blog-hero.tsx
```diff
@@ -44,28 +43,16 @@ const Attribution = ({ attribution }) => (
  */
 export const BlogHero = ({
   image,
-  imageSquare,
   attribution,
   title,
   children,
   kicker,
 }: ImageHeroProps) => {
```
chunksnbits, line 50: Please check, if any more types need updating. I see that you deleted a lot of occurences of actual code, but could find only a single type that was deleted. Just to be sure.

src/components/content/heroes/support.tsx
```diff
@@ -10,9 +10,15 @@ import { rgba } from 'polished';
  * to spread the content over the full available width.
  * That way we don't need any position absolute & friends.
  * */
-export const HeroContainer = styled.div`
+export const HeroContainer = styled.div<{ isBlog?: boolean }>`
   display: grid;
-  height: 520px;
+
+  ${({ isBlog }) =>
```
chunksnbits: Imo this change could be used for all page-types. But tbh., this is more a question in the direction of @georgiee

chunksnbits (approved):
I like it 📸
Turned out better actually than I anticipated as the source image format allows some art-direction on how the mobile crop should turn out. 

👍 from my side, but I'd say better get an approval of someone more familiar with the project as well.

src/components/content/heroes/support.tsx
chunksnbits: If you want to keep this: is there a more generic way to name the attribute?
isBlog sounds a lot like "Sonderotto"
Maybe something around the lines of fixedHeight with a smart default?
Also think about the possibility of not using a negative to check the condition

feedm3: That was also the first thing that came to my mind. `isBlog` doesn't tell anything about the image. I like `fixedHeight` (and not making the check negative) 👍

feedm3 (approved):
"What Daniel said" 😄 

Looks very good, well done 👍

georgiee (approved):
Thanks for the changes 👍 Looks good!
I just pushed a refactoring of the name `isBlog` and picked `naturalHeight` instead of `fixedHeight` as suggested by Fabian as I thought it has a fewer impact. Then I recognized the HeroContainer is not used in many different places anyway. I hope the name is still okay.

here the proofs it's unchanged on other pages (like about us) while the blog images are always visible now. Love the change! Let's merge. (We will do it on behalf of Jonathan as he is out of office today.

Screen Shot 2022-03-24 at 10 01 57
Screen Shot 2022-03-24 at 10 01 47

Screen Shot 2022-03-24 at 10 01 29

PR #528, opened 2022-03-29
feat: add structured organization data to landingpage

### LINKED ISSUE #516: Provide structured data for our logo & company
I would like to tell Google about our Logo (and maybe something else is missing in that vein?)
Find everything you need here: https://developers.google.com/search/docs/advanced/structured-data/logo

It's about adding some structural data to the header to describe which logo we want to use. 

Example:
```
  <script type="application/ld+json">
    {
      "@context": "https://schema.org",
      "@type": "Organization",
      "url": "http://www.example.com",
      "logo": "http://www.example.com/images/logo.png"
    }
    </script>

```

Then it will show up like this:
Screenshot 2022-03-25 at 07 57 22

Google seems to use random images from Google Maps for Satellytes right now:
Screenshot 2022-03-25 at 07 58 17

milljoniaer:
### What is included?
This PR adds some information about Satellytes (including the logo) as structured Data to the landingpage.

### Things to think about:
- schema.org is totally fine with the provided data 
image

- But Google only recognizing the [logo](https://search.google.com/test/rich-results/result?id=ASbOLHJujAwYJUkx3rmm7w) but not the [rest of the data](https://search.google.com/test/rich-results/result/r%2Flogos?id=ASbOLHJujAwYJUkx3rmm7w)
- shall we then better leave out the other data to fit Googles Expectations?
- According to Google, It is also possible to change the other data in the knowledge panel via claiming ownership of the card. (it seems that eric is already owner of it, and would be able to change that)

--- update ---
According to https://www.schemaapp.com/schema-markup/how-schema-markup-helps-you-gain-or-enhance-a-google-knowledge-panel/ it seems fine to write all this information into the structured data.

What do you think about all this?

This closes #516


src/components/pages/landingpage/structured-organization-data.tsx
```diff
@@ -0,0 +1,39 @@
...
+    address: {
+      '@type': 'PostalAddress',
+      streetAddress: 'Sendlinger Straße 52',
+      addressLocality: 'München',
+      addressRegion: 'Bayern',
+      postalCode: '80331',
+      addressCountry: 'DE',
+    },
+    email: 'info@satellytes.com',
+    url: 'https://www.satellytes.com',
```
feedm3: The default URL is without the `www.`

feedm3 (approved):
LGTM 👍

src/components/pages/landingpage/structured-organization-data.tsx
milljoniaer: 👍

Issue #506, opened 2022-03-22
Provide structured data for our Jobs


georgiee:
We currently have https://satellytes.jobs.personio.de/ still activated but actually we don't want to use that second channel of jobs coming in because it's rendering "meta data" (or slug data is displayed as a section, [example](https://satellytes.jobs.personio.de/job/338504?display=en#apply)) we had to add to the description plus the form is not rendering as we like.

What hinders us from disabling it is the potential bad effects on SEO, because they have integrate Google Jobs by providing structured data while we ignored that topic so far.

I found this documentation about structured data:
https://developers.google.com/search/docs/advanced/structured-data/job-posting

You can find on any personio job page generated for us like [here](https://satellytes.jobs.personio.de/job/338504?display=en). The following data is located before the closing body tag as a script tag. I think we should do that too and then we can disable the personio career page entirely 🙌

Thoughts where I need support from the one who takes this issue:
+ Is this enough?
+ By reading the documentation, can we do even more?
+ Where can we check if it's correct?
+ How does this relate to our sitemap

```
<script type="application/ld+json">
[...]

PR #529, opened 2022-03-29
feat: add utm_source channel mapping for personio

milljoniaer:
### What is included?

This PR allows us to map certain `utm_source` values to hardcoded personio Channels.
This way we can "track" where the applicants came from.

In this PR the first UTM_Source for Google Jobs is added.

This covers a suggestion made in #506


src/components/pages/career-details/career-details.tsx
```diff
@@ -16,6 +16,10 @@ interface CareerDetailsProps {
 
 const DEFAULT_CHANNEL_ID = '329206';
 
+const UTM_CHANNEL_MAP = new Map<string | null, string | null>([
```
feedm3: Looks overly complex. I would recommend to keep it a simple object:

```
const UTM_CHANNEL_MAP = {
  google_jobs_apply: '713571',
};
```

Also, add a comment above why we need to have this map.

src/components/pages/career-details/career-details.tsx
```diff
@@ -60,7 +65,9 @@ export const CareerDetails = ({
 
       <Form
         company_id="41230"
-        recruiting_channel_id={personioChannel ?? DEFAULT_CHANNEL_ID}
+        recruiting_channel_id={
+          utmChannel ?? personioChannel ?? DEFAULT_CHANNEL_ID
```
feedm3: To keep logic out of the JSX, I would also put this check into its own variable. This way, you have all related code to the `recruitingChannelId` in one place.

feedm3 (approved):
LGTM 👍 

I just have some smaller comments about refactoring, so I will already approve this PR.

src/components/pages/career-details/career-details.tsx
milljoniaer: Making it less complex seems cool, but if I make the map an Object I am not able to do this anymore:
image

I liked that I dont have to think about the `utm_source` value and can just take it as key for my map to get the correct channel_id. (even if this whole thing makes only sense with more channels than only google jobs)

Is there a less complex way to do this or shall I just leave out the whole mapping thing and something like:

```
googleChannel = searchParams.get("utm_source") === "google_jobs_apply" ? GOOGLE_CHANNEL : undefinded;
```

src/components/pages/career-details/career-details.tsx
milljoniaer: 👍

src/components/pages/career-details/career-details.tsx
feedm3: You can access an object with the array syntax (not sure if this is the official name 😄):

```
const utmChannel = UTM_CHANNEL_MAP[searchParams.get('utm_source')]
```

georgiee: While you have a point with the types @milljoniaer I think we should go the much simpler way as outlined by @feedm3. We don't benefit from a map in this context and it introduces a lot of type noise. Make it an object and access it with the `bracket notation` (I had to search for it too @feedm3  😂, [reference](https://dev.to/capscode/dot-and-bracket-notation-in-javascript-object-12ij))

src/components/pages/career-details/career-details.tsx
```diff
@@ -41,6 +45,9 @@ export const CareerDetails = ({
   const location = useLocation();
   const searchParams = new URLSearchParams(location.search);
   const personioChannel = searchParams.get('_pc');
+  const utmChannel = UTM_CHANNEL_MAP.get(searchParams.get('utm_source'));
+
+  const recruitingChannel = utmChannel ?? personioChannel ?? DEFAULT_CHANNEL_ID;
```
georgiee, line 50: I like, that's a clear order  👍

georgiee (approved):
Lovely change 🙌 Let's change the map to an object.
Functionality is good already so I approve too 👍

src/components/pages/career-details/career-details.tsx
milljoniaer: Thank you for the bracket notation tipp, I did not knew about this, and this way it feels convenient again 👍

PR #530, opened 2022-03-29
fix: remove wrong job name

milljoniaer:
### What is included? 
This should be the identifier of the Job Posting and not of the hiring Organization, therefore I removed the "Satellytes Digital Consulting GmbH"

Since the name is optional and Personio does also not provide a name in their structured data, I suggest to leave it out.


ghost:
**satellytes.com-main**

:tada: Your build was successful! [See the Deploy preview here.](https://satellytescommain21751-fixremovejobentitysname.gtsb.io)

## Build Details

[View the build logs here.](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/674be368-adf0-4ac5-8fae-f9a4493adc51/builds/64a9a0a7-2bb3-4108-bc97-79f49d27c9b1/details)

:clock1: Build time: **3m**

## Performance

### Lighthouse report

| Metric | Score |
| ----- | ---- |
| Performance |   :large_orange_diamond: 63   |
| SEO |   :large_orange_diamond: 79   |

:link: [View full report](https://build-64a9a0a7-2bb3-4108-bc97-79f49d27c9b1.gtsb.io/reports/lighthouse/index.html)

feedm3 (approved):

georgiee (commented):
Thanks, that makes sense although the [documentation](https://developers.google.com/search/docs/advanced/structured-data/job-posting) is really not clear about that fact. Did the way Personio did it convince you? Because I think they are not the best source of truth to be honest.

Could we gather yet another example to be more sure about this and the effects of the name? If it's not Satellytes what could you write there? The name of the job itself? Or is it a field to describe the identifier like "Job ID?

Let's take a quick second tour and then let's merge. I will put my feedback as comment here. So neither a blocker nor an approval. I see @feedm3 approved so you can take that to merge once you are more clear about it.

Screenshot 2022-03-30 at 10 48 58

milljoniaer:
You are right, googles documentation says, that the identifier should be the companys "unique identifier".

But If we keep the name, we should not use the job id as value, should we?
Since the unique identifier should then be satellytes not the job postings identifier

georgiee:
@milljoniaer 
No I think you misinterpreted my feedback/screenshot. That's definitely scoped to the specific job. The purpose is to give Google an ID (unique identifier) to differentiate it from other job posting with the same names and such.

The only thing I wonder is about the name and that's where Google is not clear. Should we write "Satellytes" or "Satellytes Frontend Job Id" or what is expected here? It's not the title of the job because that's covered in some other field. It probably boils down to "If we don't give it a name, it gets internally the name `ID` and if we chose to name it it's named like that, e.g. `Sagellytes Job Id` or `Satellytes/Personio Job Id` or so. I don't even know where it would surface 😄 Probably when Google communicates with us in some admin area.

Long story short, let's merge. Removing is good and we don't know and need any special name here.
Thanks for your thoughts on this.

PR #531, opened 2022-03-29
feat: add medium zoom

milljoniaer:
### What is included? 
This PR adds the `react-image-medium-zoom` to provide a lightbox for the images in the blog posts and also in the Image component.

### How to verify:
- click on some images on https://satellytescommain21751-feataddmediumzoom.gtsb.io/blog/we-work-remotely/
- or https://satellytescommain21751-feataddmediumzoom.gtsb.io/about-us/

### Question:
On large screens (or when the images are really small, like the team images), the image is not sharp in the zoomed state, how do we want do deal with that?

### DIFF
--- a/src/components/pages/blog-post/blog-post.tsx
+++ b/src/components/pages/blog-post/blog-post.tsx
@@ -17,6 +17,8 @@ import { MarkdownAst } from '../../legacy/markdown/markdown-ast';
 import { up } from '../../support/breakpoint';
 import FollowPanel from './follow-panel';
 import SharePanel from './share-panel';
+import Zoom from 'react-medium-image-zoom';
+import 'react-medium-image-zoom/dist/styles.css';
 
 interface BlogPostPageProps {
   blogPost: BlogArticleQueryData;
@@ -95,18 +97,12 @@ const contentfulRenderOptions: Options = {
         case 'image': {
           return customComponents.figure({
             children: [
-              customComponents.a({
-                key: 'link',
-                children: (
-                  <GatsbyImage
-                    image={node.data.target.gatsbyImageData}
-                    alt={node.data.target.description}
-                  />
-                ),
-                href: node.data.target.file.url,
-                target: '_blank',
-                rel: 'nofollow noopener noreferrer',
-              }),
+              <Zoom key={node.data.target.file.url}>
+                <GatsbyImage
+                  image={node.data.target.gatsbyImageData}
+                  alt={node.data.target.description}
+                />
+              </Zoom>,
               customComponents.figcaption({
                 key: 'figcaption',
                 children: node.data.target.description,
--- a/src/components/ui/image/image.tsx
+++ b/src/components/ui/image/image.tsx
@@ -4,8 +4,6 @@ import styled, { css } from 'styled-components';
 import { up } from '../../support/breakpoint';
 import { theme } from '../../layout/theme';
 import { TextStyles } from '../../typography';
-import Zoom from 'react-medium-image-zoom';
-import 'react-medium-image-zoom/dist/styles.css';
 
 const mapTextAlignToFlex = (textAlign: 'right' | 'left' | 'bottom') => {
   if (textAlign === 'right') {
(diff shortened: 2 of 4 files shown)


src/components/ui/image/image.tsx
```diff
@@ -96,7 +98,7 @@ export const Image = ({
   return (
     <BlockWrapper textPlacement={textAlign}>
       <ImageWrapper>
-        {children}
+        <Zoom>{children}</Zoom>
```
georgiee: I think we should not enable zoom for ANY image put only optional. So I'm thinking of a higher order component that wraps the children according if necessary or a property `zoomable` 

Something like `ZoomImage = ({children) => {  <Image><Zoom>{children}</Zoom></Image>}` or `<Image zoomable/>`

feedm3:
Thanks for the PR! It's a cool feature 🙂 

My personal preference: I would only like to have this feature on images in blog posts. Having it on other parts of the website would only be good if it is for specific pictures. For example on the about page I like the zoom for the office pictures. But as much as I like it for office images, I don't like it for portraits. As we cannot control this behaviour as long as we have our content in markdown and not Contentful, I think it would be best to disable it completely on these pages. Blog posts is good enough for now.

Regarding the image quality on zoomed images: It looks like the author is working on it for Version 5 https://github.com/rpearce/image-zoom/issues/166.

georgiee (changes requested):
Hey that looks easy 🙌 Unfortunately that's a big change and makes every image zoomable which brings the issues outlined that images which are too small will still zoom.

When I look at an alternative implementation like https://www.gatsbyjs.com/plugins/gatsby-remark-images-zoom/ and the udnerlying sources https://github.com/premieroctet/gatsby-remark-images-zoom it's quite a simple implementation we could integrate directly the low level `medium-zoom` [package](https://github.com/francoischalifour/medium-zoom)  and maybe enforce or even query for high-res images?

Can you explore a little bit more?

milljoniaer:
Thank you for your comments. What do you think about using this feature only in the blog and adjusting the image resolution in the query, so that they are still sharp (or at least sharper) even on larger displays and zoomed?

Or shall I checkout the other implementation and query another image with higher resolution for the zoom?

milljoniaer:
As the main problem with the current implementation is, that images are not 100% sharp on large screens, I agree with @feedm3 and removed the Zoom from the content pages like about-us.

In compliance with @feedm3 I suggest to use Zoom only in blog posts (the pictures are relatively large and fine on zoom too) and wait for the v5 of "react-medium-image-zoom" where it will be possible to pass a zoomed image src.

What do you think @georgiee ?

src/components/ui/image/image.tsx
milljoniaer: 🧹

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -95,18 +97,12 @@ const contentfulRenderOptions: Options = {
...
-                  <GatsbyImage
-                    image={node.data.target.gatsbyImageData}
-                    alt={node.data.target.description}
-                  />
-                ),
-                href: node.data.target.file.url,
-                target: '_blank',
-                rel: 'nofollow noopener noreferrer',
-              }),
+              <Zoom key={node.data.target.file.url}>
```
georgiee, line 100: I think that should do it for blog posts 👍
 @dschu-lab as you have worked on that mapping recently, what's your opinion on that? Good to add the zoom image entry here for every image in the blog post by default?

georgiee (approved):
I fully agree with the blog scope and I see that you did that exactly already. Let's merge this 🙌 


---

I also read about the upcoming `v5` support in the linked [issue](https://github.com/rpearce/image-zoom/issues/166). 

You can see a glimpse of the progress in the linked [PR](https://github.com/rpearce/image-zoom/pull/244) which points to the branch [fop](https://github.com/rpearce/image-zoom/tree/fop).

By the way:
If anyone think this is a nice project and you always wanted to participate (privately) in a open source project, the scope & complexity looks graspable and the author seems to be working alone the challenge right now. Giving feedback to the current implementation could be a simple starter in that case  👍

src/components/pages/blog-post/blog-post.tsx
JulianAnthes: I like it! Let's get this merged to main. 🙌

PR #532, opened 2022-03-30
feat(blog): append footnotes after rich text

### LINKED ISSUE #527: Blog: Creating footnotes is not possible at the moment
Due to the switch to Contentful, we lost the ability to create footnotes with links on the same page.

> "It looks like everything we need is in the rich text data (I read this https://destroytoday.com/blog/footnotes"
@georgiee

JulianAnthes:
This PR enables us to place footnotes in rich texts.

## What's changing?

When footnotes are placed inside rich text, the nodes are filtered out and then rendered separately below the content.
There will be a link to the footnote below the content created. The footnote itself contains a link back to its reference.

### Our **Footnote Content Model**
Bildschirmfoto 2022-04-01 um 09 47 50

- The `content` field is optional and can contain text that gets highlighted
- The `note` field contains the actual footnote information

### How it looks when placed inside the rich text field
Bildschirmfoto 2022-04-01 um 09 48 59

Footnotes are placed as inline entities inside a rich text field.

### Editing a footnote
Bildschirmfoto 2022-04-01 um 09 49 11
The fields have a hint text to help the author.

### The rendered content
Bildschirmfoto 2022-04-01 um 09 49 29
Here is how the rendered content looks like. When footnotes exist, the list as well as a `hr` is appended below the content.

## Good to know
The GraphQL query is failing in the Gatsby Cloud because the content is not published yet.
Background: The Contentful plugin will only generate classes for content models that contain data. Since there is no data yet (not published), the entity is skipped which is causing the error.

We need to publish at least one article which contains a `ContentfulFootnote` before the Gatsby Cloud build can succeed.

### DIFF
--- a/src/components/layout/use-anchor-tag-scrolling.ts
+++ b/src/components/layout/use-anchor-tag-scrolling.ts
@@ -7,18 +7,25 @@ export const useAnchorTagScrolling = (): void => {
       const target = document.querySelector(
         `a[href*='${window.location.hash}']`,
       );
-      if (target) {
-        const scrollTop =
-          target.getBoundingClientRect().top +
-          window.pageYOffset -
-          // we need to scroll past the header and a little offset
-          (Number.parseInt(HEADER_HEIGHT, 10) + 16);
-
-        window.scrollTo({
-          top: scrollTop,
-          behavior: 'smooth',
-        });
-      }
+      scrollToTarget(target);
     }
   }, []);
+};
+
+export const scrollToTarget = (target: Element | null) => {
+  if (!target) {
+    console.error(`Expected Element, but got: ${target}`);
+    return;
+  }
+
+  const scrollTop =
+    target.getBoundingClientRect().top +
+    window.pageYOffset -
+    // we need to scroll past the header and a little offset
+    (Number.parseInt(HEADER_HEIGHT, 10) + 16);
+
+  window.scrollTo({
+    top: scrollTop,
+    behavior: 'smooth',
+  });
 };
--- a/src/components/pages/blog-post/blog-post.tsx
+++ b/src/components/pages/blog-post/blog-post.tsx
@@ -73,12 +73,10 @@ const customContentfulRenderer = (
   /**
    * Filter out footnotes from content references
    */
-  const footnotes: ContentfulFootnoteReference[] = data.references
-    .filter(
-      (reference) =>
-        reference.__typename === ContentfulCustomModel.CONTENTFUL_FOOTNOTE,
-    )
-    .map((reference) => reference as ContentfulFootnoteReference);
+  const footnotes: ContentfulFootnoteReference[] = data.references.filter(
+    (reference) =>
+      reference.__typename === ContentfulCustomModel.CONTENTFUL_FOOTNOTE,
+  ) as ContentfulFootnoteReference[];
 
   /**
    * Create an object that contains information we'll use later
--- a/src/components/pages/blog-post/blog-post.tsx
+++ b/src/components/pages/blog-post/blog-post.tsx
@@ -284,7 +284,12 @@ const customContentfulRenderer = (
   return (
     <div>
       {renderRichText(data, contentfulRenderOptions)}
-      {footNoteElements}
+      {footnotes.length > 0 && (
+        <>
+          <hr />
+          {footNoteElements}
+        </>
+      )}
     </div>
   );
 };
--- a/src/components/pages/blog-post/blog-post.tsx
+++ b/src/components/pages/blog-post/blog-post.tsx
@@ -35,6 +35,11 @@ enum ContentfulCustomModel {
   CONTENTFUL_CODE_BLOCK = 'ContentfulCodeBlock',
   CONTENTFUL_BLOG_POST_COLLAPSIBLE = 'ContentfulBlogPostCollapsible',
 }
+
+enum EmbeddedAssetType {
+  IMAGE = 'image',
+  VIDEO = 'video',
+}
 interface FootnoteReference {
   [key: string]: {
     index: number;
@@ -194,7 +199,10 @@ const customContentfulRenderer = (
       [BLOCKS.EMBEDDED_ASSET]: (node) => {
         const { contentType, url } = node.data.target.file;
         switch (contentType.split('/')[0]) {
-          case 'image': {
+          /**
+           * Embedded images
+           */
+          case EmbeddedAssetType.IMAGE: {
             return customComponents.figure({
               children: [
                 <Zoom key={node.data.target.file.url}>
@@ -210,7 +218,10 @@ const customContentfulRenderer = (
               ],
             });
           }
-          case 'video': {
+          /**
+           * Embedded videos
+           */
+          case EmbeddedAssetType.VIDEO: {
             return customComponents.figure({
               children: [
                 <video
--- a/src/components/pages/blog-post/blog-post.tsx
+++ b/src/components/pages/blog-post/blog-post.tsx
@@ -2,8 +2,12 @@ import { Options } from '@contentful/rich-text-react-renderer';
 import { BLOCKS, INLINES, MARKS } from '@contentful/rich-text-types';
 import { GatsbyImage } from 'gatsby-plugin-image';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
-import { renderRichText } from 'gatsby-source-contentful/rich-text';
-import React from 'react';
+import {
+  ContentfulRichTextGatsbyReference,
+  renderRichText,
+  RenderRichTextData,
+} from 'gatsby-source-contentful/rich-text';
+import React, { MouseEvent } from 'react';
 import styled from 'styled-components';
 import { BlogArticleQueryData } from '../../../templates/blog-post';
 import { BreadcrumbEntry } from '../../../types';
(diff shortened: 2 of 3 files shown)


JulianAnthes:
GraphQL query is failing in the Gatsby Cloud because the content is not published yet.
Background: The Contentful plugin will only generate classes for content models that contain data. Since there is no data yet (not published), the entity is skipped which is causing the error.

src/components/layout/use-anchor-tag-scrolling.ts
```diff
@@ -7,18 +7,25 @@ export const useAnchorTagScrolling = (): void => {
...
-          top: scrollTop,
-          behavior: 'smooth',
-        });
-      }
+      scrollToTarget(target);
     }
   }, []);
 };
+
+export const scrollToTarget = (target: Element | null) => {
```
georgiee, line 15: I see how you need access to that part of the functionality. Good to split and export it then.

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -12,12 +16,31 @@ import { BlogHero } from '../../content/heroes/blog-hero';
...
 import { up } from '../../support/breakpoint';
 import FollowPanel from './follow-panel';
 import SharePanel from './share-panel';
 
+interface ContentfulFootnoteReference
+  extends ContentfulRichTextGatsbyReference {
+  note: any; // TODO: Improve typing here
+}
+
+enum ContentfulCustomModel {
```
georgiee, line 33: Good idea to collect those identifiers here.
I would love to have embed asset enums for image, video etc, just for the sake of aligning our approaches. Or do those constants have some enum in the contentful package itself

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -33,114 +56,246 @@ const PanelContainer = styled.div`
...
+/**
+ * Custom Contentful Rich Text Renderer
+ *
+ * Will automatically filter out ContentfulFootnote elements and appends an
+ * ordered list of footnotes after the rendered rich text.
+ *
+ * @param data rich text data
+ * @returns rendered elements
+ */
+const customContentfulRenderer = (
```
georgiee, line 75: Great refactoring to allow for the footnotes to be included. I think having an explicit renderer for contentful it's time to put it into a separate file and add a few tests to ensure the matching data is output (not in this PR but as some follow up or delegate to a working student for the learning effect)

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -180,7 +335,7 @@ export const BlogPostPage = ({ blogPost, breadcrumb }: BlogPostPageProps) => {
       breadcrumb={breadcrumb}
     >
       <BlogHeader headline={blogPost.title} byline={heroByLine} />
-      <div>{renderRichText(blogPost.content, contentfulRenderOptions)}</div>
+      {customContentfulRenderer(blogPost.content)}
```
georgiee, line 338: That looks splendid now ✨

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -33,114 +56,246 @@ const PanelContainer = styled.div`
...
+                    e.preventDefault();
+                    const target = document.querySelector(
+                      `[id='${reference?.anchor}']`,
+                    );
+                    history.pushState({}, '', `#${reference?.anchor}`);
+                    scrollToTarget(target);
+                  },
+                })}
+              </>
+            );
```
georgiee: Great, simple to follow what you are doing here 👍

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -33,114 +56,246 @@ const PanelContainer = styled.div`
...
+  /**
+   * Create an object that contains information we'll use later
+   */
+  const footnoteReferences: FootnoteReference = {};
+  footnotes.map(
+    (footnote, index) =>
+      (footnoteReferences[footnote.contentful_id] = {
+        index: index + 1,
+        anchor: `footnote-${index + 1}`,
+        referenceAnchor: `footnote-${index + 1}-reference`,
```
georgiee: So first collect all footnote references from the given data to see what we have. Then build the indexed list to have unique links to output them inside the richtext and output the actual note in the bottom.

Looks like I can follow here 🙌

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -33,114 +56,246 @@ const PanelContainer = styled.div`
...
+) => {
+  /**
+   * Filter out footnotes from content references
+   */
+  const footnotes: ContentfulFootnoteReference[] = data.references
+    .filter(
+      (reference) =>
+        reference.__typename === ContentfulCustomModel.CONTENTFUL_FOOTNOTE,
+    )
+    .map((reference) => reference as ContentfulFootnoteReference);
```
georgiee: You map just for the purpose of getting the types right? Can't you just type case to `ContentfulFootnoteReference[]` instead? The last mapping operation looks useless from a functional point ?

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -33,114 +56,246 @@ const PanelContainer = styled.div`
...
-    [BLOCKS.HEADING_1]: (props, children) => customComponents.h1({ children }),
-    [BLOCKS.HEADING_2]: (props, children) => customComponents.h2({ children }),
-    [BLOCKS.HEADING_3]: (props, children) => customComponents.h3({ children }),
-    [BLOCKS.HEADING_4]: (props, children) => customComponents.h4({ children }),
-    [BLOCKS.HEADING_5]: (props, children) => customComponents.h5({ children }),
-    [BLOCKS.HEADING_6]: (props, children) => customComponents.h6({ children }),
-    [BLOCKS.HEADING_6]: (props, children) => customComponents.h6({ children }),
-    [BLOCKS.OL_LIST]: (props, children) => customComponents.ol({ children }),
-    [BLOCKS.UL_LIST]: (props, children) => customComponents.ul({ children }),
-    [BLOCKS.QUOTE]: (props, children) =>
```
georgiee, line 59: The diff got mixed up and list a lot of removals and additions, but I can easily follow looking directly into the file 👍

src/templates/blog-post.tsx
```diff
@@ -101,6 +101,19 @@ export const BlogPostPageQuery = graphql`
...
+            contentful_id
+            __typename
+            content {
+              content
+            }
+            note {
+              childMarkdownRemark {
+                htmlAst
+              }
+            }
```
georgiee: Welcome `content` and `note` from the content model enters the stage here 🙌

georgiee (changes requested):
Thanks for that. That's the approach outlined in the linked [reference](https://destroytoday.com/blog/footnotes) isn't it ? Looks pretty cool👍 I always thought the footnotes are somehow recognized in the rich text — I guess I was wrong 😄 Because the inline entry makes absolutely sense here from a data structure point of view.

Do we have or need `content`? I don't think that I have seen this in books. Because the reference is usually made for an entire paragraph or similar never on individual words. I also don't think that we have a style for that.

Should we drop `content` from the beginning? 

Otherwise I would be happy to see the rendered result. I think we can simply publish the content, can't we ? It should be ignored anyway ?

I block the PR until we discuss the content issue and I've seen the deployed preview. Great job with that so far 💪

JulianAnthes:
> Do we have or need `content`? I don't think that I have seen this in books. Because the reference is usually made for an entire paragraph or similar never on individual words. I also don't think that we have a style for that.

I guess you're right, thought it would be a nice addition, first but don't think it is necessary. 😄 
Just checked out [Wikipedia](https://en.wikipedia.org/wiki/Universe#cite_note-spacetime-10) as well and noticed that active footnotes have a background color. Might be a nice addition in the future. 🚀 
 
> Should we drop `content` from the beginning?

I'm voting for dropping `content`. 👍 

> Otherwise I would be happy to see the rendered result. I think we can simply publish the content, can't we ? It should be ignored anyway ?

I have published a `ContentfulFootnote` which is not referenced by any blog post. But this should trick GraphQL to create the missing classes. So next time we push something to the branch, a preview deployment should run through.

> I block the PR until we discuss the content issue and I've seen the deployed preview. Great job with that so far 💪

Thanks a lot! 🙏

src/components/pages/blog-post/blog-post.tsx
JulianAnthes: Oh...Absolutely! 😅

src/components/pages/blog-post/blog-post.tsx
JulianAnthes: We are splitting the files `contentType` (mime type), Contentful does not provide any enums for this, let's create our own.

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -20,6 +25,29 @@ import SharePanel from './share-panel';
...
+  note: any; // TODO: Improve typing here
+}
+
+enum ContentfulCustomModel {
+  CONTENTFUL_FOOTNOTE = 'ContentfulFootnote',
+  CONTENTFUL_CODE_BLOCK = 'ContentfulCodeBlock',
+  CONTENTFUL_BLOG_POST_COLLAPSIBLE = 'ContentfulBlogPostCollapsible',
+}
+
+enum EmbeddedAssetType {
```
georgiee, line 39: thanks for that 🙌

src/components/pages/blog-post/blog-post.tsx
georgiee: Saw that, thanks!

src/templates/blog-post.tsx
```diff
@@ -101,6 +101,16 @@ export const BlogPostPageQuery = graphql`
             __typename
           }
 
+          ... on ContentfulFootnote {
+            contentful_id
+            __typename
+            note {
```
georgiee, line 107: so field `content` is gone 👋✅

georgiee (approved):
Code good. Deployment good. There is nothing left that could hold you back from merging that beauty anymore 💫 Let's merge it! Thank you for your efforts.

PR #549, opened 2022-04-04
feat: improve career form validation

milljoniaer:
### What is included?
- makes documents required
- sets a minimum length of 100 characters for the cover letter
- uses a honeypot field


src/components/pages/career-details/new-career-form/career-form.tsx
```diff
@@ -139,10 +147,18 @@ export const Form = (props: CareerFormProps) => {
             control={control}
           />
         </FormLayout>
+        <HoneypotField name="phone" label="Phone" control={control} />
```
feedm3: I think this can result in a problem: If you fill in forms with autocomplete in Chrome, this field may also be filled. Then, the user cannot submit the form anymore. Please check if this is actually the case and if there are any other issues with autocomplete in browsers. 

In case of doubt let's not add this field. There isn't that much spam.

milljoniaer: yes, in case of auto completion also the phone field is filled. I will leave the honeypot field out.

But I have tried this also at our contact form. I guess there was the same problem, since there was not message in slack.

feedm3 (approved):

Issue #548, opened 2022-04-04
Improve HTML headers


feedm3:
Currently, some of our pages are not using headers correctly. Especially the main page only has one `h1` and nothing else.

<img width="617" alt="Screenshot 2022-04-04 at 09 30 34" src="https://user-images.githubusercontent.com/7814926/161495179-329efd56-352c-4415-be07-26def595fefa.png">

Make sure that all our headlines are using the correct HTML headers to make the site more accessible and SEO friendly.

Some resources:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements
- https://chrome.google.com/webstore/detail/seo-meta-in-1-click/bjogjfinolnhfhkbipphpdlldadpnmhc?hl=de

milljoniaer:
#553 dealt with this

PR #553, opened 2022-04-05
feat/improve header order

milljoniaer:
### What is included?

This PR makes some smaller changes to improve the header structure of our pages.

### Some inconsistencies
There are some pages where I was not sure how to achieve a correct structure

#### Contact Page
- I would suggest to change the "Email" headline to "Contact us" and put it as h1 on top of the page
- Shall Address be then a h2 ? Or shall we add a new h1 and render "contact us" and "address" as h2?

#### Markdown Pages (Imprint and Data Privacy)
- The section headlines are currently "####" in markdown -> h4
- But for correct header structure they should be h2s
- Shall I change it to "##"? But it looks a little bit weird since they are so large then

image

closese #548

### DIFF
--- a/src/components/content/header-block/header-block.tsx
+++ b/src/components/content/header-block/header-block.tsx
@@ -46,7 +46,7 @@ const Topline = styled.p`
   margin: 0;
 `;
 
-const Headline = styled.p<LargeProps>`
+const Headline = styled.h2<LargeProps>`
   ${TextStyles.headlineL}
   letter-spacing: -0.01em;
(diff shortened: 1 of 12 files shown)


feedm3 (commented):
I did a smoke test on the deployed website. 2 things:
1. On the contact page, the "Contact" pre-title should be `h1`
2. We don't care about the correct headers on the imprint or data-privacy pages

feedm3 (approved):
Nice 👍

PR #554, opened 2022-04-05
chore: remove angular.png

milljoniaer:
### What is included?

This PR removes a last blog-post picture which was used in the teaser story. 

I used https://picsum.photos/ instead, they call themselves the "lore ipsum for photos"


src/components/content/teaser/teaser.stories.mdx
```diff
@@ -17,52 +15,61 @@ You can't pass in an image **and** illustration. If you pass in an image and ill
...
       title="Teaser Title"
-      linkTo='#'
-      image={<img style={{ width: "100%" }} src={ExampleImage} alt="" />}
+      linkTo="#"
+      image={
+        <img
+          style={{ width: '100%' }}
+          src={'https://picsum.photos/seed/sy-image/600/400'}
+          alt=""
+        />
```
JulianAnthes: I suggest we should increase the image size (old angular.png was 2000x900) and also use the `id` parameter to [pin to a specific image](https://picsum.photos/images).
```suggestion
        <img
          style={{ width: '100%' }}
          src="https://picsum.photos/id/1031/2000/900"
          alt=""
        />
```

JulianAnthes (approved):
I wrote a small suggestion, otherwise looks totally fine. ✨ 
Thanks for the cleanup!

src/components/content/teaser/teaser.stories.mdx
milljoniaer: You are right about the resolution, but regarding a specific image, I suggest to stick to this seed, this way we will also always have the same image.

JulianAnthes: > I suggest to stick to this seed, this way we will also always have the same image.

When we use a specific image identifier we will get exactly the image we specify (see the linked page). 
Seed uses a randomized image but will cache the result afterwards.

PR #557, opened 2022-04-05
Feat/improve rss feed

### LINKED ISSUE #535: Add RSS feed to the blog page
That's the most convenient way to get the updates from multiple websites in one place. 
I'm still using RSS feeds. Maybe I'm not the only one =)
**Update:** as RSS feed already exists, maybe we should add an RSS link to the blog articles overveiew page.

milljoniaer:
What is included?
- RSS Follow Panel on the Blog Overview Page
image

- Link to the RSS Feed in the headers of Blog and Blog Posts

closes #535


src/components/layout/seo.tsx
```diff
@@ -6,6 +6,8 @@ import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { I18nNextData } from '../../types';
 
 const DEFAULT_META_IMAGE_URL_PATH = '/sy-share-image.jpg';
+const RSS_URL = 'https://satellytes.com/blog/rss.xml';
+const RSS_TITLE = 'Satellytes Blog - RSS Feed';
```
feedm3: `Satellytes Blog` is enough.

feedm3 (approved):
Very good 👍

src/components/layout/seo.tsx
milljoniaer: 👍

PR #558, opened 2022-04-05
chore: localized links in markdown

### LINKED ISSUE #293: Localized links in rendered Markdown
It looks like our markdown renderer does not take the current language into account when turning `[...](...)` into `<a>` elements. That means that for links in our markdown files (e.g. `data-privacy-de.md`) we need to manually add `/de` to links. 

### Expected outcome
It would be nice if we wouldn't need to hardcode any locale id into the links in our Markdown.

milljoniaer:
### What is included?

I dont now if it is still needed but this closes #293 by using the Link component in the markdown custom components.


data/pages/data-privacy-de.md
```diff
@@ -8,7 +8,7 @@ language: de
 1. Im Folgenden informieren wir über die Erhebung personenbezogener Daten bei Nutzung unserer Website. Personenbezogene
    Daten sind alle Daten, die auf Sie persönlich beziehbar sind, z. B. Name, Adresse, E-Mail-Adressen, Nutzerverhalten.
 2. Verantwortlicher gem. Art. 4 Abs. 7 EU-Datenschutz-Grundverordnung (DSGVO) ist Satellytes Digital Consulting GmbH,
-   Carola-Neher-Straße 10, 81379 München, info@satellytes.com (siehe [Impressum](/de/imprint/)).
+   Carola-Neher-Straße 10, 81379 München, info@satellytes.com (siehe [Impressum](/imprint/)).
```
feedm3, line 11: That link was correct?

src/components/pages/blog-post/follow-panel.tsx
```diff
@@ -51,6 +50,11 @@ const SocialLinkItem = styled.li`
   }
 `;
 
+const SimpleLink = styled.a`
```
feedm3: Do we still need this?

src/components/legacy/markdown/custom-components.tsx
```diff
@@ -43,7 +44,7 @@ const OrderedList = styled.ol<{ doubleDigit: boolean }>`
   }
 `;
 
-export const SimpleLink = styled.a`
+const SimpleLink = styled(Link)`
```
feedm3: `SimpleLink` is a bit too generous, maybe `MarkdownLink`?

data/pages/data-privacy-de.md
milljoniaer: Yes, this link is now rendered as /de/imprint:
image

src/components/pages/blog-post/follow-panel.tsx
milljoniaer: Do you mean, if we can use a Gatsby Link instead? I will check again

milljoniaer: gatsby link works

src/components/legacy/markdown/custom-components.tsx
milljoniaer: renamed it

data/pages/data-privacy-de.md
feedm3: Yeah, but this is the german data privacy page, which should link to `/de/imprint` and not `/imprint` like in your change. Do I miss something? 😄

src/components/pages/blog-post/follow-panel.tsx
```diff
@@ -35,6 +35,7 @@ export const PanelText = styled.p`
 const SocialLinks = styled.ul`
   all: unset;
   order: 2;
+  color: ${theme.palette.text.link.default};
```
feedm3: You need to access the theme via props:

```
color: ${props => props.theme.palette.text.link.default}
```

feedm3 (changes requested):

data/pages/data-privacy-de.md
milljoniaer: Yes, but the thing in #293 is, that all links on all our pages should automatically append "de" if language === "de".

But somehow, this did not work in markdown, but by using the Link component in our custom Components this Automation also works here.

I changed the "german" links in markdown, so that they do not get rendered as "/de/de/imprint" on german.

src/components/pages/blog-post/follow-panel.tsx
milljoniaer: Ohps, fixed it 👍

data/pages/data-privacy-de.md
feedm3: You are right, thanks for the clarification 👏

feedm3 (approved):

PR #560, opened 2022-04-07
feat: add blog pagination

### LINKED ISSUE #533: Introduce Blog Pagination
Our blog is growing and growing and with the introduction of a CMS I think it's a good time to start thinking of a pagination to split the content in multiple pages. There are some resources available for Gatsby and I'm sure they help to get us running. 

1. This will involve adapting the contentful GraphQL query, providing a new component on the different blog pages and have a page query or path segment to hold the information. 
2. In addition we should think about SEO, do we need anything because of the changing page content? Changing or updated page dates (which @milljoniaer mentioned anyway in another issue about SEO, can't find the related issue though)
3. I think of 10-20 blog articles per page. No need to provide a filter or different page size configurations. We anticipate a change of the blog towards a "news outlet" in the future to hold more content (besides blog), so we will talk about filtering content at that time I guess. Let's keep things simple.

One related topic is the actual design. We have an initial pagination component provided through (#191) but I think we should see if that works with the actual implementation. Nonetheless we should pick that to get things running.

Resources:
+ https://www.gatsbyjs.com/docs/adding-pagination/
+ https://github.com/satellytes/satellytes.com/pull/191

milljoniaer:
### what is included?

This PR adds Pagination to our Blog Pages witch 10 Posts per Page.

The slug for the "main" blog overview page stays `/blog/`.
The next pages will be `/blog/2/`, `/blog/3/`.

### Screenshots
image

### Try it out
https://satellytescommain21751-feataddblogpagination.gtsb.io/blog/

This closes #533


georgiee (dismissed):
That looks great already 🙌
I only did a functional test without checking the code yet.

1. What will we do about SEO? Prevent index of the generated pages? Only allow the main page ?
2. You're in touch with Verena about improving the design? Because we talked about dropping the page selector as we can keep it more simple for example. Should we wait for that or do you think it's good enough to merge already because of the pagination logic and only later improve the visuals?

I block because of the SEO impact we need to discuss. Thanks🙌

JulianAnthes:
I like the feature, just want to throw in that the proposed url pattern `blog/[index]` could in theory match with a blog post.
Maybe we should rearrange the pages like this:

## Proposal 1

| Route                   | Description            |
|-------------------------|------------------------|
| `/blog/`                | Overview page          |
| `/blog/page/[index]`    | Pagination             | 
| `/blog/post/[slug]`     | Post                   | 
| `/blog/author/[slug]`   | Blog posts by author   |  
| `/blog/cateogry/[slug]` | Blog posts by category | 

_Author & category are possible future additions that came into my mind and are just for demonstration purposes_

## Proposal 2

Another possibility is to structure the blog like this:

| Route                                                  | Description          |
|--------------------------------------------------------|----------------------|
| `/blog/`                                               | Overview page        |
| `/blog/cateogry/[category-slug]/[year]/[month]/[slug]` | Post                 |
| `/blog/page/[index]`                                   | Pagination           |
| `/blog/author/[slug]`                                  | Blog posts by author |

All articles are reachable via `/blog/category/[slug]` like this:
`/gaming/2022/04/how-to-beat-elden-ring-in-under-9-minutes/`

I took this scheme from [arstechnica.com](https://arstechnica.com) and I think I would vote for proposal 2. 😄

milljoniaer:
Verena decided to display the page number for a small amount of pages and the dropdown once we have a larger amount.

Since we will (hopefully) have a larger amount of pages one day, I left the Dropdown in the code. It will be displayed, when there is an onDropSelect provided.

We can provide such a Callback once we want to have this Dropdown.


I also changed the slug pattern to this:
first blog overview: `/blog/`
rest blog overviews: `/blog/page/[index]`
blog posts: `/blog/post/[blog-post-slug]`

The `/blog/page/[index]` pages have a `no-index` tag.

I dont know if it is a problem, but there are no pages for `/blog/page/` and `/blog/post/`. Shall we provide a redirect to `/blog/` or is it fine to leave those pages out?

georgiee:
I like the new urls. If you have changed the urls for blog posts you must ensure that there is a redirect. Maybe change this later and focus on the technical changes for pagination?

No pages on `/blog/post/` etc should not be a problem but we could think of redirect. @feedm3 can help with the topic urls pretty good

milljoniaer:
I am currently working on the redirect of the blog pages from `/blog/[slug]` to `/blog/post/[slug]`and I was thinking about whether there should be a redirect for all blog post or only "manually" for those which are already there now?

milljoniaer:
I have now added redirects for all currently available public posts. 
The design should also be fine for now.

Please have another look on the PR.

feedm3:
I just talked with @milljoniaer: The redirects are important. We keep them for now in the redirects file manually but will refactor it after merging this PR. 
The idea is to make the **404 page an SSR page**. On the server-side, we can do a lookup on all blog posts to see if the URL is an old blog post URL, we can also move our redirects array there. It's a nice opportunity to introduce and I can also think of other projects where this could be powerful.

gatsby/create-pages/create-blog-post-overview-pages.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (contentfulBlogPages.errors || !contentfulBlogPages.data) {
+    reporter.panicOnBuild(
+      `Error while running GraphQL query for blog pagination pages.`,
+    );
+    return;
+  }
+
+  const posts = contentfulBlogPages.data.allContentfulBlogPost.nodes;
+  const postsPerPage = 10;
+  const numPages = Math.ceil(posts.length / postsPerPage);
```
feedm3: Like `postsPerPage` you can also use the full name here: `numberOfPages`. It's just simpler to read.

gatsby/create-pages/create-redirects.ts
```diff
@@ -57,14 +57,88 @@ const REDIRECTS = [
...
 
 /**
  * Sometimes we mess up URLs. Fixing them is necessary, but if Google or other
  * some other external links still link to the old URL they will only get the 404
  * page. That is why we need to redirect the old links to the new ones.
  */
-export const createRedirects = ({ actions }: CreatePagesArgs) => {
+export const createRedirects = async ({
+  actions,
+  graphql,
```
feedm3: Unused variable?

src/components/ui/pagination/pagination.tsx
```diff
@@ -46,43 +48,53 @@ const StyledButton = styled.button`
...
     const selectedPage = parseInt(selectedOption);
-    onDropdownSelect(selectedPage);
+    onDropdownSelect?.(selectedPage);
   };
 
   return (
-    <PaginationContainer>
-      <StyledButton onClick={onPreviousClick} disabled={currentPage === 0}>
+    <PaginationContainer className={className}>
+      <StyledButton onClick={onPreviousClick} disabled={currentPage === 1}>
```
feedm3: I think from an accessibility perspective, this should be a native link rather than a button. Instead of a link, you can also use `aria` labels to make the button more accessible. Can you please do some research?

feedm3 (changes requested):
Well done. I have left some comments, also I figured out that the `blog/pages` are still in the [sitemap.xml](`https://satellytescommain21751-feataddblogpagination.gtsb.io/sitemap/sitemap-0.xml`). They shouldn't be there - everything with `NO_INDEX` shouldn't be there.

gatsby/create-pages/create-blog-post-overview-pages.ts
milljoniaer: 👍

gatsby/create-pages/create-redirects.ts
milljoniaer: Ohps 🧹

src/components/ui/pagination/pagination.tsx
milljoniaer: You are right, According to https://www.a11ymatters.com/pattern/pagination/, Links are better than buttons and I also added area-labels.

They also suggested to wrap the navigation in a `<nav>` tag, but they used a simple list as pagination. I would suggest to leave that out, but what are your thoughts on that?

milljoniaer:
Thank you for comments. I have added tried to make the pagination more accessable and excluded the pagination pages "manually" from the sitemap.

Regarding the automatic exclusion of noindex pages of the sitemap, I raised a new issue #601 .
Same for the redirects via SSR on the 404 page: #602 

Please have another look at this PR

src/components/ui/pagination/pagination.tsx
feedm3: Very good! We can leave out the list with the links for now. Let's see how this feature evolves over time, maybe we can implement that later.

feedm3 (approved):

PR #571, opened 2022-04-11
chore: upgrade all dependencies

feedm3:
Dependabot only updates 10 dependencies. This PR updates all dependencies at once.

The types from `unified` don't fit to React 18, that's why I needed to one `any` to it. It still functions perfectly. The library is used for our Markdown parsing, which we only use on the imprint and privacy pages, and for the code blocks and footprint on the blog. The more we move to Contentful the less likely it is we will still need it, so my recommendation is to stick to this "any hot fix" and try to remove this library completely in the future.


georgiee (approved):
Welcome react 18. Works just like that? 😳✨
Gatsby & Storybook both build so I guess yes ? I would never ever upgrade like that but the build speaks for the results I guess or any objections?

feedm3:
Yep, it just works. There was a minimal API change that popular libraries already adapted 🥳

PR #572, opened 2022-04-12
feat: cache fonts on the CDN

feedm3:
# Problem
Our fonts are not served with a proper `Cache-Control` header. This lowers the page speed and lighthouse score.

Screenshot 2022-04-12 at 07 47 40

# Solution
Caching headers can be set via the [Gatsby Cloud plugin](https://www.gatsbyjs.com/plugins/gatsby-plugin-gatsby-cloud/#headers). I added an "infinite" caching time, as the fonts won't change.

Screenshot 2022-04-12 at 07 55 50

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -272,6 +272,11 @@ const gatsbyConfig: GatsbyConfig = {
     },
     {
       resolve: `gatsby-plugin-gatsby-cloud`,
+      options: {
+        headers: {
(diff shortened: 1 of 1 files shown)


georgiee (approved):
Nice! The hard dependency to Gatsby Cloud  is something I don't like i general, but I see the benefits of embracing it . Let's merge this 🙌

PR #574, opened 2022-04-12
fix: change mode to onTouched

milljoniaer:
### What is included?
There was always the "Please fill in all the mandatory fields" message even on valid message while submitting.

That was the case, since the `isValid` const does not work on the onSubmit mode. (according to the [docu](https://react-hook-form.com/api/useform/formstate))

I changed the mode to onTouched

It works now, the message is only shown when the Form is not valid


feedm3 (approved):
Well done 👍

PR #576, opened 2022-04-14
feat(blog): add fcb case components

milljoniaer:
### This adds the new components needed for the fcb case

the stats
image
[preview](https://6257f90a41c03a7096b4fd43--satellytes-website-storybook.netlify.app/?path=/docs/components-content-stats--regular&globals=locale:en;locales.en:English;locales.de:Deutsch)

the full width images
image

#### Build Fail
The build failes currently, because there is no ContentfulStats Object published so far. Therefore the query fails.

Maybe we should wait with merging this until we can publish the FCB Case in Contentful.

You can still test this component (also within the fcb case) by running this branch locally


georgiee (approved):
Thanks, looks good. Did you interact with @dschu-lab. I think it's a good idea in terms of contentful integration. I only glimpsed through the files and it looks good 👍

src/components/legacy/markdown/custom-components.tsx
```diff
@@ -228,6 +229,20 @@ const customSatellytesComponents = {
...
+        {statItems.map((item, index) => (
+          <StatItem
+            key={item.label + index}
+            title={item.label}
+            content={item.value}
+          />
+        ))}
+      </Stats>
+    );
+  },
```
JulianAnthes: I think we can move the rendering part directly into `src/components/pages/blog-post/blog-post.tsx` since the properties of `customComponents` represent html tags that are rendered by the markdown plugin.

src/templates/blog-post.tsx
```diff
@@ -111,6 +111,15 @@ export const BlogPostPageQuery = graphql`
...
 
+          ... on ContentfulStats {
+            contentful_id
+            __typename
+            statItems {
+              label
+              value
+            }
+          }
+
```
JulianAnthes, line 125: Welcome `ContentfulStats` 👋

src/components/content/stats/stats.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const StatsWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 48px;
+
+  ${up('sm')} {
+    flex-direction: row;
+  }
+`;
```
JulianAnthes: Lets replace those `div` tags with more meaningful ones, like the [dl tag](https://www.w3schools.com/TAGs/tag_dl.asp):

### Example HTML result
```html
<dl>
  <dt>Team Size</dt>
  <dd>6 people</dd>
  <dt>Start</dt>
  <dd>December 2018</dd>
  <dt>End</dt>
  <dd>Ongoing</dd>
</dl>
```

Since we need to remove the `Wrapper` (`StatsItem` wrapper), we should switch to a css grid for the `StatsWrapper`. 

I have created a small codepen as orientation for you: [Codepen Example](https://codepen.io/dschu/pen/LYegRNv)

JulianAnthes (changes requested):
Looks like a really great feature! Can't wait to see this released. But we should take the time and improve on the semantic html. 

Moving the stats rendering depends on if it is possible to use it in plain markdown. If it works, we can leave it as is, but if not then it would be better to "lift it one level higher".

src/components/legacy/markdown/custom-components.tsx
milljoniaer: you are right, I fixed it 👍

src/templates/blog-post.tsx
milljoniaer: 🥳

src/components/content/stats/stats.tsx
milljoniaer: thank you for suggestion, this is convenient, I changed it 👍

milljoniaer:
Thank you for your suggestions. I changed them, I also added the storybook deployment label and added the preview link to my description.

Please have another look.

JulianAnthes (approved):
Thanks for refactoring this into a description list. 
I have nothing more to add, great work! 💪

milljoniaer:
Since the fcb case also requires full width images, I added these also in this branch to have the possibility to see the whole fcb case in action when starting this branch locally.

PR #600, opened 2022-04-19
feat: add gesture handling to leaflet

milljoniaer:
### What is included?

This adds `leaflet-gesture-handling` to our leaflet on the contact page. That way it is possible to:
- scroll down the page even when hovering over the map
- zoom the map when hovering with pressing command or ctrl during scrolling
- zoom with two fingers on touchpad or mobile

### Some thougths

My implementation feels a little bit ugly, since `leaflet-gesture-handling` does not provide types (casting to any needed). 
And I added component returning null just to get access to useMap(). I dont know if this way is still fine, but I did not know a better way.

### DIFF
--- a/src/components/pages/contact/leaflet/leaflet.tsx
+++ b/src/components/pages/contact/leaflet/leaflet.tsx
@@ -1,3 +1,10 @@
+import * as L from 'leaflet';
+import { LatLngExpression } from 'leaflet';
+import { GestureHandling } from 'leaflet-gesture-handling';
+import 'leaflet-gesture-handling/dist/leaflet-gesture-handling.css';
+import 'leaflet/dist/leaflet.css';
+import React, { useEffect, useState } from 'react';
+import { Helmet } from 'react-helmet';
 import {
   CircleMarker,
   LayersControl,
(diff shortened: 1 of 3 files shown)


src/components/pages/contact/leaflet/leaflet.tsx
```diff
@@ -1,20 +1,25 @@
+import * as L from 'leaflet';
+import { LatLngExpression } from 'leaflet';
+import { GestureHandling } from 'leaflet-gesture-handling';
+import 'leaflet-gesture-handling/dist/leaflet-gesture-handling.css';
```
feedm3: I guess we don't need those imports if we put them into React Helmet below with:

```
 <link
          rel="stylesheet"
          href="//unpkg.com/leaflet-gesture-handling/dist/leaflet-gesture-handling.min.css"
          type="text/css"
        />
```

It's better to use the React Helmet import, instead of `import`

src/components/pages/contact/leaflet/leaflet.tsx
```diff
@@ -1,20 +1,25 @@
...
+import 'leaflet/dist/leaflet.css';
```
feedm3: Same for this. Shouldn't be needed

src/components/pages/contact/leaflet/leaflet.tsx
```diff
@@ -57,13 +62,23 @@ const MapWrapper = styled.div`
...
 
+const EnableGestureHandling = () => {
+  /**
+   * leaflet-gesture-handling does not add gestureHandling to the types
+   * but map.gestureHandling exists and works
+   */
+  const map = useMap() as any;
+  map.gestureHandling.enable();
+  L.Map.addInitHook('addHandler', 'gestureHandling', GestureHandling);
+  return null;
```
feedm3, line 71: You can put this method directly into the `Leaflet` component. It's usually nice to put features into own components, but this time it looks like too much.

feedm3 (changes requested):
Well done, I have 3 suggestions 🥳

src/components/pages/contact/leaflet/leaflet.tsx
milljoniaer: I tried to do so, but useMap() can only be used in child components of the MapContainer, that is why it was not possible to do so

src/components/pages/contact/leaflet/leaflet.tsx
milljoniaer: 👍

src/components/pages/contact/leaflet/leaflet.tsx
milljoniaer: 👍

milljoniaer:
Thank you for the suggestions, I have implemented them or added a comment.
Please have another look

feedm3 (approved):

PR #603, opened 2022-04-19
feat: fix translation & cleanup gestureHandling code

milljoniaer:
### What is included?
- adds translation for the helper messages for the leaflet gesture handling
- cleans up the code by using the gestureHandlingOptions of MapContainer (with ts-ignore, since leaflet-gesture-handling does not provide types)

The documentation of `leaflet-gesture-handling` is unclear about the possible props. The `gestureHandlingOptions` Prop does not work as expected especially with duration and text together. But at least gestureHandlingText has worked now. Read this for more information: https://github.com/elmarquis/Leaflet.GestureHandling/issues/47


feedm3 (approved):

feedm3 (changes requested):
I just checked it on the deployed preview and the text and information don't show up anymore when you scroll. Something seems to be broken.

milljoniaer:
I checked it again, and it is weird, when it works and when it doesn't.

But did you try it with a mouse or with a touchpad?

It seems, that is only working while you are actually scrolling which is only possible to do with the touchpad. 
I will do some research on this.

feedm3:
Touchpad on Mac. Everything works fine on satellytes.com, so the bug must come from this change.

milljoniaer:
I think I figured the problem out. When setting the custom text, somehow the duration of Text is set to 0.
It is possible to set this duration, but it works only if there is no text set.

What shall we do? Do we leave the text out? (when there is no text set, the browser language is selected fro the map text)

feedm3 (approved):
Very nice feature! Also works now properly with the right props 🙂

PR #605, opened 2022-04-21
fix: break long links and inline codes in blogs

### LINKED ISSUE #604: Wrap long links in blog content
Reported by @maxmarkus 
Long links kill the layout and we should force a wrapping of that by using some css on links (something around break-word, overflow-wrap).

I can't reproduce it in Safari, not in Firefox.
Referenced article is https://satellytes.com/blog/post/cloudfront-cache-efficiency/

Screenshot 2022-04-20 at 17 17 46

milljoniaer:
### What is included?

This PR breaks links and inline codes in blog posts when overflowing, to prevent them from killing the layout on mobile.

closes #604

### DIFF
--- a/src/components/legacy/markdown/custom-components.tsx
+++ b/src/components/legacy/markdown/custom-components.tsx
@@ -118,10 +118,6 @@ const Figure = styled.figure`
   }
 `;
 
-const StyledCode = styled.code`
-  overflow-wrap: anywhere !important;
-`;
-
 const Figcaption = styled.figcaption`
   ${TextStyles.textS};
   color: ${theme.palette.text.timestamp};
@@ -232,9 +228,6 @@ const customSatellytesComponents = {
   figcaption(props) {
     const { children, ...rest } = props;
     return <Figcaption {...rest}>{children}</Figcaption>;
-  },
-  code(props) {
-    return <StyledCode {...props}>{props.children}</StyledCode>;
   },
 };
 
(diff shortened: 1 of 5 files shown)


src/components/legacy/markdown/custom-components.tsx
```diff
@@ -117,6 +118,10 @@ const Figure = styled.figure`
   }
 `;
 
+const StyledCode = styled.code`
+  overflow-wrap: anywhere !important;
```
hofmann-felix: For code formatting we use [PrismJS](https://www.gatsbyjs.com/plugins/gatsby-remark-prismjs/). For this there are some CSS files in the project. Maybe instead of the new `StyledCode` component and the `!important` you can add the CSS directly here. I have not tried it myself yet.

https://github.com/satellytes/satellytes.com/blob/9e091ad924196e9df2d92c0c4f32e743ead77ccb/src/assets/prism-themes/coy-with-shadows.scss#L33-L44

hofmann-felix (commented):
Thanks for the quick fix, looks great 🙏 I made one suggestion to get rid of the `!important`

src/components/legacy/markdown/custom-components.tsx
milljoniaer: Thank you for your suggestion, this is way cleaner, I was already wondering where the language css comes from.
Fixed it 👍

feedm3 (approved):
Perfect! Also, excellent suggestion from @hofmann-felix 👍

Issue #611, opened 2022-04-27
Blog Post: Extend `introText` with rich text functionalities


feedm3:
It should be possible to have some basic formatting in the `introText` of a blog post:

- Paragraphs
- Bold Text
- Links

Also make sure to update the RSS feed!

<details>
<summary>Example of a use-case</summary>
<img src="https://user-images.githubusercontent.com/7814926/165520782-0212dca8-77ff-44c7-a59c-41c67ae358eb.png"</img>
</details>

It should not be possible to add something else, like images or more advanced rich text components.

<details>
<summary>It's possible to limit the options on Contentful</summary>
<img width="880" alt="Bildschirmfoto 2022-04-27 um 14 25 55" src="https://user-images.githubusercontent.com/7814926/165521195-8ac54ee1-3a30-4682-b02b-ad5530724d03.png">
</details>

PR #618, opened 2022-04-29
feat: wrap RichText into component

milljoniaer:
### What is included?
I dont know if this is actually necessary, but on the ADAC project I came in contact with RichText wrapped into one component.

That way also the blog post page gets way cleaner.

This could in addition also be a first step for adding Rich Text Functionality in other places (e.g. #611 or add imprint and data-privacy in contentful via rich Text)

### Further Thoughts

The type of the richText in graphql is ContentfulBlogPost and should better be ContentfulRichText (naming and use in other places). This name is auto generated by the plugin. 
https://github.com/satellytes/satellytes.com/blob/65b7c32a7597ceb2279f94786d9ad1e00d082e52/src/components/content/rich-text/rich-text-query.ts#L4

Maybe we could rename it and provide better types for the queries by hardcoding the schema like gatsby suggests [here](https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/#creating-type-definitions). That way, we also would not have the problems we had with new components like in the fcb case.

Let me hear your thoughts on this.


src/components/content/rich-text/rich-text-query.ts
```diff
@@ -0,0 +1,96 @@
+import { graphql } from 'gatsby';
+
+export const richTextFragment = graphql`
+  fragment richText on ContentfulBlogPostContent {
```
feedm3: As you already proposed in your description, please give it a more meaningful name.

src/components/content/rich-text/rich-text.tsx
```diff
@@ -0,0 +1,330 @@
...
+/**
+ * Custom Contentful Rich Text Renderer
+ *
+ * Will automatically filter out ContentfulFootnote elements and appends an
+ * ordered list of footnotes after the rendered rich text.
+ *
+ * @param data rich text data
+ * @returns rendered elements
+ */
+const customContentfulRenderer = (
```
feedm3: This can also have a better name, like `ContentfulRichTextRenderer`, same with the file name. The rich text is explicitly for Contentful rich text, so `contentful` should also be in the file name.

src/components/content/rich-text/rich-text.tsx
```diff
@@ -0,0 +1,330 @@
...
+        <>
+          <hr />
+          {footNoteElements}
+        </>
+      )}
+    </div>
+  );
+};
+
+export const RichText = ({ data }: RichTextProps) => {
```
feedm3: Is this wrapper even necessary? Just a thought, but maybe the `customContentfulRenderer` can be a react component in the first place.

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -396,7 +76,8 @@ export const BlogPostPage = ({ blogPost, breadcrumb }: BlogPostPageProps) => {
         <BlogHeader headline={blogPost.title} byline={heroByLine}>
           {blogPost.introText?.introText}
         </BlogHeader>
-        {customContentfulRenderer(blogPost.content)}
+
+        <RichText data={blogPost.content} />
```
feedm3: It's so much better 👏

feedm3 (changes requested):
Very good idea to put the rich text component into its own file! 👍  I think you can already improve the code with some better namings, see my comments.

georgiee:
Oh yeah, thanks! That looks clean. I would love to get feedback from @dschu-lab as he provided the implementation of the current state 🙏

src/components/content/rich-text/rich-text-query.ts
milljoniaer: The problem with with renaming it is, that the type is autogenerated by the plugin using the published entities of that type in contentful. I do not know how to change that behavior. I have only read that there is a possibility to hardcode to types somehow. At ADAC we also had a `schema.gql` where all types have been hardcoded.

(This would have also prevented the build fail we had with the fcb case)

milljoniaer: I opened a new pull request to add a hardcoded schema for graphql. 
That way we can rename this graphql type just like we want.

see here for more information #622

src/components/content/rich-text/rich-text.tsx
milljoniaer: I changed the name to `ContentfulRichText` and removed the Wrapper 👍

src/components/content/rich-text/rich-text.tsx
milljoniaer: See previous comment 👍

milljoniaer:
I have added some changes like @feedm3 suggested and opened #622 to be able to adjust the graphql schema to use the rich text fragment wherever needed.

Please have another look.

src/components/content/rich-text/rich-text-query.ts
feedm3: Sorry, that's not what I meant. I just meant to rename the variable and fragment name. Like this:

```
export const ContentfulRichText = graphql`
  fragment ContentfulRichText on ContentfulBlogPostContent {
    ...
```

This works without changing anything to GraphQl.

feedm3:
Sorry for the confusion, the GraphQl PR is not necessary 🙈 I added a comment, after implementing that feel free to merge. I approve in advance.

feedm3 (approved):

src/components/content/rich-text/rich-text-query.ts
milljoniaer: Oh, well, now I got your point, I changed it 👍

PR #615, opened 2022-04-29
feat: make 404 page use ssr

milljoniaer:
### What is included?

The 404 page is now rendered with server-side rendering. Once the 404 page is triggered, the server checks for a redirect for the requested URL and returns a 301 redirect, otherwise a 404 page is returned.

We also added some convenience redirects to some of our internal tools. Those are marked with the `x-robots-tag: noindex`  header to not get indexed by Google.

Test
- https://satellytescommain21751-featmake404ssr.gtsb.io/office: Should redirect to `/about-us/`
- https://satellytescommain21751-featmake404ssr.gtsb.io/github: Should redirect to `https://github.com/satellytes`

<details>
<summary>Screenshot of internal redirect</summary>

</details>

<details>
<summary>Screenshot of redirect to Github with `noindex`</summary>

</details>

### Current status
`yarn build && yarn serve` work locally, but for `yarn start` and our [preview build](https://satellytescommain21751-featmake404ssr.gtsb.io/) the fallback page is not triggered by non-existing URLs and no redirect is returned. 

We create an issue for this bug here: https://github.com/gatsbyjs/gatsby/issues/35527

> **edit:** Let's merge this PR with the release of Gatsby v4.15 on 24.5.. We can remove the `ts-ignore` and maybe even have better support in `yarn start` or the preview builds on Gatsby Cloud.

PR #620, opened 2022-05-02
feat: add some convenience redirects

feedm3:
Every now and then I manually enter the URL to one of our tools. My browser always autocompletes to `satellyes.com`, but never to `slack.satellytes.com` (too many different URLs with `slack` in the title or history). Same with the other URLs. 

With those redirects, the autocomplete in the browser gets more useful.

Test:
- https://satellytescommain21751-featconvenienceredirects.gtsb.io/github


gatsby/create-pages/create-redirects.ts
```diff
@@ -156,16 +180,25 @@ export const createRedirects = async ({ actions }: CreatePagesArgs) => {
...
       toPath: redirect.toPath,
       isPermanent: true,
       redirectInBrowser: true,
     });
   });
+
+  EXTERNAL_REDIRECTS.forEach((redirect) => {
+    createRedirect({
+      fromPath: redirect.fromPath,
+      toPath: redirect.toUrl,
```
feedm3, line 199: On my local machine, the redirect to an external domain was working - to my surprise. I always thought that external URLs don't work. Pretty sure it also works on Gatsby Cloud then.

georgiee (commented):
That's convenient indeed 🙌 I'm worrying about google indexing those. I think we should prevent that as I don't want that those urls appear in google for any reason. How would we put those urls on a robots list or similar? Or is that exactly risky because we expose specifically our "secret" urls ? 😄

This is mixing internal and external interests which makes me uncomfortable. Is there any risk or issue beyond google doing that? Maybe we should wait for some other opinions on that 🙏

feedm3:
We can use the `x-robots-tag: noindex` header to prevent indexing the URLs: 

https://developers.google.com/search/docs/advanced/robots/robots_meta_tag#xrobotstag

I will put the code into #615 where we have control over all headers.

georgiee:
Thank you for the clarification and the idea to put it into the 404 PR🙌

PR #621, opened 2022-05-03
Feat/add parallax

### LINKED ISSUE #394: Design (optional): Parallax Scroll Image Masking
We are using narrow full-width images as separators on the website, for example on home (https://satellytes.com/) twice. I would like the images to be full-width and full-height, but masked by the current size of the cut-out (height: 360 px). When scrolling, the cut-out should reveal the image behind. Nice parallax effect.

milljoniaer:
### What is included?

- add gbimage-bridge as dependency like suggested [here](https://www.gatsbyjs.com/plugins/gatsby-background-image/#how-to-use)
- use gatsby-image-background to build a parallax component
- use Parallax on LandingPage for the Spacers
- the parallax is only rendered in the browser to prevent [this](https://reactjs.org/docs/error-decoder.html/?invariant=418) error

### Disclaimer

The parallax does not work on Safari, since it does not support `background-attachement: fixed` ([reference](https://caniuse.com/?search=background-attachment))

### Further Thoughts

The effect looks a little weird on mobile and on some mobile browsers does not work properly.
I suggest to think of using another image for mobile devices or disabling this effect on mobile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "gatsby-transformer-json": "^4.13.0",
     "gatsby-transformer-remark": "^5.13.0",
     "gatsby-transformer-sharp": "^4.13.0",
+    "gbimage-bridge": "^0.2.1",
     "html-entities": "^2.3.2",
     "i18next": "^21.6.16",
     "leaflet": "^1.8.0",
--- a/src/components/pages/landingpage/image-spacer.tsx
+++ b/src/components/pages/landingpage/image-spacer.tsx
@@ -37,5 +37,9 @@ export const ParallaxImageSpacer = ({ image }) => {
   const gatsbyImageData = getImage(image);
   const bgImage = convertToBgImage(gatsbyImageData);
 
-  return <StyledBackgroundImage {...bgImage} preserveStackingContext />;
+  return (
+    // eslint-disable-next-line
+    // @ts-ignore get rid of "error TS2589: Type instantiation is excessively deep and possibly infinite."
+    <StyledBackgroundImage fluid={bgImage?.fluid} preserveStackingContext />
+  );
 };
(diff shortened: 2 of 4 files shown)


feedm3 (changes requested):
This PR looks simple at the first glance but actually is tricky to implement for all the edge cases.

First off, I'm not a big fan of this parallax effect personally 😄  But let's ignore that. The image is now not embedded anymore in the HTML that is built. This means the first view of the website is without those pictures. The problem here is, that the content now jumps if the images are not loaded fast enough: 

Screenshot 2022-05-10 at 10 40 45

A solution can be to implement an empty placeholder diff. This prevents the page to jump. A seconds additional solution is to wrap the image with an `IntersectionObserver` and preload the image before it comes into the viewport. Both solutions are necessary to make the page look "cheap". 

Maybe you try to use the native Gatsby image again and put a wrapper `div` around it to simulate the parallax effect? Could be simpler as implementing a place holder div and the `IntersectionObserver`.

georgiee (changes requested):
Great that you managed to get the parallax effect into the images on the landingpage 👏 I can only second what Fabian said technically. What's missing here is, whenever we add something that animates we should always respect the user and if they [prefer reduced motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) we should reduce or disable such effects.

In addition I also don't like that effect. This makes me wonder @feedm3 why did we have let it come that far? Why did we not prevent @milljoniaer from working on that topic by blocking the issues in before hand? I know it came from the design team but I think we should raise our concerns and beliefs when it comes to new features.

I also don't like to see this feature on our page. It brings complexity without any added value. That's now of course a stupid situation as @milljoniaer put quite some effort into delivering this and now we doubt it's not the best idea. What should we do?

milljoniaer:
Thank you for your comments, also about the `prefer-reduced-motion`. I did not know about that feature and I am personally even thinking about using that on my PC.

But regarding the parallax, I suggest that you talk about your concerns with eric, since he already reviewed this feature from a design point of view. If you agree to drop that Issue/feature, it might be the best to simply close this PR.

Therefore I will not implement the `prefer-reduced-motion` functionality.

PR #622, opened 2022-05-06
feat: use gatsby-schema-snapshot

milljoniaer:
### What is included?

This adds [`gatsby-plugin-schema-snapshot`](https://www.gatsbyjs.com/plugins/gatsby-plugin-schema-snapshot/) as dependency to satellytes.com.

I added [`schema.gql`](https://github.com/satellytes/satellytes.com/blob/e9c8da251269b7e9d6b1495d8ece0093a08a5a4a/schema.gql) using this plugin. That way our qraphql schema is now hardcoded. We have to change schema.gql whenever changing something on the content model but we can also change the naming of types (like needed in #618).

This will also prevent us from the build fails whenever adding something to our contentful content model (like the stats on the fcb case)

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -323,6 +323,13 @@ const gatsbyConfig: GatsbyConfig = {
       resolve: 'gatsby-source-contentful',
       options: contentfulConfig,
     },
+    {
+      resolve: `gatsby-plugin-schema-snapshot`,
+      options: {
+        path: `schema.gql`,
+        update: false,
+      },
+    },
   ],
 };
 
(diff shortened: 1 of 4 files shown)


feedm3 (changes requested):
We don't need this extra layer, so **this PR can be closed**. I still give you some feedback:

From an architectural point of view, this is a big change. We now created a new layer between our data and Contentful. This layer adds complexity and needs to be explained and understood by all colleagues. This explanation is missing in the PR.
There is also no description of how to update the types. If someone changes the schema in Contentful, she would be pretty confused for a while why the changes are not working. So what I would expect is 2 things:

1. A description of how and why the schema is hardcoded.
2. A script to update the schema. In the best case, this even works automatically.

PR #625, opened 2022-05-13
refactor(blog): full Width layout

### LINKED ISSUE #608: Blog Post: Implement full-width layout for blog posts with CSS Grid
Our blog post can have images, that are `100vw` in width. Currently the solution is quite hacky, as we cannot really breakout of our layout. But if we use CSS Grid as our standard layout in the blog post, this is possible without any hacks.

A perfect tutorial for this can be found here: https://www.joshwcomeau.com/css/full-bleed/

milljoniaer:
### What is included? 

This refactors the Blog Post Layout using grid, just like [here](https://www.joshwcomeau.com/css/full-bleed/). I had to make some adjustments to make the css fit our html structure.

I am not sure if this is the best way, but I could not think of better adjustments.

Please have a look at these changes.

### How to verify 

- check that the full Width still works [here](https://satellytescommain21751-refactorbloglayout.gtsb.io/blog/post/fc-bayern-website-frontend-relaunched-by-satellytes/)


ghost:
**satellytes.com-main**

:tada: Your build was successful! [See the Deploy preview here.](https://satellytescommain21751-refactorbloglayout.gtsb.io)

## Build Details

[View the build logs here.](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/674be368-adf0-4ac5-8fae-f9a4493adc51/builds/bd24649c-878c-4261-8e06-2d4b14497c28/details)

:clock1: Build time: **3m**

## Performance

### Lighthouse report

| Metric | Score |
| ----- | ---- |
| Performance |   :large_orange_diamond: 76   |
| SEO |   :large_orange_diamond: 86   |

:link: [View full report](https://build-bd24649c-878c-4261-8e06-2d4b14497c28.gtsb.io/reports/lighthouse/index.html)

src/components/content/advanced-asset/advanced-asset.tsx
```diff
@@ -1,49 +1,40 @@
...
-  width: ${(props) => props.$width};
-  position: relative;
-  left: 50%;
-  right: 50%;
-  margin-left: -50vw;
-  margin-right: -50vw;
+const FullWidthContainer = styled.div`
+  width: 100%;
+  && {
+    grid-column: 1 / 4;
```
feedm3: So simple and good 👌

src/components/layout/layout.tsx
```diff
@@ -30,28 +30,34 @@ const FullHeightContainer = styled.div`
...
 `;
 
-const Main = styled.main`
+const Main = styled.main<{ fullWidth?: boolean }>`
   /** make the element take up all available space */
   flex-grow: 1;
 
-  display: grid;
+  ${({ fullWidth }) =>
+    !fullWidth &&
```
feedm3: This looks wrong. If you remove the `display: grid`, the full width trick with `grid-column: 1 / 4;` shouldn't work anymore. Also, the main component now needs to know if there is a full width children, which shouldn't be the case.

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -28,6 +28,15 @@ const PanelContainer = styled.div`
   }
 `;
 
+const StyledArticle = styled.article`
+  display: grid;
```
feedm3: Ah, here you introduce the grid again. This shouldn't be necessary if you don't remove the grid from the parent.

feedm3 (changes requested):
Really like direction this PR goes. The layout is just so much simpler.

milljoniaer:
Thank you for your comments. My problem was, that we have the following structure:

```
<Layout>
    <Article>
        <RichText />
    </ Article>
</ Layout> 
```

If I put the `display: grid` into the Layout, I can not make RichText Elements full width by using `grid-column: 1 / 4` since there is a layer in between.
Or do I get something wrong?

The `fullWidth` flag should only disable the grid in the Layout, so I could put it in the `<article>` tag.

I know that this is not very clean, do you have a suggestion how we could make it better?

feedm3:
@milljoniaer: Every page should have the same structure. The blog post is different because it didn't matter in the first place, but now it's a problem. We need to align the blog post with all other pages.

I guess that if you remove the `article` HTML tag, the page should work fine? In this case, make the layout component dynamic in the case of the HTML tag and remove the extra `article` nesting. This can be done with the styled-components `as=""` tag, with which you can make the Layout rendered as `article`.

milljoniaer:
I tried to implement your suggestion, please have another look and verify that the fullWidth still works in the [blog posts](https://satellytescommain21751-refactorbloglayout.gtsb.io/blog/post/fc-bayern-website-frontend-relaunched-by-satellytes/).

src/components/content/advanced-asset/advanced-asset.tsx
```diff
@@ -1,49 +1,40 @@
...
-const FullWidthContainer = styled.div.attrs((props: ContainerProps) => props)`
-  width: ${(props) => props.$width};
-  position: relative;
-  left: 50%;
-  right: 50%;
-  margin-left: -50vw;
-  margin-right: -50vw;
+const FullWidthContainer = styled.div`
+  width: 100%;
+  && {
```
feedm3: - Why is the `&&` needed?
- Please add an empty line between those 2 lines, if the selector is still needed. CSS needs some space to breath ;-)

src/components/layout/layout.tsx
```diff
@@ -82,6 +81,7 @@ interface LayoutProps {
   translation?: string;
   leadbox?: LeadboxProps;
   breadcrumb?: BreadcrumbEntry[];
+  mainAs?: any;
```
feedm3: We should not overwrite `main`.

Here is how it's implemented in zeit.de:
grafik

So `main` should always stay there. A solution could be to add an additional HTML tag inside main, that is either a `div` or an `article`.

feedm3 (changes requested):

src/components/content/advanced-asset/advanced-asset.tsx
milljoniaer: Ohps, it is not needed => 🧹

src/components/layout/layout.tsx
milljoniaer: okay, yes that seems to be convenient. I added such a wrapper. Please have another look

milljoniaer:
I have tried to implement your changes. Please have another look

feedm3 (approved):
Well done, it's so much better now 👍

PR #626, opened 2022-05-13
feat: add rich Text functionality to introText

### LINKED ISSUE #611: Blog Post: Extend `introText` with rich text functionalities

milljoniaer:
### What is included?

This PR adds the implementation for Rich Text for the intro Text of blog Posts.

Unfortunately we can not use the richText fragment in `content/rich-text/rich-text-query.ts` since we have no Embedded Entries published so far in the BlogPosts introText.

Maybe we could again think about adding `gatsby-plugin-schema-snapshot`. With that plugin it would be possible to implement everything, so that we can even use embeds in contentful.

### Example

The [FCB Case](https://satellytescommain21751-featrichtextforintrotext.gtsb.io/blog/post/fc-bayern-website-frontend-relaunched-by-satellytes/) is implemented using the new RichText Intro Field. I added a link [`www.fcbayern.com`](https://www.fcbayern.com) at the end to "proof" rich text functionality.


ghost:

:triangular_flag_on_post: Your build failed. [See the build logs here](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/674be368-adf0-4ac5-8fae-f9a4493adc51/builds/42a7b816-d7ef-48a1-949a-411348319749/details)

src/types.ts
```diff
@@ -73,6 +73,7 @@ export interface ContentfulBlogPost {
   introText?: {
     introText: string;
   };
+  introRichText?: RenderRichTextData<ContentfulRichTextGatsbyReference>;
```
feedm3: The type is wrong here, as it's not really all the rich text fields but only `raw`. Can you please align that and also add a comment on why this is (currently) necessary?

Also, the `introText` above can be removed, right?

src/components/pages/blog-post/blog-post.tsx
```diff
@@ -28,6 +29,15 @@ const PanelContainer = styled.div`
   }
 `;
 
+const RichTextContainer = styled.div`
+  > * {
```
feedm3, line 33: Please do it in the styled-components way:

```
const IntroRichText = styled(ContentfulRichText)`
   ....
`
```

Also, `> * {` seems like a too general selector that would also overwrite possible internals in the rich text component. Try to make the selector a bit more strict.

feedm3 (changes requested):
Nice feature, looking forward to having it on the blog. Thanks :)

src/types.ts
milljoniaer: I removed the introText and changed the type, but to fit the RichText Props I had to do this ugly thing:

https://github.com/satellytes/satellytes.com/blob/bcf99dd66afddf1ea685562cecce0dfc68ad714a/src/components/pages/blog-post/blog-post.tsx#L82-L87

Have you any idea how I could avoid this? The types come directly from gatsby, so I cannot make references optional.

src/components/pages/blog-post/blog-post.tsx
milljoniaer: I tried to do this, but since the `ContentfulRichText` Component returns an empty JSX.Element tag (which we need for the fullWidth layout in the blog), it did not work out.

This was the only way I was able to do this.

src/components/content/rich-text/rich-text.tsx
```diff
@@ -57,7 +57,7 @@ export const ContentfulRichText = ({ data }: ContentfulRichTextProps) => {
   /**
    * Filter out footnotes from content references
    */
-  const footnotes: ContentfulFootnoteReference[] = data.references.filter(
+  const footnotes: ContentfulFootnoteReference[] = data.references?.filter(
```
feedm3: `?` is a code smell and should be avoided. Especially with lists, try to make sure that they are never `null` or `undefined` but always empty. 

If you change this line here to 
```
data.references?.filter(...) | | []
```
you can remove all further `?` checks here in this file.

feedm3 (changes requested):

src/components/content/rich-text/rich-text.tsx
milljoniaer: Much better 👍

milljoniaer:
I tried to add your suggestions, please have another look

feedm3 (approved):
Thanks for your effort, cool to have this feature now :-)

PR #640, opened 2022-06-08
feat: use Team Member content type

### LINKED ISSUE #639: Contentful Migration for Team Images
I would like to manage the team through contentful. Right now it's a json file plus images.

+ Provide Content Model in Contentful
+ Provide the actual limplementation

The initial contentful migration was coordinated by @dschu-lab so I think it's a good idea to get guidance from him🙏

Phil147:
resolves #639 

Instead of the json file there is now a new `Team Member` content type in contentful which can hold the name and the picture of a lovely colleague 🖼️ 
With this current implementation the images will be loaded from the contentful image api and will not be part of the build output as before. There seems to be an option `downloadLocal` if we want to restore that.

A thing I noticed was that the people in the source pictures are not always centered. So the gatsby image plugin seems to have done some magic to crop the images and center the persons for the build output to some extent. With the switch to the contentful image api the pictures where shifted. Luckily the api has an option to focus on faces. A minor detail is some people are now a bit more shifted to the left because the faces got more centered but others are now better centered.


src/pages/about-us.tsx
```diff
@@ -38,20 +38,23 @@ export const AboutUsPageQuery = graphql`
...
+            aspectRatio: 1
+            layout: CONSTRAINED
+            placeholder: DOMINANT_COLOR
+            formats: [AUTO, WEBP, AVIF]
+            # people on the photos are not always in the center.
+            # this option from the contentful image api tries to resolve that.
+            cropFocus: FACE
+            resizingBehavior: FILL
+            jpegProgressive: false
+          )
```
georgiee: oh, that's actually much less code than expected 🤩 Thanks for the content model too!
Screenshot 2022-06-09 at 09 08 03

src/pages/about-us.tsx
```diff
@@ -38,20 +38,23 @@ export const AboutUsPageQuery = graphql`
...
-            )
-          }
+          gatsbyImageData(
+            width: 400
+            aspectRatio: 1
+            layout: CONSTRAINED
+            placeholder: DOMINANT_COLOR
+            formats: [AUTO, WEBP, AVIF]
+            # people on the photos are not always in the center.
+            # this option from the contentful image api tries to resolve that.
```
georgiee: Actually the comment is not accurate. These are options from gatsby-image and the underlying image process sharp. See [gatsby-plugin-sharp](https://www.gatsbyjs.com/plugins/gatsby-plugin-sharp) and [gatsby-image](https://www.gatsbyjs.com/plugins/gatsby-image/)

While I appreciate that you want to provide a hint to future struggling folks I suggest to remove the comment. People working into this repository should be comfortable querying the gatsby documentation for those keywords 🙏

I also suggest to remove the comment.

```suggestion
```

georgiee (approved):
Thank you for the change! And also for the nite about the image sources.
I'm unsure. I think by retrieving it locally we would enable them to be served by gatsby cloud again. On the other side contentful assets are also served through their own CDN (https://www.contentful.com/developers/docs/references/content-delivery-api/).

So let's merge this and change once we have some more specific opinion (maybe by @feedm3 ).

Let's merge!

src/pages/about-us.tsx
Phil147: > I'm unsure. I think by retrieving it locally we would enable them to be served by gatsby cloud again. On the other side contentful assets are also served through their own CDN (https://www.contentful.com/developers/docs/references/content-delivery-api/).

I couldn't judge the pros and cons really well here tbh as I'm pretty new to the gatsby world so I went with the version that I got running pretty quickly. Retrieving the images locally should be a change we can make pretty easily next :)

PR #662, opened 2022-09-09
Fix/rss render embedded entries

### LINKED ISSUE #648: RSS Feed: Embedded entries are not rendered to feed xml
## What happened?

When reading our blog using the rss feed, not all content is shown.

The following images show the original blog post (left) compared to `source.content.raw` (right):
Bildschirmfoto 2022-06-30 um 15 34 33

https://github.com/satellytes/satellytes.com/blob/5ba8b2fe119569c9149b21fcadc5bedd94096069/gatsby-node.ts#L139

The code block itself is rendered empty:
```js
const sourceContentRaw = [
  // ...
  {
    data: {
      target: {
        sys: { id: "xxxxxxxxxxxxxxxx", type: "Link", linkType: "Entry" },
      },
    },
    content: [],
    nodeType: "embedded-entry-block",
  },
  {
    data: {},
    content: [{ data: {}, marks: [], value: "", nodeType: "text" }],
    nodeType: "paragraph",
  },
  // ...
];

```

## What is affected?

The xml feed does not contain any code snippets. This clouds  the user experience when consuming our blog by the rss feed. But also the import to dev.to is affected.

## Possible solution

Needs some investigation. I'd guess we need to provide a custom renderer or similar.

timonsommer:
The intro text and all embedded entries and assets should be rendered properly now. In order to view the referenced assets in your RSS interpreter, you will need to run both `build` and `develop`, since some asset file paths rely on the data on port 8000 (opened by `npm run develop`). Alternatively, you can also change the `BASE_URL` in `gatsby-config.ts` accordingly.

For reference: original blog post (left) compared to a snippet from `blog/rss.xml` (right):

RSS XML


.husky/pre-commit
```diff
@@ -1 +1,4 @@
-yarn precommit
```
feedm3: Why this change? Can you please reference some docs on why the husky command needs to be added?

Also, we don't use `npm` in the repo, only `yarn`. And in order to manage what needs to be run it's better to maintain that in the `package.lock` file rather then here in the husky file. That's why we run `yarn precommit`, which could run `yarn test` or anything else.

gatsby-node.ts
```diff
@@ -136,6 +136,8 @@ export const createResolvers = ({ createResolvers }) => {
             ? documentToHtmlString(JSON.parse(source.introRichText.raw))
             : '';
 
+          // console.log(inspect(source, true, 10, true));
```
feedm3: 🧹

src/components/content/rss/rss-render-options.ts
```diff
@@ -0,0 +1,137 @@
...
+
+enum EmbeddedAssetType {
+  IMAGE = 'image',
+  VIDEO = 'video',
+}
+
+const converter: Converter = new Converter();
+const markdownToHyperText = (markdown: string) => converter.makeHtml(markdown);
+
+export const rssRenderOptions: Partial<Options> = {
```
feedm3: I'm not so much into the rendering topic, can you explain why we need a separate renderer for the content blocks? Also add the explanation as comment for devs that need to work on that in the future.

Is there some way of sharing code with the rich-text renderer? How can we make sure that developers don't forget that there are now 2 places which need to be edited when a change in Contentful happens (one idea: change the file structure and make sure both files are very close to each other)?

src/components/content/rss/rss-render-options.ts
```diff
@@ -0,0 +1,137 @@
...
+import { ContentfulCustomModel } from '../../../types/contentful.types';
+import { Converter } from 'showdown';
+
+enum EmbeddedAssetType {
+  IMAGE = 'image',
+  VIDEO = 'video',
+}
+
+const converter: Converter = new Converter();
+const markdownToHyperText = (markdown: string) => converter.makeHtml(markdown);
```
feedm3: Why does our markdown converter from rich-text not work here? It should do the same.

gatsby-config.ts
```diff
@@ -164,6 +166,51 @@ const gatsbyConfig: GatsbyConfig = {
                     <img src="${imageUrl}" alt="">
                   `;
 
+                // combine intro text and copy/body text
```
feedm3: There is now a lot of code in the file. Can you try to extract it and put it somewhere under `/gatsby`?

feedm3 (changes requested):
Thanks for the contribution!

Whenever the change gets quite large (like in this case) it's cool if you add a bit more commentary and explanation about:

- what is the problem
- why did you go for the solution
- are there some parts in the PR that need a closer look

Also, it helps a lot to get into context if there is a screenshot of the "bug" and the solution..

.husky/pre-commit
timonsommer: My local husky configuration had some issues finding the `yarn` command, so I used `npm` instead. Ultimately, this change was pushed by mistake.

src/components/content/rss/rss-render-options.ts
timonsommer: We introduced a dedicated renderer for the RSS XML generation, as the rich-text renderer returns custom styled components as JSX Elements as opposed to the HTML string that the Gatsby feed plugin requires. I've tried reusing the code from `rich-text.tsx` by employing `ReactDOMServer.renderToStaticMarkup(...)`. However, this lead to build problems that I haven’t been able to resolve yet.

src/components/content/rss/rss-render-options.ts
timonsommer: I ran into issues with `parcel` when I tried to use the existing HTML AST solution. @dschu-lab might know more about this, as he suggested using a different converter in the first place. I will ask him when he's back.

gatsby-config.ts
timonsommer: I did some refactoring in 8c7d6729bbeb6f8850c899647498becb108eac0c to remedy this.

gatsby/config-options/feed-options.tsx
```diff
@@ -0,0 +1,382 @@
...
+} from './config-constants';
+import * as ReactDOMServer from 'react-dom/server';
+import React from 'react';
+import { BLOCKS } from '@contentful/rich-text-types';
+import {
+  ContentfulCustomModel,
+  EmbeddedAssetType,
+} from '../../src/types/contentful.types';
+import { Converter } from 'showdown';
+import { MarkdownAst } from '../../src/components/legacy/markdown/markdown-ast';
```
feedm3: Unused import

feedm3 (changes requested):
Nice, already got a lot better :)  Thanks!

Some smaller stuff I have found:
- `package-lock.json` should be removed, it's only used by npm and we stick to yarn
- `feed-options` should be renamed to `gatsby-plugin-feed-options` to make it more obvious what is meant
- `config-constants` can be called just `constants`, as this is already in the config folder

feedm3 (approved):

PR #670, opened 2022-10-13
fix(rss): remove duplicates in rss feed

JulianAnthes:
We have to filter by `node_locale` so that we only get the blog post entries once.

### DIFF
--- a/gatsby/config-options/gatsby-plugin-feed-options.tsx
+++ b/gatsby/config-options/gatsby-plugin-feed-options.tsx
@@ -2,19 +2,18 @@ import {
   documentToHtmlString,
   Options,
 } from '@contentful/rich-text-html-renderer';
+import { BLOCKS } from '@contentful/rich-text-types';
+import * as ReactDOMServer from 'react-dom/server';
+import { Converter } from 'showdown';
+import {
+  ContentfulCustomModel,
+  EmbeddedAssetType,
+} from '../../src/types/contentful.types';
 import {
   BASE_URL,
   DEFAULT_META_IMAGE_URL_PATH,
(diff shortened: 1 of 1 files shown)


gatsby/config-options/gatsby-plugin-feed-options.tsx
```diff
@@ -2,19 +2,18 @@ import {
...
 import {
   ContentfulCustomModel,
   EmbeddedAssetType,
 } from '../../src/types/contentful.types';
-import { Converter } from 'showdown';
+import {
+  BASE_URL,
+  DEFAULT_META_IMAGE_URL_PATH,
+  RSS_FEED_URL,
+} from './constants';
```
JulianAnthes, line 17: This part changed due to "optimizing imports" which runs once the file is saved.
It was not intended to do this, but it doesn't hurt so I will not revert that. 🤠

gatsby/config-options/gatsby-plugin-feed-options.tsx
```diff
@@ -244,7 +243,10 @@ const feedOptions = {
       },
       query: `
           {
-            allContentfulBlogPost(sort: { order: DESC, fields: [publicationDate] }) {
+            allContentfulBlogPost(
+              filter: { node_locale: { eq: "en" } }
+              sort: { order: DESC, fields: [publicationDate] }
+            ) {
```
JulianAnthes, line 249: This is the actual fix. We filter by `node_locale` to get rid of the duplicates.

georgiee (approved):
Thanks. I did not check the output but the code itself looks convincing with the additional filter.

PR #675, opened 2022-10-28
fix: condense blog post layout

timonsommer:
Changes were made to the overall design of blog post pages, as requested by @esreverse. 

These include:
- reduce margins of headline elements 
- reduce margins of `<li>` elements in both `<ul>` and `<ol>`
- prevent lower resolution images from being scaled to fit content width


src/components/content/rich-text/rich-text-query.ts
```diff
@@ -22,7 +22,7 @@ export const ContentfulRichText = graphql`
           contentType
           url
         }
-        gatsbyImageData(width: 1440)
+        gatsbyImageData(layout: CONSTRAINED)
```
milljoniaer, line 25: What did you add this, what is it doing exactly?

milljoniaer (commented):
Thank you for the PR. I added one question about the rich-text query.

And did you change any dependencies?

As far as I can see, there has been aa `package-lock.json` added. Afaik this is added when you update dependencies with `npm` instead of `yarn`. 

Is it on purpose that this file is on the PR or should it be deleted?

Otherwise I would say LGTM 👍

src/components/content/rich-text/rich-text-query.ts
timonsommer: This option instructs the Gatsby Image plugin not to scale images beyond their resolution, so they won't appear overly blurry. It still respects the enclosing container though, thus images are never rendered wider than the rich-text content.

milljoniaer: ah okay, thank you 👍

milljoniaer (approved):

PR #676, opened 2022-10-28
fix: undefined check for asset

milljoniaer:
### Description

This PR fixes the build error:
image

Somehow this error only accurs with the images of the new blog article by eric and not the other ones. But this fixes the build error. Also the main build fails currently because of this.

### DIFF
--- a/gatsby/config-options/gatsby-plugin-feed-options.tsx
+++ b/gatsby/config-options/gatsby-plugin-feed-options.tsx
@@ -106,6 +106,11 @@ export const rssRenderOptions: Partial<Options> = {
       }
     },
     [BLOCKS.EMBEDDED_ASSET]: (node) => {
+      if (!node.data.target.file) {
+        console.log(JSON.stringify(node));
+        return '';
+      }
+
       const { url, contentType } = node.data.target.file;
       switch (contentType.split('/')[0]) {
         /**
--- a/gatsby/config-options/gatsby-plugin-feed-options.tsx
+++ b/gatsby/config-options/gatsby-plugin-feed-options.tsx
@@ -108,7 +108,7 @@ export const rssRenderOptions: Partial<Options> = {
     [BLOCKS.EMBEDDED_ASSET]: (node) => {
       if (!node.data.target.file) {
         console.log(JSON.stringify(node));
-        return '';
+        //return '';
       }
 
       const { url, contentType } = node.data.target.file;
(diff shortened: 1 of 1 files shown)


gatsby/config-options/gatsby-plugin-feed-options.tsx
```diff
@@ -106,6 +106,11 @@ export const rssRenderOptions: Partial<Options> = {
       }
     },
     [BLOCKS.EMBEDDED_ASSET]: (node) => {
+      if (!node.data.target.file) {
+        console.log(JSON.stringify(node));
```
georgiee, line 110: well maybe that's a little bit too debugging? Could you log a message and maybe mention some other details.
Or alternative to get things going: Remove the console .log
Or alternative 2: I know that you're working on this topic anyway so let it be  😅

georgiee (approved):
Thanks for the quick fix.

gatsby/config-options/gatsby-plugin-feed-options.tsx
milljoniaer: Thank you for your review, I added a fix for the original problem which should be a better way of fixing the build

gatsby/config-options/gatsby-plugin-feed-options.tsx
```diff
@@ -197,26 +231,8 @@ const feedOptions = {
           if (introRaw.content)
             rawContent.content = [...introRaw.content, ...copyRaw.content];
 
-          const contentWithEmbeddedEntries = rawContent.content.map(
```
georgiee, line 200: I see that you refactored this part into a dedicated function. That makes it difficult to find the actual change. What did you fix aside from the refactoring? What was the issue? Can you highlight it on the counter part if this removal

milljoniaer: On all the other types than images, (so paragraphs, lists and so on) i did not only return the content as it was in there previous version, but checked if it has children (content.content), and then call the function recursively on those children.

In that way we get all images, even if they are very cascadenated.

georgiee: nice, thanks for the clarification!

gatsby/config-options/gatsby-plugin-feed-options.tsx
```diff
@@ -152,6 +157,35 @@ export const rssRenderOptions: Partial<Options> = {
...
+  // Skip everything but entry and asset blocks
+  if (
+    content.nodeType !== 'embedded-entry-block' &&
+    content.nodeType !== 'embedded-asset-block'
+  ) {
+    if (content.content) {
+      return {
+        ...content,
+        content: content.content.map((el) =>
+          addImageDataToLinks({ content: el, node }),
```
georgiee, line 170: power to the recursion. thanks for handling this case so easily 👍

georgiee (approved):
here we go, my second approval 🙌 Great how you solved it.
I did not run any code and just checked the code on the surface combined with your given feedback. Thank you.

PR #679, opened 2022-11-14
chore: expect node 18

hofmann-felix:
* The `dependency-update` action failed with this error message:
```error gatsby@5.0.1: The engine "node" is incompatible with this module. Expected version ">=18.0.0". Got "16.18.0"```
(https://github.com/satellytes/satellytes.com/actions/runs/3458079706/jobs/5772136982)
* We now expect node 18 in the `package.json` so the action should also be executed with node 18

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16'
+          node-version: '18'
           cache: 'yarn'
       - name: Install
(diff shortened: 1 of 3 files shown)


fischeversenker:
The change itself looks fine :dizzy:
I don't have access to the failed Gatbsy Build Logs, can you take a look at what's wrong there?

feedm3:
`NODE_VERSION=18` env variable was missing in Gatsby Cloud. Unfortunately, they don't take it from the `package.json`. I added it, build is green now. 

Will merge it, thanks for the PR :)

feedm3 (approved):

PR #682, opened 2022-11-22
feat: add field applicantLocationRequirements to job data

milljoniaer:
Google needs a mandatory field "applicantLocationRequirements" which was missing in our Job Postings structured Data. 
See [Google Search Console](https://search.google.com/search-console/r/jobs?resource_id=sc-domain%3Asatellytes.com) for reference.

I added the field with the value "EU" as Location Requirement. Validate the Markup by using [googles validator](https://search.google.com/test/rich-results)

### DIFF
--- a/src/components/pages/career-details/career-details-structured-data.tsx
+++ b/src/components/pages/career-details/career-details-structured-data.tsx
@@ -71,6 +71,10 @@ export const CareerDetailsStructuredData = ({
       },
     },
     jobLocationType: 'TELECOMMUTE',
+    applicantLocationRequirements: {
+      '@type': 'Country',
+      name: 'EU',
+    },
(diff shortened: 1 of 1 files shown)


feedm3 (approved):

georgiee:
Yeah thank you for finding and fixing it.
While I think the EU value is okay I wonder if we want to limit this value to Germany (is `EU` a valid `country` in this aspect anyway?).

The field is defined as:
> The location(s) applicants can apply from. This is usually used for telecommuting jobs where the applicant does not need to be in a physical office. Note: This should not be used for citizenship or work visa requirements.

via 
https://schema.org/applicantLocationRequirements

I understand this as "accepted remote working location" where we currently do not officially support anything outside Germany (tax related)

PR #690, opened 2022-12-29
fix: footer won't stick to bottom of the page

### LINKED ISSUE #688: NotAvailableInGerman page: the page doesn't fill the viewport in some formats
### What is the problem?
* The `NotAvailableInGerman` page (https://satellytes.com/de/blog/) on the blog does not completely fill the viewport in some mobile formats. This results in a white bar below the navigation. 
* Is rather an edge case, but when you use the browser only on one half of the screen, this bug occurs

### Preview
* Format: 740 x 1126
Bildschirm­foto 2022-12-28 um 10 00 19

timonsommer:
The footer now stays at the bottom of the page at all times.

### DIFF
--- a/src/components/layout/layout.tsx
+++ b/src/components/layout/layout.tsx
@@ -21,6 +21,7 @@ import { useAnchorTagScrolling } from './use-anchor-tag-scrolling';
 const FullHeightContainer = styled.div`
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


feedm3:
Closes https://github.com/satellytes/satellytes.com/issues/688

feedm3 (approved):

PR #695, opened 2023-01-16
Chore/change career ordering

alcocere:
I have done some changes following the Figma new design for the Career page:
- The order from openings and application process blocks had changed. 
- I have also added two image spacer between content blocks.
- And finally I have also changed the size of the title and text from application process. 
For this last part I have reused the <Teaser> component that had the same properties I needed.  And I have deleted the <ContentBlockContainer> that was wrapping the <ApplicationProcess> because I needed to have a different margin top.


src/components/pages/career/career-page.tsx
```diff
@@ -47,18 +62,20 @@ export const CareerPage = ({ positions, heroImageData }: CareerPageProps) => {
         </SectionHeader>
       </ContentBlockContainer>
 
+      <ApplicationProcess />
```
hofmann-felix: I would put the `ApllicationProcess` inside the first `ContentBlockContainer`, since the contents belong together (and they are below the same `SectionHeader` on the page)
```suggestion
        </SectionHeader>
        <ApplicationProcess />
      </ContentBlockContainer>

```

src/components/pages/career/application-process.tsx
```diff
@@ -24,13 +24,17 @@ const AccordionText = styled.p`
   margin-bottom: -3px;
 `;
 
+const ApplicationWrapper = styled.div`
```
hofmann-felix: I would name it `ApplicationProcessWrapper` so it doesn't sound ambiguous

src/components/pages/career/career-page.tsx
```diff
@@ -11,13 +11,28 @@ import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { SyPersonioJob } from '../../../types';
 import { ImageHero } from '../../content/heroes';
 import { IGatsbyImageData } from 'gatsby-plugin-image';
+import { ImageSpacer } from '../../pages/landingpage/image-spacer';
```
hofmann-felix: Because the `ImageSpacer` is now used not only on the landing page, it makes sense to move the `ImageSpacer` component into the `components/ui` folder

src/components/pages/career/career-page.tsx
```diff
@@ -11,13 +11,28 @@ import { useTranslation } from 'gatsby-plugin-react-i18next';
...
+  officeImages: OfficeImages;
 }
 
-export const CareerPage = ({ positions, heroImageData }: CareerPageProps) => {
+interface OfficeImage {
+  relativePath: string;
+  childImageSharp: {
+    gatsbyImageData: IGatsbyImageData;
+  };
+}
```
hofmann-felix: You created exactly the same interface in the parent. In my opinion it makes sense to delete this interface and import the one from the parent

src/pages/career.tsx
```diff
@@ -6,16 +6,31 @@ import { CareerPage } from '../components/pages/career/career-page';
...
 }
 
 const Career = (props: PageProps<CareerPageQueryProps>) => {
   const { t } = useTranslation();
 
+  const officeImages = props.data.officeImages.nodes.reduce((memo, image) => {
+    memo[image.relativePath] = image;
+    return memo;
+  }, {});
+
```
hofmann-felix, line 27: Nice idea 🙌 This makes it very easy to access the images in the `CareerPage` component

hofmann-felix (changes requested):
First of all, congratulations on your first satellytes.com PR 🥳 I think the changes are very good, nevertheless I have added a few small suggestions for improvement

Regarding your PR description: It helps the reviewers a lot if you add a screenshot of the changes (especially for pure design PRs). Then you can see directly which pages/components are involved

src/components/pages/career/career-page.tsx
alcocere: Hi Felix, thank you for all your suggestions and feedback. 
I have just updated this part and put the `ApllicationProcess` inside the `ContentBlockContainer` from the `SectionHeader`

src/components/pages/career/application-process.tsx
alcocere: great suggestion! Thanks, I have changed it too.

src/components/pages/career/career-page.tsx
alcocere: you are right, that makes more sense. Thanks for the suggestion , I have changed it.

src/components/pages/career/career-page.tsx
alcocere: Yes, it makes totally sense as you suggested. Thanks I have changed it too.

src/pages/career.tsx
```diff
@@ -6,16 +6,31 @@ import { CareerPage } from '../components/pages/career/career-page';
 import { SyPersonioJob } from '../types';
 import { IGatsbyImageData } from 'gatsby-plugin-image';
 
+export interface OfficeImage {
```
milljoniaer: This type already exists in the landingpage. I would either import it from there or maybe create a `src/types/types.ts` and import it from there in both landingpage and career.

alcocere: Thanks for the suggestion. I have finally import it from `src/pages/index`

hofmann-felix (approved):
@alcocere Thanks for implementing the suggestions 🙏 Looks good!

PR #706, opened 2023-03-01
Feat: changing the spacing and size of components for an updated design

dastfs:
### What's included?
* Spacing changes
* Margin and padding changes
* GridItem parameter change

### Description
The main goal is to make the container for the content wider and adjust all other elements on the page to it.

Previous | Updated
-- | --
Screenshot 2023-03-01 at 16 06 09 | Screenshot 2023-03-01 at 16 05 55
Screenshot 2023-03-01 at 16 09 23 | Screenshot 2023-03-01 at 16 09 37
Screenshot 2023-03-01 at 16 08 53 | Screenshot 2023-03-01 at 16 08 34


ghost:
## :white_check_mark: satellytes.com-feat-main-design-update deploy preview ready
* [Deploy preview](https://satellytescomfeatmaindesignupd-designupdatesizeandspacing.gatsbyjs.io)
* [Build logs](https://gatsbyjs.com/dashboard/f215130d-15e3-4176-9252-e13166fe48de/sites/6edb236d-9b25-45e0-8a0d-5e044af7c366/builds/d789596a-3d22-461b-afb7-94945854bf22/details) · **1m** build time
* Performance **72** :orange_circle: · Accessibility **97** :green_heart: · Best Practices **100** :green_heart: · SEO **86** :orange_circle: — [View full report](https://build-d789596a-3d22-461b-afb7-94945854bf22.gtsb.io/reports/lighthouse/index.html)

src/components/pages/career/application-process.tsx
```diff
@@ -25,7 +25,11 @@ const AccordionText = styled.p`
 `;
 
 const ApplicationProcessWrapper = styled.div`
-  margin-top: 80px;
+  margin-top: 48px;
+
+  ${up('md')} {
+    margin-bottom: 60px;
```
hofmann-felix:
```suggestion
    margin-top: 60px;
```

src/components/pages/blog/posts.tsx
```diff
@@ -4,10 +4,16 @@ import styled from 'styled-components';
...
 import { LONG_DATE_FORMAT, useLocaleFormat } from '../../i18n-helpers';
+import { up } from '../../support/breakpoint';
 
 const BlogTeaserGrid = styled.div`
   display: grid;
-  gap: 24px;
+  gap: 48px;
+
+  ${up('md')} {
+    gap: 24px;
```
hofmann-felix:
```suggestion
    row-gap: 60px;
    column-gap: 24px;
```

The gaps on desktop don't have the same values in the design

src/components/layout/theme.tsx
```diff
@@ -2,7 +2,8 @@ import { DefaultTheme } from 'styled-components';
 
 export const HEADER_HEIGHT_VALUE = 65;
 export const CONTENT_SPACE_SMALL = 80;
-export const CONTENT_SPACE_LARGE = 180;
+export const CONTENT_SPACE_LARGE = 120;
+export const CONTENT_SPACE_LEFT_RIGHT = 76;
```
hofmann-felix: I think it makes more sense to remove the left and right padding (= `CONTENT_SPACE_LEFT_RIGHT`) and use the maximum content width instead of the 1400px. According to Figma this is currently 1184px. Since the value seems a bit random to me, I would definitely clarify this again with the design team.

src/components/layout/navigation/navigation.tsx
```diff
@@ -230,7 +230,7 @@ const Navigation: React.FC<NavigationProps> = ({
     <NavigationBackground>
       <Grid center>
         <GridItem xs={0} md={1} />
-        <GridItem xs={12} md={10} noGap>
+        <GridItem xs={12} md={11} noGap>
```
hofmann-felix: Bildschirm­foto 2023-03-01 um 17 19 09

There is something wrong with the alignment. Maybe it makes sense to move the navigation changes to an extra PR? I can not estimate how complex they are

src/components/legacy/grid/grid.tsx
```diff
@@ -25,7 +25,7 @@ export const Grid = styled.div<GridProps>`
   padding: 0
     ${(props) => (props.disableGap || props.nested ? 0 : GRID_GAP_MOBILE)}; // outer gap
 
-  max-width: ${(props) => props.theme.maxWidth};
+  // max-width: ${(props) => props.theme.maxWidth};
```
hofmann-felix: Is that commented out on purpose?

hofmann-felix (changes requested):
Thanks for your changes. The page definitely looks much cleaner now ✨ I have added a few more suggestions for improvement

Btw: The markdown table in the PR description is really handy. I'll definitely keep that in mind for my future PRs.

src/components/pages/career/application-process.tsx
dastfs: 🤦 thanks

src/components/pages/blog/posts.tsx
dastfs: fixed it

src/components/layout/theme.tsx
dastfs: I though the box which is selected on the screen is our content-block-container

image

then I found that child element of this container (text-wrapper) is 1248

image

then I did simple calculation (1400-1248)/2 = 76px to find the padding value 
Now I understand that all these calculations were made with the originally wrong approach :)

And yes, let us clarify and do it in a right way

dastfs: I though the box which is selected on the screen is our content-block-container
image
And it should be 108 px not 76px. I came to 76 using calculation with the originally wrong approach :)

But anyway, let us clarify and do it in a right way.

src/components/layout/navigation/navigation.tsx
dastfs: Agreed, it is better to do an extra PR, otherwise it is a partial solution.

src/components/legacy/grid/grid.tsx
dastfs: it should be deleted, but it is also for the navigation bar, since we decided to make it in extra PR, I will revert this change

src/components/layout/theme.tsx
dastfs: @hofmann-felix I learned from the design team that the container's width should actually be 1400px and the content inside is 1184px. Since there is no need for a container (and it doesn't exist in the project), I set the content-block width to 1184px and removed paddings.

hofmann-felix (approved):
Thanks for implementing the changes 🙏 Looks good!

PR #708, opened 2023-03-02
feat: add new hero style

hofmann-felix:
### What's included?
* Background color is added to `<HeroContainer />`
* Hero text color is adjusted
* `<ImageHero />` and `<Aurora />` component are removed
* The `kicker` field of the hero is removed since it isn't needed anymore
* No typography changes since there is a open PR for it

### Preview
* URL: https://satellytescomfeatmaindesignupd-featherobackground.gatsbyjs.io/
Bildschirm­foto 2023-03-02 um 15 41 41


hofmann-felix:
There is still the `Aurora` component with its corresponding child components. I'm not quite sure if the components should be deleted or if we want to keep those. It would be nice to here some opinions about that. Currently they don't have any usage

src/components/content/heroes/heroes.stories.mdx
```diff
@@ -1,63 +1,8 @@
-import { Canvas, Meta } from '@storybook/addon-docs';
-import { AuroraHero } from './aurora-hero'; import { AuroraType } from './aurora/aurora';
+import { Meta } from '@storybook/addon-docs';
 
 <Meta
   title="Components/Content/Heroes" />
 
-# Aurora Hero
```
georgiee, line 7: In hindsight we had good ambitions creating storybook content for our website but I doubt folks had using or maintained it properly 😬  Maybe we can remove storybook content entirely over time

georgiee (changes requested):
The technical changes are looking good. I really don't agree with the decision to introduce that grey in grey hero. It's dull, it's not interesting. I would like to keep the Aurora effect at least on the landingpage. On the other side, you already implementing it is a sign that I'm too late giving that feedback.

What's your opinion on that hero issue?
I would like to hold it back for a short while just to ask the question internally once 🙏

hofmann-felix:
@georgiee I've also talked to Gholam and Dastan about it and we're all not big fans of the new background either. I'll switch the PR to draft until a design decision is made here.

hofmann-felix:
After the requirements have changed and the background does not change at all, I am closing this PR. For the text/size changes I will create a new PR.

PR #709, opened 2023-03-02
feat: New illustrations, image and some content

dastfs:
### What's included?
* New or changed illustrations of the updated design
* New section in service area of landing page and in service page
* Hiding illustrations in mobile version in service page (see screenshot)

image


src/assets/illustrations/052-bike.svg
```diff
@@ -0,0 +1,19 @@
+<svg viewBox="0 0 80 80" fill="none" xmlns="http://www.w3.org/2000/svg">
```
hofmann-felix, line 1: The illustration SVGs should be consistent, which means that they should all have the same `viewbox`. Currently there are different variants (e.g. `viewBox="0 0 80 80"` and `viewBox="0 0 144 144"`).

How I would import the illustrations:
1. Delete all existing illustrations in the project
2. Go to the illustrations overview in Figma: https://www.figma.com/file/FWlI63N28LP7BxqNqj1bGw/sy-004-(Design!)?node-id=5310%3A20878&t=OfCV4q3sqkTirQzm-4
3. Select all illustrations at once and export them (s. screenshot)
4. Remove `width="144"` and `height="144"` in the first line of each svg file
5. Adjust `illustration-files.tsx`
6. Add the illustrations to the page

Bildschirm­foto 2023-03-02 um 17 08 31

src/components/ui/image/image-spacer.tsx
```diff
@@ -7,11 +7,11 @@ const ImageContainer = styled.div`
...
-  height: 210px;
+  height: 240px;
   margin-top: 80px;
 
   ${up('md')} {
-    height: 360px;
+    height: 620px;
     margin-top: 180px;
   }
 `;
```
hofmann-felix: I think the image spacer it to high in some formats (screenshot is from a 13" MacBook) but I am not a designer 😅 I think it makes sense to ask the design team if a third breakpoint with a third height (between 240 and 620) for the image spacer would be helpful

Bildschirm­foto 2023-03-02 um 17 29 05

src/components/pages/service/service.tsx
```diff
@@ -5,10 +5,10 @@ import { SectionHeader } from '../../content/section-header/section-header';
 import { LeadboxProps } from '../../content/leadbox/leadbox';
 import { useTranslation } from 'gatsby-plugin-react-i18next';
 import { Consulting } from './consulting';
-import { Industries } from './industries';
```
hofmann-felix, line 8: You can delete `industries.tsx` if it isn’t used in other places

src/components/pages/service/product-design.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <ContentBlockContainer>
+      <Intro
+        illustration="eclipse_039"
+        headline={t('services.product_design.title')}
+      >
+        <p>{t('services.product_design.text')}</p>
+      </Intro>
+    </ContentBlockContainer>
+  );
+};
```
hofmann-felix: Even though the component is very small, I like that you put it in an extra file. 
This way everything is well structured 🙌

hofmann-felix (changes requested):
In this PR illustrations are added and removed in some components, the ImageSpacer is adjusted and the content of the service page is changed. These are three different changes that have nothing to do with each other. Therefore, a division into three different PRs would make sense in any case. Since the scope of the changes is rather small, I would leave it as it is and only do it differently in the future.

src/assets/illustrations/052-bike.svg
dastfs: When I looked before the pull request, the component library hadn't been updated yet, so I used the ones in the page designs. Of course, it is much more convenient to export all the icons at once.

hofmann-felix (approved):
Thanks for implementing the feedback. I approve the PR, but we definitely need a follow up PR, which:
* safely deletes all existing illustrations
* adds all new illustrations to the project ([like this](https://github.com/satellytes/satellytes.com/pull/709#discussion_r1123361671))
* adds the new illustrations to the corresponding places in the code 

Also, we should clarify the [image spacer issue](https://github.com/satellytes/satellytes.com/pull/709#discussion_r1123393572) with the design team before going live

PR #711, opened 2023-03-03
feat: Add Kicker to services

dastfs:
### What's included?
Kicker was added to services in service page

Previous | Updated
-- | --
Screenshot 2023-03-03 at 12 59 58 | Screenshot 2023-03-03 at 12 59 46


src/components/pages/service/support.tsx
```diff
@@ -36,16 +36,30 @@ const ContentStyled = styled.div`
   ${TextStyles.textR}
 `;
 
-export const Intro = ({ illustration, headline, children }: any) => {
+const Kicker = styled.span`
+  ${TextStyles.toplineS}
+  display: block;
+  margin-bottom: 12px;
+  color: #3e61ee;
```
hofmann-felix: I know there are many other places in the code where we hardcode the colors. But in general it is better to use the colors from the theme file. If we want to change them later, it will be much faster:
`color: ${theme.palette.text.topline};`

src/components/pages/service/support.tsx
```diff
@@ -36,16 +36,30 @@ const ContentStyled = styled.div`
...
-      </div>
-    </IntroLayout>
+    <>
+      {kicker && <Kicker>{kicker}</Kicker>}
+      <IntroLayout>
+        <IllustrationStyled size={IllustrationSize.LARGE} show={illustration} />
+        <div>
+          <HeadlineStyledWithAnchor>{headline}</HeadlineStyledWithAnchor>
+          <ContentStyled>{children}</ContentStyled>
+        </div>
```
hofmann-felix: I would place the kicker in this `div` for two reasons: 
* Kicker, headline and content belong together semantically and should therefore also be together in the code
* Currently the design looks broken on mobile because of the illustration. I know that the illustration will be hidden on mobile. But if we show it again in the future, it would work that way directly.

hofmann-felix (changes requested):
Thanks again for the screenshots in the PR description. Through the before/after comparison you can see as a reviewer at first glance what has been changed 💪

The changes look good too, I added two small suggestions for improvement

src/components/pages/service/support.tsx
dastfs: good point, let's put it in div

src/components/pages/service/support.tsx
dastfs: i completely agree with you, the use of theme's constants is much better

hofmann-felix (approved):
Nice thank you for the quick implementation of the changes

PR #712, opened 2023-03-03
feat: add customer logos

hofmann-felix:
### What's included?
* Customer logo SVGs are added to the project locally
* `<Customers />`component which shows and aligns the logos on the landingpage 

### Preview
* URL: https://satellytescomfeatmaindesignupd-featcustomerlogos66676.gatsbyjs.io/
* Mobile: 
Bildschirm­foto 2023-03-03 um 13 34 08

* Desktop:
Bildschirm­foto 2023-03-03 um 13 34 38


src/pages/index.tsx
```diff
@@ -100,6 +110,21 @@ export const IndexPageQuery = graphql`
       }
     }
 
+    allContentfulCustomerLogos(
```
georgiee: So did you create a new contentful model? That should be mentioned in the PR
ideally with some notes about how to manage them (or at least a screenshot?)

src/components/pages/landingpage/customers.tsx
```diff
@@ -0,0 +1,72 @@
...
+        {Logos[3]}
+        {Logos[4]}
+      </LogoLine>
+      <LogoLine>
+        {Logos[5]}
+        {Logos[6]}
+        {Logos[7]}
+      </LogoLine>
+    </LogosContainer>
+  );
```
georgiee: I see that there is a contentful model for logos, but on the other side you rely on the first 8 logos.
+ What happens if I delete one?
+ All logos beyond the count of 8 are ignored

I wonder, wouldn't it be more pragmatic not to adjust the CMS and just display a set of 8 logos locally?
Wouldn't it even easier to display a set of 1 or 2 svg files that contains all logos already, especially because everything is so static here?

georgiee (changes requested):
Thanks for this, looks great 👍

While I really want to see our clients on the website I would like to discuss your approach. Can you check my comments? Thanks.

hofmann-felix:
@georgiee I agree with you that the CMS doesn't offer any advantages for the logos and that's why I added them locally. I have reverted the changes in Contentful.

georgiee:
Thanks for the changes and keeping it simple. 
There is one level of simplicity left and I would like to ask if this is by purpose.

This would suffice here:
```
<img src="clients.svg">
```

Because our client list will not change very often I think we don't need the individual svgs. A single svg would be enough and makes it even more simple. What do you think?

hofmann-felix:
@georgiee In the current implementation, the number of rows adjusts to the viewport (s. video). If this behavior is not necessary and a fixed number of rows is sufficient, I agree with your suggestion. This would make the implementation much simpler. 

https://user-images.githubusercontent.com/57712895/223400564-615e1471-41f3-47b9-9ce2-c2a48b34a61e.mov

georgiee (approved):

georgiee:
Thanks for the confirmation. I see the aspects why you chose to have distinct svg files per client. Let's go with this solution. It's always easy to remove code once we realize we don't need that. Thanks.

hofmann-felix:
@georgiee I also think there may be a better solution. I'll merge the PR, maybe someone has time for a refactoring later. Thanks for your thoughts and review on this 🙏

PR #714, opened 2023-03-06
feat: new hero style

hofmann-felix:
### What's included?
* Some minor adjustments in the hero component:
  * `kicker` prop is removed since it isn't used anymore
  * Desktop font-size is adjusted
* Spacing was already adjusted in https://github.com/satellytes/satellytes.com/pull/706

### Preview
* https://satellytescomfeatmaindesignupd-featherostyle.gatsbyjs.io/

### DIFF
--- a/src/assets/locales/de/translations.json
+++ b/src/assets/locales/de/translations.json
@@ -108,7 +108,6 @@
   "career.leadbox.mail": "career@satellytes.com",
   "about-us.title": "Über uns",
   "about-us.office.title": "Office",
-  "about-us.office.kicker": "Satellytes",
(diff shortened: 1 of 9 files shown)


georgiee (approved):
Thanks for changes.
Screenshot 2023-03-07 at 10 58 12


Will we adjust the height/spaces ? Or solve it by text?
Do you know that already?

Technically it's okay but I don't understand the plan. Less content and more space is not what we aim for with the current refactoring.

hofmann-felix:
Bildschirm­foto 2023-03-07 um 11 06 11

@georgiee According to the design, there will be a description on each page. Then the higher spaces should also make sense

PR #715, opened 2023-03-06
change number of posts per page to 8

dastfs:
### What's included?
* number of posts per page is changed to 8 (desktop version): 
- 2 rows by 4 posts (see screenshot)

image

### DIFF
--- a/src/components/pages/blog/posts.tsx
+++ b/src/components/pages/blog/posts.tsx
@@ -19,7 +19,7 @@ const BlogTeaserGrid = styled.div`
   // we fetch teaser of size 600px
   // this means we can search for a column size of 300px - 12px (half gap) = 288px
   // to auto fit our teasers
-  grid-template-columns: repeat(auto-fill, minmax(288px, 1fr));
+  grid-template-columns: repeat(auto-fill, minmax(278px, 1fr));
 `;
 
 interface PostsProps {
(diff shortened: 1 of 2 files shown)


src/components/pages/blog/posts.tsx
```diff
@@ -19,7 +19,7 @@ const BlogTeaserGrid = styled.div`
   // we fetch teaser of size 600px
   // this means we can search for a column size of 300px - 12px (half gap) = 288px
   // to auto fit our teasers
-  grid-template-columns: repeat(auto-fill, minmax(288px, 1fr));
+  grid-template-columns: repeat(auto-fill, minmax(278px, 1fr));
```
dastfs, line 22: Hey @georgiee,
I just wanted to ask about the comments here.
Here is the grid to store blog posts, and to fit them automatically, it uses a fixed width (288px). I did the same thing, just used a smaller value (278px) from the Figma layout to accommodate 4 articles in a one row.
In your comment you determined this value by calculation, but I did not understand the meaning of the calculation.
So I am not sure if my solution works. 
Is it ok if I just use the value from Figma?

georgiee: Oh I can't even remember I wrote this 😅 I think whatever works is good enough here. And if you can now fit 4 columns in a row then it's a good choice. When you have some spare time you can also try to understand the underlying auto fill algorithm and maybe answer it yourself by combining that knowledge with the figma design. Thanks for asking 👍

hofmann-felix (approved):
The blog overview looks much clearer now 👍

PR #717, opened 2023-03-07
feat(LandingPage): view four teasers

hofmann-felix:
### What's included? 
* 4 teasers are now always displayed on the landing page
* Global `TeaserGrid` component is removed since it was only used by the landing page
* minor spacing adjustments 

### Preview
* URL: https://satellytescomfeatmaindesignupd-featlandingpagedesign.gatsbyjs.io/
* Career Teasers on landing page(Blog & Service teasers look similar):
Bildschirm­foto 2023-03-07 um 14 41 21


src/components/pages/landingpage/service.tsx
```diff
@@ -45,9 +40,14 @@ export const Service = () => {
...
-
-      <Spacer />
+        <Teaser
+          title={t('services.design.title')}
+          linkTo="/services#design"
+          illustration="scientist_042"
+        >
+          {t('services.design.teaser')}
+        </Teaser>
+      </LandingPageTeaserGrid>
```
hofmann-felix: Correct content will be added in a later PR

src/components/pages/landingpage/blog.tsx
```diff
@@ -31,7 +38,7 @@ export const Blog = ({ posts }: BlogProps) => {
         {t('main.blog.text')}
       </HomePageHeaderBlock>
 
-      <TeaserGrid>
+      <StyledLandingPageTeaserGrid>
```
georgiee, line 41: Thanks for making this component local 🙏

georgiee (approved):
Nice, thanks!

One thing, well an edge case thing:

If there are no more than 1 item the teaser does expand to the full width. The blog posts will always have more than four but the jobs might get reduced as some point. Maybe worth to fix it here on in a following PR?

Screenshot 2023-03-07 at 15 22 53



Visually and technically good though so it's approved!

PR #721, opened 2023-03-08
chore: adjust spacing

hofmann-felix:
### What's included? 
* Spacing adjustments in different components
* Main change: The combination of kicker, headline and content now always has the same spacing (just like in Figma)
* Preview URL: https://satellytescomfeatmaindesignupd-chorespacingcleanup.gatsbyjs.io/

### DIFF
--- a/src/components/content/header-block/header-block.tsx
+++ b/src/components/content/header-block/header-block.tsx
@@ -50,14 +50,14 @@ const Topline = styled.p`
   }
 `;
 
-const Headline = styled.h2<LargeProps>`
+const Headline = styled.h2`
   ${TextStyles.headlineM}
   letter-spacing: -0.01em;
   color: #202840;
-  margin-top: 16px;
-  margin-bottom: ${({ large }) => (large ? '32px' : '16px')};
+  margin: 12px 0 16px;
 
   ${up('md')} {
+    margin: 16px 0 24px;
     ${TextStyles.headlineL}
   }
 `;
@@ -75,13 +75,12 @@ const HeaderBlockText = styled.p<LargeProps>`
   ${up('md')} {
     ${({ large }) => (large ? TextStyles.textR : TextStyles.textSR)};
   }
-
+  margin: 16px 0 0;
   letter-spacing: -0.01em;
   color: #202840;
-  margin-top: 32px;
 
-  ${down('sm')} {
-    margin-top: 24px;
+  ${up('md')} {
+    margin: 24px 0 0;
   }
 `;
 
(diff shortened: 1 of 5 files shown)


src/components/content/header-block/header-block.tsx
```diff
@@ -50,14 +50,14 @@ const Topline = styled.p`
   }
 `;
 
-const Headline = styled.h2<LargeProps>`
+const Headline = styled.h2`
   ${TextStyles.headlineM}
   letter-spacing: -0.01em;
   color: #202840;
-  margin-top: 16px;
-  margin-bottom: ${({ large }) => (large ? '32px' : '16px')};
```
hofmann-felix, line 58: There is no place in the design where the margin has to be changed because of `large`. Therefore, this condition is removed

src/components/pages/about-us/team.tsx
```diff
@@ -11,7 +11,7 @@ const TeamLayout = styled.div`
   margin-top: 48px;
 
   display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(160px, 1fr));
+  grid-template-columns: repeat(auto-fit, minmax(150px, 1fr));
```
hofmann-felix: Ensures that the team member grid has two columns also on small mobile devices (e.g. iPhone SE)

src/components/content/header-block/header-block.tsx
```diff
@@ -75,13 +75,12 @@ const HeaderBlockText = styled.p<LargeProps>`
...
   }
-
+  margin: 16px 0 0;
   letter-spacing: -0.01em;
   color: #202840;
-  margin-top: 32px;
 
-  ${down('sm')} {
-    margin-top: 24px;
+  ${up('md')} {
```
georgiee, line 82: Always better to go up then fix something down the breakpoint change. Thanks

src/components/pages/about-us/team.tsx
georgiee: We might lower that even more to have two columns on any mobile device. Which means to target device width of 320px. Right now it breaks into one column at 356px because of the space taken by the grid container and the gap it wants between the images. We probably want (320 - 48 - 8)/2 = 132px as the lower bounds ( for your convenience, I was in the mood to check on the numbers)

Screenshot 2023-03-10 at 08 48 32
Screenshot 2023-03-10 at 08 48 28

georgiee (approved):
Thanks for chose changes. I have one comment that you might want consider.

hofmann-felix:
@georgiee Thanks for checking the numbers. I applied your suggestion 👍

PR #727, opened 2023-03-09
fix: show adac logo with correct color

hofmann-felix:
### What's included?
* the ADAC Motorsport logo color on the landing page is adjusted
* URL: tbd

Before:
Bildschirm­foto 2023-03-09 um 14 20 11

After:
Bildschirm­foto 2023-03-09 um 14 19 57


georgiee (approved):
Well this doesn't need review, actually you could push changes like that directly to the target feature branch. We don't need to overdo it 🙏

PR #731, opened 2023-03-14
feat: allow new lines in hero

hofmann-felix:
### What's included?
* For the Hero Title there is now a requirement that new lines can be added manually. This is made possible by this PR.
* There may be other components (e.g. Teaser) in the future where new lines will be needed. Does it make sense to set `white-space: pre-line` globally or do we want to do this only in the affected components? 

### Preview
* URL: https://satellytescomfeatmaindesignupd-featheronewline.gatsbyjs.io/de/
Bildschirm­foto 2023-03-14 um 10 31 06

### DIFF
--- a/src/components/content/heroes/hero-text.tsx
+++ b/src/components/content/heroes/hero-text.tsx
@@ -28,6 +28,7 @@ const HeroTextStyled = styled.div`
 `;
 
 const Headline = styled.h1`
+  white-space: pre-line;
   ${TextStyles.headlineL}
   margin: 0;
   margin-bottom: 24px;
--- a/src/assets/locales/de/translations.json
+++ b/src/assets/locales/de/translations.json
@@ -1,5 +1,5 @@
 {
-  "main.title": "Ihre digitale Vision in Einklang bringen - mit modernster Technologie",
+  "main.title": "Ihre digitale Vision in Einklang bringen \n- mit modernster Technologie",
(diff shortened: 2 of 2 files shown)


src/components/content/heroes/hero-text.tsx
```diff
@@ -28,6 +28,7 @@ const HeroTextStyled = styled.div`
 `;
 
 const Headline = styled.h1`
+  white-space: pre-line;
```
georgiee, line 31: Nice that you leverage this css property. When this feature was discussed I already had br tags in the templates in mind and inject potentially unsafe html tags. I like the solution but I have some respect about the consequences. Do we have some disadvantages by swapping br with newlines? Or is it just equal and I learned some technique from you to use in the future 😄🙌

georgiee (approved):
Thanks, looks good and you deliver as requested. See my comment. 
Technically good and we can merge.

Regarding line breaks in general: It's a pattern fitting more static environments. Usually we want the browser to break lines. Otherwise we run into situations where this newline is breaking text in an unexpected or at least not desired way. Maybe not in this example but in bigger paragraphs were you suddenly have a single word on a line and then a line break already.

Here a contrived example by resizing the window and having a smaller font size. The word `bringen` hangs around pretty lonely. 

Screenshot 2023-03-15 at 13 15 34


While without your change it would look good:
Screenshot 2023-03-15 at 13 16 54


Should we do it or not? Maybe you can discuss that with GA one time? My approval is there in case you want to move forward.

hofmann-felix:
@georgiee Thanks for your thoughts on this. The sentence has now been changed, which means we no longer need the line break. I close the PR

PR #732, opened 2023-03-14
feat: show hero text on different positions

hofmann-felix:
### What's included?
* There is a new requirement for the Hero Text on the Career and About-Us page: The text in the Hero shouldn't be displayed in the Hero on mobile, but under the first heading.
* To hide the text in the hero, a flag `hideMobileText` was added to the hero
* In addition, there is an `MobileOnlyText` component that can be added under the corresponding headlines

### Preview
* URL: https://satellytescomfeatmaindesignupd-featherotext.gatsbyjs.io/de/career/
Bildschirm­foto 2023-03-14 um 14 04 48

### DIFF
--- a/src/components/content/heroes/aurora-hero.tsx
+++ b/src/components/content/heroes/aurora-hero.tsx
@@ -14,6 +14,7 @@ interface AuroraHeroProps extends HeroWithText {
 
 export const AuroraHero = ({
   title,
+  hideMobileText,
   children,
   auroraType,
 }: AuroraHeroProps) => {
@@ -22,7 +23,9 @@ export const AuroraHero = ({
       <AuroraFullSize type={auroraType} />
 
       <TextContainer>
-        <HeroText title={title}>{children}</HeroText>
+        <HeroText hideMobileText={hideMobileText} title={title}>
+          {children}
+        </HeroText>
       </TextContainer>
     </HeroContainer>
   );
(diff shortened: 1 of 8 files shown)


src/components/content/heroes/aurora-hero.tsx
```diff
@@ -22,7 +23,9 @@ export const AuroraHero = ({
       <AuroraFullSize type={auroraType} />
 
       <TextContainer>
-        <HeroText title={title}>{children}</HeroText>
+        <HeroText hideMobileText={hideMobileText} title={title}>
```
georgiee, line 26: Discussion. No action required:

Is this flexible in usage or every hero behaves like that? If so I would prefer not to make it an option.
Or if it's the default case and there is only one scenario to keep it we could flip it and say "swapMobile=false" or similar.

georgiee (approved):
Technically good and we can approve! One discussion point added.
Also it feels invasive to change the location of the text and put it, also semantically, in a different location on the page for mobile users.

I don't know how search engines handle that because it's kind of serving similar but different content to different users. Blind guessing though.

It's good for now but maybe we can check on the implications?

src/components/content/heroes/aurora-hero.tsx
hofmann-felix: The text should be hidden only on the Career and About-Us page. So I think it is okay to only hide the text with the active flag.

Regarding your concerns with the search engines: I did some quick research and all I found is that it's basically not a good idea to show different content on mobile and desktop. This can be interpreted as cloaking by search engines. For our specific case, where the content is completely identical, I have not found anything. Do you have any experiences from other projects or an idea how to solve this better?

georgiee: Thanks for the clarification and the research. Yeah it's the same content, so you're probably right it's fine. I was worried about the technically different structure and the slightly different hierarchy of the content that could get flagged. I think I worried too much 🙏

PR #735, opened 2023-03-16
chore: minor design changes

hofmann-felix:
### What's included?
* Too long short descriptions of job ads on the career page are now shortened with ellipsis:
Bildschirm­foto 2023-03-16 um 10 22 52

* The `max-width` of the leadbox text is increased on desktop since the whole page content has also a larger width:
Bildschirm­foto 2023-03-16 um 10 24 32

* New leadbox text on German About-Us page
* Preview: https://satellytescomfeatmaindesignupd-featminordesignchanges.gatsbyjs.io/


src/components/support/text-ellipsis.ts
```diff
@@ -0,0 +1,8 @@
+export const textEllipsis = (text, maxLength) => {
```
georgiee, line 1: We don't want to apply ellipsis through css, do we ? Because for wider boxes, more text would fit so we could trigger the ellipsis through css by enforcing the bounding box/height instead of truncating the actual text.

I bet you had already some thoughts on this, can you give me some context for the chose truncation approach?

georgiee (approved):
Looks good! I have one question but technically approved and good to merge.

src/components/support/text-ellipsis.ts
hofmann-felix: The only way I know to display ellipsis after multiple lines with CSS is something like this: 
```
display: -webkit-box;
-webkit-line-clamp: 9;
-webkit-box-orient: vertical;
overflow: hidden;
text-overflow: ellipsis;
```

I prefer the already existing solution with the `textEllipsis` function. If you know another way or prefer my CSS suggestion, please let me know.

PR #736, opened 2023-03-16
fix: make filed-dropper clickable in firefox

hofmann-felix:
### What was the problem?
* The file dropper wasn't clickable in Firefox (only drop worked)

### How it is solved?
* By adding a label inside the container the file dropper is also clickable in Firefox

Preview URL: https://satellytescommain21751-fixfiledropper.gatsbyjs.io/career/senior-ux-ui-designer/

### DIFF
--- a/src/components/forms/file-dropper/file-dropper.tsx
+++ b/src/components/forms/file-dropper/file-dropper.tsx
@@ -195,7 +195,6 @@ export const FileDropper = ({
     validator,
     onDropRejected,
     maxFiles,
-    useFsAccessApi: true,
   });
 
   const hasFiles = Boolean(currentFiles.length);
--- a/src/components/forms/file-dropper/file-dropper.tsx
+++ b/src/components/forms/file-dropper/file-dropper.tsx
@@ -76,6 +76,7 @@ const FileDropperContainer = styled.div<{
   hasFiles: boolean;
   hasErrors: boolean;
 }>`
+  position: relative;
   display: flex;
   gap: 16px;
   flex-direction: column;
@@ -109,6 +110,15 @@ const FileDropperContainer = styled.div<{
   &:hover {
     border: 1px solid ${theme.palette.text.default};
   }
+`;
+
+const StyledLabel = styled.label`
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  top: 0;
+  left: 0;
+  cursor: pointer;
 `;
 
 const Description = styled.p`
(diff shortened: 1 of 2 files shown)


src/components/forms/file-dropper/file-list-item.tsx
```diff
@@ -104,7 +104,7 @@ export const FileListItem = ({
             ))}
           </FileCategoryDropdown>
         )}
-        <RemoveButton onClick={() => onRemove(index)}>
+        <RemoveButton type="button" onClick={() => onRemove(index)}>
```
hofmann-felix, line 107: I noticed that when you click remove, the form is sometimes submitted by mistake. This shouldn't happen anymore.

georgiee, line 107: But is the type button not implicitly given by the underlying `button`. Why does this change anything?

src/components/forms/file-dropper/file-dropper.tsx
```diff
@@ -197,7 +208,9 @@ export const FileDropper = ({
         hasFiles={hasFiles}
         hasErrors={errors?.documents}
       >
-        <input {...getInputProps({ ...register(name) })} />
+        <StyledLabel>
+          <input {...getInputProps({ ...register(name) })} />
+        </StyledLabel>
```
georgiee, line 213: Do you know why that changes the behavior? I mean I understand the thing behind label + input but why does it not work without? How did you came to this fix?

georgiee (approved):
Thanks it works 🙌
Can you maybe try to answer my questions?

Technically good and approved!

src/components/forms/file-dropper/file-list-item.tsx
hofmann-felix: Before the change:
Bildschirm­foto 2023-03-16 um 15 22 36


After the change:
Bildschirm­foto 2023-03-16 um 15 22 57


Apparently `type="button"` is not set implicitly. Currently, if you upload two files on a version without the fix (e.g. https://satellytescommain21751-choreremovestorybook.gatsbyjs.io/career/senior-ux-ui-designer/) and remove the first one with the remove button, the form is submitted. With the fix this is no longer the case.

src/components/forms/file-dropper/file-dropper.tsx
hofmann-felix: I found out that there can be problems with file inputs in combination with `diaplay: none`. The solution with the label resulted from trial and error

PR #740, opened 2023-03-20
feat: add paragraphs to translated texts

hofmann-felix:
### What's included?
* The new content of the page needs new paragraphs in several places.
* To implement this, a hook was added that extends `useTranslation` with the function `tWithParagraphs`. This function checks the texts in translation.json for `\n` and...
  * inserts a `<br />` if a new line (= `\n`) is required
  * wraps the content in a new `<p>` tag if a new paragraph (= `\n\n`) is required 
* ⚠️ The function is currently only implemented in the [`Impressions` component of the AboutUs page](https://satellytescomfeatmaindesignupd-featparagraphs.gatsbyjs.io/de/about-us/) to show the functionality. If we decide to apply this solution, the new functionality will be added to the other pages. 

### Discussion / Other possibilities
There are also other ways to include new paragraphs:
* We could create a new key-value pair for each paragraph in `translation.json`. However, with many paragraphs, this becomes messy both in `translation.json` and in the JSX.
* We could set the CSS property `white-space: pre-line`. This will make CSS set new lines at `\n`. However, there are then no dedicated `<p>` tags in the DOM for the paragraphs. 

The solution with the extended hook does not have these disadvantages. It would be great to hear some opinions about the different solutions.

### Preview
URL: https://satellytescomfeatmaindesignupd-featparagraphs.gatsbyjs.io/de/about-us/

Bildschirm­foto 2023-03-20 um 11 25 47


src/components/i18n-helpers.tsx
```diff
@@ -30,3 +32,27 @@ export const useLocaleFormat = (dateFormat) => {
...
+
+  const tWithParagraphs = (key) => {
+    const texts = t(key).split(/\n\s*\n/);
+
+    return texts.map((text) => {
+      const textWithBreaks = text
+        .split(/\n/)
+        .map((textContent) => [
+          textContent,
+          <br key={textContent.substring(0, 7)} />,
```
georgiee: interesting keying approach. I'm not sure if the key is much better than a plain counter here though. Something you saw somewhere or an idea that you wanted to test? Refreshing to see you solving it like that  for sure ✨

src/components/i18n-helpers.tsx
```diff
@@ -30,3 +32,27 @@ export const useLocaleFormat = (dateFormat) => {
...
 };
+
+export const useTranslationParagraphs = (
+  ns?: string,
+  options?: UseTranslationOptions,
+) => {
+  const { t, ...useTranslationResponse } = useTranslation(ns, options);
+
+  const tWithParagraphs = (key) => {
+    const texts = t(key).split(/\n\s*\n/);
```
georgiee, line 43: Does this cover CR LF for Windows machines? Obviously not but I'm wondering how a Windows text would behave here?

https://en.wikipedia.org/wiki/Newline#Representation

src/components/i18n-helpers.tsx
```diff
@@ -30,3 +32,27 @@ export const useLocaleFormat = (dateFormat) => {
...
+
+    return texts.map((text) => {
+      const textWithBreaks = text
+        .split(/\n/)
+        .map((textContent) => [
+          textContent,
+          <br key={textContent.substring(0, 7)} />,
+        ])
+        .flat();
+      return <p key={text.substring(0, 7)}>{textWithBreaks}</p>;
```
georgiee: You calculate the same id for two different elements. The purpose of the key is to recognize the elements after a re-rendering but with two ids this will not happen. Maybe a counter is again easier?

src/components/pages/about-us/impressions.tsx
```diff
@@ -59,7 +59,7 @@ export const Impressions = ({ impressions }: ImpressionsProps) => {
         kicker={t<string>('about-us.impressions.title')}
       >
         <MobileOnlyText>{t('about-us.description')}</MobileOnlyText>
-        {t('about-us.impressions.text')}
+        {tWithParagraphs('about-us.impressions.text')}
```
georgiee, line 62: That's cool! I wonder if there are some middleware available in the transition engine to always convert newlines to paragraphs. I like the approach but I would prefer to play it more conservative:

+ Put it where we need it (like you started it) and do not sprinkle `tWithParagraphs` everywhere
+ If we want it everywhere, find a way to be not so invasive. Either a middleware or some other processing hook if possible. Last ressort is of course the customized `t` function as here.

georgiee (changes requested):
Can you check the key generation and the regex?
Thanks for your efforts 🙌

src/components/i18n-helpers.tsx
hofmann-felix: My thought was to avoid the index as key but in this case the index will probably not have any unwanted effects and in any case better than the risk to have the same keys. I removed both substring keys and added the index

src/components/i18n-helpers.tsx
hofmann-felix: Cool thank you for sharing this article. I knew there are differences but have never read about them in detail. 

Do we need to differentiate in our case? The new lines (e.g. insert `\n`) are added manually to translation.json. Or do I miss something?

src/components/pages/about-us/impressions.tsx
hofmann-felix: Yes my plan is also to use the new feature only when needed. Currently, according to the design, there are only two places anyway. I have added the function to the other place as well.

src/components/i18n-helpers.tsx
georgiee: When a windows user adds a newline the OS will insert `\r\n`.
@fischeversenker  can you jump in as a windows veteran 🙈 Can you help with this ?

I think the regex is not working once someone on a Windows machines edits the json file and inserts two newlines. The regex would not find the pattern `\r\n\r\n`, would it ?

fischeversenker: > When a windows user adds a newline the OS will insert `\r\n`.

...

> Do we need to differentiate in our case? The new lines (e.g. insert `\n`) are added manually to translation.json. Or do I miss something?

I have to agree with @hofmann-felix here. JSON does not support multi-line strings, so all line breaks will be added manually and will thus always be in the form of `\n`.
If we wanted to provide a bullet-proof solution that works with all sorts of inputs we would need to include a solution for Windows' `\r\n` line breaks, but in this case (assumption: we are only working with our very own JSON files) I claim that handling `\n` is enough.

georgiee: Oh I totally missed that fact. That we put in actual `\n` characters. Sorry for that confusion and thanks for the clarification.

src/components/i18n-helpers.tsx
```diff
@@ -30,3 +32,24 @@ export const useLocaleFormat = (dateFormat) => {
...
+
+  const tWithParagraphs = (key) => {
+    const texts = t(key).split(/\n\s*\n/);
+
+    return texts.map((text, index) => {
+      const textWithBreaks = text
+        .split(/\n/)
+        .map((textContent, index) => [textContent, <br key={index} />])
+        .flat();
+      return <p key={index}>{textWithBreaks}</p>;
```
georgiee, line 50: index 👍

georgiee (approved):
Let's merge this now. Thank you!

PR #743, opened 2023-03-21
fix(RSS): show blog article hero images on dev.to

hofmann-felix:
### What was the problem?
* We use our RSS feed to sync the blog articles with dev.to. All assets are displayed correctly except the hero images:
Bildschirm­foto 2023-03-21 um 16 23 49

### How it is solved?
* The image is now provided directly by contentful (without resize). The image is then displayed in dev.to before the actual content. To display it on dev.to as a cover image, the URL of the image must be manually added to the `cover_image` entry. I have tried several ways to load the image directly from the RSS as cover_image. Unfortunately without success. I think as long as dev.to does not implement this function, we have no influence on it.

### How to test?
* Currently the feature branch is not connected to our technical user on dev.to. To test the changes, you can connect your own dev.to account to the RSS feed. I have added the changes to one article from the technical user as a test and the image is displayed correctly now: https://dev.to/technicalusersatellytes/a-semantic-approach-to-buttons-more-2l1c-temp-slug-2550192?preview=0311b025c182bc4a976683d85cc49e485327d72073463e2f7610afd46b607d8647c80447eed6cdd6b57831c09da25e50e75e0fb67203a042d98a84ca
* To display the image as a cover image, you have to add the URL to the cover_image field as said before:
Bildschirm­foto 2023-03-21 um 16 51 50

### DIFF
--- a/gatsby/config-options/gatsby-plugin-feed-options.tsx
+++ b/gatsby/config-options/gatsby-plugin-feed-options.tsx
@@ -202,9 +202,8 @@ const feedOptions = {
     {
       serialize: ({ query: { site, allContentfulBlogPost } }) => {
         return allContentfulBlogPost.nodes.map((node) => {
-          const imageUrl = `${BASE_URL}${
-            node.heroImage.image?.resize?.src ?? DEFAULT_META_IMAGE_URL_PATH
-          }`;
+          const imageUrl = node.heroImage.image?.url;
+
           const coverImage =
             node.heroImage && node.heroImage.creator
               ? toHtml(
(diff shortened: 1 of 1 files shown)


georgiee (approved):
Nice, thanks for finding a proper fix for this 🙌 I did not test it myself but checked the linked post and otherwise trust your carefully written description.

What are the next steps? That's it? We can cross post now to dev.to ? How is it triggered?

hofmann-felix:
@georgiee Currently, all articles from dev.to are already in draft form. In the settings of the technical user there is a button which fetches all articles from our RSS feed and adds them as a draft. But I don't know if the existing drafts will be updated (= adopt the right hero image). If this is not the case, I would adjust the existing drafts manually. But for new articles this should be done automatically.

Then we actually just have to decide when to publish which literary delicacies :)

PR #757, opened 2023-06-01
fix(career): provide correct channel id to post messages

georgiee:
I renamed our internal channel and did not update the channel id. I've now included the internal channel id number and not the channel name to make it more robust. I will submit the form on the feature deployment.


georgiee:
works.
image

please approve 🙏

Phil147 (approved):
🚀

PR #759, opened 2023-06-19
feat: main design update

feedm3:
This PR includes the new design.

> Only open to not forget about it. Final merge date and reviews are not yet decided. Some todos are still open.

PR #769, opened 2023-08-28
chore: clean up

brataaaan:
Clean-up PR before main merge:
- remove obsolete markdown pages


georgiee (changes requested):
Thanks for the idea to put that clean up in top of the PR but I would prefer working inside the PR #759. There is not much more to cleanup anyway 🙏 Please close this PR.

PR #784, opened 2023-11-28
fix: adapt base url to netlify

milljoniaer:
### What was done?
The Base URL was previously adapted to gatsby cloud pr deployments. This PR changes that to the netlify pattern.

### How to verify?
Verify that on the [preview](https://deploy-preview-784--satellytes.netlify.app/) of this PR, the base URL of the meta images is the same as the preview URL:
image


gatsby/util/build-netlify-preview-url.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Get the Netlify Preview URL by previewId
+ * (this equals the github entity id, e.g. PR id).
+ * If no previewId is given the fallback URL gets return.
+ *
+ * On netlify the deployment urls follow a strict pattern:
+ * - https://deploy-preview-${PREVIEW_ID}--${DOMAIN_NAME}.netlify.app/
```
fischeversenker: This seems to be slightly out of sync with the code ("preview" vs. "review", "fallback URL"). I claim that the tests nicely describe how this function should behave and suggest completely removing this comment.

gatsby/config-options/constants.ts
```diff
@@ -1,10 +1,11 @@
-import { buildGatsbyCloudPreviewUrl } from '../util/build-gatsby-cloud-preview-url';
+import { buildNetlifyPreviewUrl } from '../util/build-netlify-preview-url';
 export const RSS_FEED_URL = '/blog/rss.xml';
 export const DEFAULT_META_IMAGE_URL_PATH = '/sy-share-image.jpg';
-export const GATSBY_SITE_PREFIX = process.env.GATSBY_SITE_PREFIX || '';
-export const BRANCH_PREVIEW_URL = buildGatsbyCloudPreviewUrl({
-  prefix: GATSBY_SITE_PREFIX,
+export const NETLIFY_DOMAIN_NAME = process.env.GATSBY_NETLIFY_DOMAIN_NAME || '';
```
georgiee, line 4: where is this env coming from? netlify itself or something we have to provide? Did you change the netlify settings?

gatsby/config-options/constants.ts
```diff
@@ -1,10 +1,11 @@
...
+export const BRANCH_PREVIEW_URL = buildNetlifyPreviewUrl({
```
georgiee, line 5: the preview being null is mostly the case for non-preview environments. It will then use `GATBSY_BASE_URL` (following line). Is this something we create or given by netlify? Asking to make sure this will work when we merge it, because right now I can only verify the preview part of this PR and I fear it's not working later on production. Do you have confidence that this will work then? Thanks!

georgiee (approved):
Hey thanks for finding and fixing this.
Can you share your level of confidence if this will work on production later too? 

Regarding the preview url I was able to verify everything works with a link tester tool like https://socialsharepreview.com

gatsby/config-options/constants.ts
milljoniaer: I have configured it in the netlify settings.

gatsby/config-options/constants.ts
milljoniaer: The `GATSBY_BASE_URL` (and also the `GATSBY_PREFIX` which I changed to `GATSBY_NETLIFY_DOMAIN_NAME`) was part of our environment configuration on gatsby cloud and got lost on the migration to netlify. I readded it in the netlify settings. Hence, it currently works again on production. This PR only fixes the feature branch BASE_URL. for the main branch the BRANCH_PREVIEW_URL will still be "null" and the `GATSBY_BASE_URL` will be used. And since that works currently on production, I am confident, that it will still work once this PR is merged. But I will monitor it.

PR #788, opened 2023-12-11
Chore/a11y fix

noa-santo:
This PR aims to improve accessibility on the satellytes website and make it WCAG AA compliant.

### Changes:
- Disabled the button for back in the pagination module when on the first site
- Fixed swapped aria labels for pagination buttons
- Added aria label to customer logos so screen readers can read which logos they are 
- Made customer logos tab accessible so it’s easier to navigate to them
- Edited labels of social links in the footer to clearly state that they will open another site in a new tab
- Made legal links and language switcher bigger on mobile so it's easier to use them
- Added aria label to link in header to state that it leads back to the homepage
- Slightly changed contrast between timestamps and background on blog posts
- Dot between legal links is now a `<li>` instead of a `<span>`
- Language selection now has the aria label on the correct element
- made rss button understandable with screen reader
- hidden the honypot fields in the form (made to combat bots) from screen reader
- add aria-lable to google maps link
- made pagination buttons tab accessible when not disabled
- increased text size of contact leadbox
- add aria-hidden to links in front of headings in rich contentful text (e.g. on the imprint page)
- added placeholder text into contact form
- better phrasing in contact form errors
- AA compliment color contrast on contact form error message color
- better phrasing on "Learn more >" button on landing page (Text is now "More about our services")


georgiee:
Hi @virus-rpi 
you requested a review for this PR. We briefly talked about splitting up the PR into smaller chunks to make it easier to review. Especially with accessibility it's probably a good idea as there is often not a real wrong or right.

You started splitting with https://github.com/satellytes/satellytes.com/pull/790 maybe you can continue so we can discard this PR in the end? Or do you want to give this PR into a review?

Thanks for your clarification

PR #791, opened 2024-01-09
fix: blog page accessibility problems

noa-santo:
### Changes:

- The < and > buttons now have the correct aria-lables. 
         Before: 
           <:   Next Page
           >:   Previous Page
          After:
           <:   Previous Page
           >:   Next Page
- When on the first/last page you can't press the next / previous page button anymore
- The button are now tab focusable if enabled
- Improved screen reader text of RSS button
          Before:
          "List 1 item"
          "link, item, image, rss.xml, list 2 item"
          After:
          "link, image, Subscribe to blog"
- Made the color contrast for the timestamp WCAG AA compliant


src/components/layout/theme.tsx
```diff
@@ -18,7 +18,7 @@ export const theme: DefaultTheme = {
       defaultDark: '#FFFFFF',
       secondary: 'rgba(32,40,64,0.5)',
       topline: '#3E61EE',
-      timestamp: 'rgba(0, 0, 0, 0.5)',
+      timestamp: 'rgba(0, 0, 0, 0.65)',
```
georgiee, line 21: 👍

src/components/pages/blog-post/follow-panel.tsx
```diff
@@ -19,7 +19,7 @@ const FollowPanelContainer = styled.div`
   font-size: 16px;
 
   ${up('md')} {
-    margin-top: 0px;
+    margin-top: 0;
```
georgiee, line 22: auto fix by your IDE or by purpose? Change is reasonable and okay but should not happen accidentally

src/components/ui/pagination/pagination.tsx
```diff
@@ -83,7 +84,8 @@ export const Pagination = ({
       <StyledLink
         onClick={onPreviousClick}
         $disabled={currentPage === 1}
-        aria-label="Next Page"
+        aria-label="Previous Page"
+        tabIndex={currentPage === 1 ? -1 : 0}
```
georgiee, line 89: Thanks for the catch it's really bad to have a button that goes nowhere. Your fix has a good idea: take it out of the tab list to prevent it from being accessed, but it's wrong as it won't remove it from the a11y tree. The navigator will still list it as an element (I used the Voice Over rotator and the list of links to check it there you can see Previous Page being listed).
The link should be disabled. This will remove it entirely from the a11y tree and the tab index will be removed automatically.

CleanShot 2024-01-16 at 17 37 36

src/components/ui/pagination/pagination.tsx
```diff
@@ -101,7 +103,8 @@ export const Pagination = ({
       <StyledLink
         onClick={onNextClick}
         $disabled={currentPage === amountOfPages}
-        aria-label="Previous Page"
+        aria-label="Next Page"
+        tabIndex={currentPage === amountOfPages ? -1 : 0}
```
georgiee, line 109: same as for the other button, we have to use `disabled` attribute here.

georgiee (dismissed):
Thanks for the catch! I have an important change to make and then we can merge it

src/components/pages/blog-post/follow-panel.tsx
noa-santo: When I pushed the code my local linter complained about it so I thought why not and removed the "px"

src/components/ui/pagination/pagination.tsx
noa-santo: There is no `disabled` attribute because the StyledLink is based on a link and only buttons can be disabled. I just added aria-hidden which should have the same effect in combination with the other changes I made to the button.

maxmarkus:
LGTM. It works as expected. The failing test is the same prettier `seo.tsx` issue as described in the previous PR. Once this is fixed, PR should be green and ready to merge. 

It is the same formatting change that occurred in the other PR, so I wonder if this is some formatting by your IDE, since prettier doesn't create that change.

maxmarkus (approved):

src/components/layout/seo.tsx
```diff
@@ -102,9 +102,8 @@ const SEO = ({
    * The i18n object (languages, language, originalPath, defaultLanguage)
    * is replaced with constants and locales data for building the alternate meta tag.
    */
-  const dataNode = locales.edges.find(
-    (e) => e.node.ns === 'translations',
-  )?.node;
+  const dataNode = locales.edges.find((e) => e.node.ns === 'translations')
+    ?.node;
```
maxmarkus: This formatting change leads to the failing test. `yarn format` will revert it to the correct format.

noa-santo: `yarn format` did not revert it. I did it manually but after I ran `yarn format` again it formatted it to the form that causes the test failure

noa-santo: It worked after running `yarn install --frozen-lockfile`

maxmarkus (approved):

PR #792, opened 2024-01-09
fix: contact page accessibility problems

noa-santo:
### Changes:
- Error message color changed from #FF0D35 to #EB0027 to make the color contrast WCAG AA compliant 
- Changes error messages so that they suggest a fix (e.g. "Your name is missing" to  "The field name is not filled in. Please fill in the field name to send the message.") 
- Made the honeypot fields aria-invisible so that they don't get read by screen readers
- Added a aria-label to the google maps link
- Add aria-hidden to mandatory note at the bottom of the contact form


src/assets/locales/en/translations.json
```diff
@@ -48,14 +48,17 @@
   "contact.info-link": "<0>Please use our contact form or send us an email to <1>info@satellytes.com</1></0>",
   "contact.info": "Please use our contact form or send us an email to beep@satellytes.com",
   "contact.name": "Name",
+  "contact.name-placeholder": "e.g. John Doe",
   "contact.email": "E-mail address",
+  "contact.email-placeholder": "e.g. your-email@adress.com",
```
georgiee: never ever use actual domains in test-data, dummy data etc because you an always end up sending some mail to that address. There is a domain example.com that is guaranteed never to be owned by anyone. Always use that domain for constructed emails addresses. Here you could use `email@example.com`

src/components/forms/text-input/text-input.tsx
```diff
@@ -66,6 +69,7 @@ export const TextInput = (
         $hasError={Boolean(errorMessage)}
         disabled={formState.isSubmitting}
         id={props.name}
+        placeholder={props.placeholder}
```
georgiee, line 72: can you explain why you add placeholders. Is that read by a screen reader? They are not adding much here because forms are pretty common and if I use a screen reader to fill out a form I probably now what to provide if the input is named "email" and such. Maybe we can remove it if there is not a deeper thought involved?

src/components/layout/theme.tsx
```diff
@@ -19,7 +19,7 @@ export const theme: DefaultTheme = {
       secondary: 'rgba(32,40,64,0.5)',
       topline: '#3E61EE',
       timestamp: 'rgba(0, 0, 0, 0.5)',
-      errorMessage: '#FF0D35',
+      errorMessage: '#EB0027',
```
georgiee, line 22: 👍

src/components/pages/contact/address.tsx
```diff
@@ -29,7 +29,10 @@ export const Address = () => {
       80331 München
       <br />
       <br />
-      <StyledLink to="https://g.page/satellytes?share">
+      <StyledLink
+        to="https://g.page/satellytes?share"
+        aria-label={'Open Satellytes Google Maps profile in a new tab'}
```
georgiee, line 34: I can see it with the VO rotor. Much better. The "open" is kind of weird because any link is kind of "to be opened" anyway? But I have no better proposal 

CleanShot 2024-01-16 at 17 53 45
CleanShot 2024-01-16 at 17 54 26

src/components/pages/contact/honeypot.tsx
```diff
@@ -14,14 +15,15 @@ export const HoneypotField = (
...
-      <label>
+      <label aria-hidden={true}>
         {props.label}
         <input
           id={props.name}
           type="text"
           {...field}
           tabIndex={-1}
           autoComplete="false"
+          aria-hidden={true}
```
georgiee: I think that aria hidden is not required at all.

> Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. 

via https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

src/components/pages/contact/honeypot.tsx
```diff
@@ -6,6 +6,7 @@ import { FormData } from './form';
 
 const HoneypotFieldContainer = styled.div`
   ${hideVisually()}
+  aria-hidden: true;
```
georgiee: That is wrong. aria hidden is not a valid css statement. Not needed as you apply aria hidden on the label already

src/templates/blog-post.tsx
```diff
@@ -46,7 +46,10 @@ export const Head = ({
   const { author, seoMetaText, title, publicationDate, heroImage } =
     data.contentfulBlogPost;
 
-  const shareImagePath = heroImage.shareImage.resize.src;
+  let shareImagePath = '';
+  if (heroImage && heroImage.shareImage && heroImage.shareImage.resize) {
+    shareImagePath = heroImage.shareImage.resize.src;
+  }
```
georgiee: please replace with a more idiomatic fix: 

```suggestion
  const  shareImagePath = heroImage.shareImage?.resize.src;
```

georgiee (dismissed):
Thanks for the good catches! Especially the honeypot thing is embarasing 😬 I've found some issues to fix, thanks!

src/components/forms/text-input/text-input.tsx
noa-santo: They are read by screen readers. But you're right they are a bit redundant to the labels. Should I remove them?

maxmarkus: I just tried it with and without placeholders and VoiceOver and it feels a bit better without. Its a hell lot of fast read text in between what it reads out as similar to "name, textfield, main" and "you are in a text field, by typing you enter text into that field". 
So I would vote for removing the place holders.

maxmarkus:
Idea: How about setting the whole map as aria-hidden? For visually impaired people it seems to bring no benefits, it has 5-6 links that can be tabbed and the readout is weird (screenshot below). Since we have the google maps link below, that might be enough.
 
screenshot of open streetmap zoom-in voice-over

Edit: After writing this, I realized that [leafletjs is accessible](https://leafletjs.com/examples/accessibility/). But the idea can still be valid.

noa-santo:
> Idea: How about setting the whole map as aria-hidden? For visually impaired people it seems to bring no benefits, it has 5-6 links that can be tabbed and the readout is weird (screenshot below). Since we have the google maps link below, that might be enough.
> 
> screenshot of open streetmap zoom-in voice-over
> Edit: After writing this, I realized that [leafletjs is accessible](https://leafletjs.com/examples/accessibility/). But the idea can still be valid.

i tried aria hiding the map but VoiceOver still reads it despite the aria-hidden=true.

virus-rpi:
> > Idea: How about setting the whole map as aria-hidden? For visually impaired people it seems to bring no benefits, it has 5-6 links that can be tabbed and the readout is weird (screenshot below). Since we have the google maps link below, that might be enough.
> > screenshot of open streetmap zoom-in voice-over
> > Edit: After writing this, I realized that [leafletjs is accessible](https://leafletjs.com/examples/accessibility/). But the idea can still be valid.
> 
> i tried aria hiding the map but VoiceOver still reads it despite the aria-hidden=true.

should i keep looking for a way to make the map aria-hidden or should i just leave it as it is, since leaflet is accessible

noa-santo:
I found a way to make the map aria-hidden but it is still tab focusable and i could not figure out how to remove it from the tab order (`tabIndex = {-1}` gets ignored)

maxmarkus (approved):
Despite the fact that the general description "see no evil monkey emoji, because it's embarrassing" feels a bit weird I am approving it. We might think about rewording anyway.

PR #793, opened 2024-01-15
fix: footer accessibility problems  

noa-santo:
## Changes:
- On mobile or narrow screens the text size of the language switcher and the legal links are increased from 14px to 18px
- Add a aria-label to the language switcher ("Sprachauswahl" / "Language selection")
- State that the social links will open a new tab in the aria-label (https://www.a11yproject.com/checklist/#identify-links-that-open-in-a-new-tab-or-window)
- Converted the <span> elements in the legal links list to <li> elements (Improves the Lighthouse accessibility score => better SEO)


src/templates/blog-post.tsx
```diff
@@ -46,7 +46,10 @@ export const Head = ({
   const { author, seoMetaText, title, publicationDate, heroImage } =
     data.contentfulBlogPost;
 
-  const shareImagePath = heroImage.shareImage.resize.src;
+  let shareImagePath = '';
+  if (heroImage && heroImage.shareImage && heroImage.shareImage.resize) {
```
maxmarkus: FYI, the equation could also be written in this short form:
```ts
  if (heroImage?.shareImage?.resize) {
```

Or the complete `if` block replaced by:
```ts
const shareImagePath = heroImage?.shareImage?.resize?.src || '';
```

src/components/layout/navigation/navigation.tsx
```diff
@@ -241,7 +244,7 @@ const Navigation: React.FC<NavigationProps> = ({
             {t('navigation.imprint')}
           </LegalLink>
         </li>
-        <span>·</span>
+        <li>·</li>
```
maxmarkus: Think we should add `aria-hidden` here to avoid it being counted as a menu element.
3 elements

maxmarkus (changes requested):
Looks fine so far, just some small remarks.

maxmarkus (approved):
LGTM, thank you for the changes.

PR #800, opened 2024-03-05
build: disable npm update cron job

georgiee:
Let's disable it for a while.

### DIFF
--- a/.github/workflows/update-npm.yml
+++ b/.github/workflows/update-npm.yml
@@ -5,8 +5,8 @@ on:
   workflow_dispatch:
 
   # every monday at 02:00
-  schedule:
-    - cron: '0 2 * * 1'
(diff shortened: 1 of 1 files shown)


fischeversenker (approved):
Sounds reasonable to me. Thanks for taking care of this 👍 

We could also enable Dependabot alerts here in GitHub to at least do manual updates in case of serious security concerns.

georgiee:
> We could also enable Dependabot alerts here in GitHub to at least do manual updates in case of serious security concerns.

The alerts introduced quite some noise in the past, that's why we switched to a manual upgrade cron job.
Security wise we're on the little safer side with our static content and lack of logins. I'm open for suggestions how to not let it rot entirely. We can discuss this internally 🙏

PR #805, opened 2024-09-02
fix (a11y): breadcrumb a11y issues on blog page

noa-santo:
Chages: 
- change breadcrumb text color from rgb(0, 0, 0, 0.5) to rgb(0, 0, 0, 0.65) to make color contrast WCAG AA compliant 
- add aria-label to breadcrumb container
- wrap breadcrumb container with nav landmark
- replace `overflow: hidden` with `overflow-x: clip` to only hide the overflow in the x direction
- replace `padding: 0` with `padding-left: 4px` so the outline isn't cut of at the left side
- add `transform: translateX(-4px)` to breadcrumb container to counteract the visual changes through the padding change
- add `display: inline-block` to the breadcrumb link to fix the shape of the focus outline   
 
 
Before: 
image
After (without `display: inline-block` added ):
image

### DIFF
--- a/src/components/layout/breadcrumb/breadcrumb.tsx
+++ b/src/components/layout/breadcrumb/breadcrumb.tsx
@@ -8,7 +8,9 @@ const BreadcrumbContainer = styled.ol`
   list-style: none;
   display: flex;
   flex-wrap: nowrap;
-  padding: 0;
+  overflow-x: clip;
+  padding-left: 4px;
+  transform: translateX(-4px);
 `;
 
 const BreadcrumbListItem = styled.li`
@@ -31,6 +33,10 @@ const BreadcrumbListItem = styled.li`
   }
 `;
 
+const BreadcrumbLink = styled(Link)`
+  display: inline-block;
+`;
+
 interface BreadcrumbProps {
   breadcrumbEntries: BreadcrumbEntry[];
 }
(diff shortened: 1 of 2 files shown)


georgiee (changes requested):
Hey thanks for that. 
It would help if you can describe the effects in addition to the changes because a11y is pretty abstract for many folks. Even with some experience I've issues following the actual motivation for your changes

I found two issues while the color change is good. 

1. I can see the difference with the colour and I understand the contrast thing. Fine!

CleanShot 2024-09-04 at 09 40 04

2. I do not understand the role "navigation" because that seems to remove the list + listitem role already given by the `ol` tag. I think a better alternative would be to just wrap the `ol` with an additional `nav` tag. That way people can use the semantics of the list (screen reader more easily navigate lists) plus the nav semantic, that a screen reader may use to tell about the significant navigation options on the page

CleanShot 2024-09-04 at 09 41 25


Ref:
+ https://a11y-style-guide.com/style-guide/section-navigation.html
+ https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role


3. I don't see any difference of the focus outline. Can you demonstrate? I also wonder why removing the overflow has no other impact. Why do we have it in the first place anyway?

---

You see that I have feedback to all three changes. I have the vague feeling that 3 very small PRs would have been much better here as you could focus on documenting and demonstrating the purpose of the change. On the other hand, it's quite an overhead. Let's go with this PR, fix the issues and then continue.

src/components/layout/breadcrumb/breadcrumb.tsx
```diff
@@ -8,7 +8,6 @@ const BreadcrumbContainer = styled.ol`
   list-style: none;
   display: flex;
   flex-wrap: nowrap;
-  overflow: hidden;
```
hofmann-felix, line 11: We need this to prevent an overflow of the `BreadcrumbContainer`, e.g. on this page on mobile: https://deploy-preview-805--satellytes.netlify.app/blog/post/backstage-the-developer-portal-framework-for-optimizing-your-development-processes/

Bildschirmfoto 2024-09-19 um 12 53 14

PR #806, opened 2024-09-02
fix (a11y): add nav landmark to social and legal links

noa-santo:
Changes: 
- add nav landmark to the legal links in the menu
- add nav landmark to the social links in the menu

### DIFF
--- a/src/assets/locales/de/translations.json
+++ b/src/assets/locales/de/translations.json
@@ -78,6 +78,8 @@
   "navigation.blog": "Blog",
   "navigation.contact": "Kontakt",
   "navigation.about-us": "Über Uns",
+  "navigation.legal-links": "Rechtliche Links",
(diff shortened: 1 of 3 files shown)


georgiee (changes requested):
Thanks for bringing this up! The idea is correct but I think there is an issue with the depth of the placement.

The `nav` tag is far too low by wrapping individual links. It should at least wrap the `LegalInks`. Maybe even higher to catch the social links.

Please gather some knowledge about navigation landmarks so you can decide properly. Also ensure each nav element has a unique aria-label or aria-labelledby attribute as we have more than one `nav` then.

Some starters I found:
+ https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/navigation.html
+ https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/

PR #810, opened 2024-09-09
fix (a11y): social link focus outline

noa-santo:
Changes: 
- added `display: inline-block` to SocialLink to fix the issue that the focus ouline around social links was invisible 

Before:
image

After:
image


hofmann-felix (approved):
Nice solution, works great 🚀

PR #811, opened 2024-09-09
fix (a11y): add aria hidden to navigation flyout when visually hidden

noa-santo:
Changes: 
- add `aria-hidden={!props.visible}` to FullscreenOverlay


hofmann-felix (approved):
A small change with a big impact 💪 Finally, the menu is not read out every time before the actual content

PR #812, opened 2024-09-09
fix (a11y): make pagination buttons keyboard accessible

noa-santo:
Changes: 
- made StyledLink a button component

### DIFF
--- a/src/components/pages/blog/blog-page.tsx
+++ b/src/components/pages/blog/blog-page.tsx
@@ -1,4 +1,3 @@
-import { navigate } from 'gatsby';
 import { useI18next, useTranslation } from 'gatsby-plugin-react-i18next';
 import React from 'react';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


hofmann-felix (changes requested):
The scope of this issue is _"Pagination buttons don’t work with keyboard"_ and this PR solves this problem 👏 Nice!

Nevertheless, I challenge the existing implementation of the pagination buttons. These buttons just navigate to new URLs and should therefore be visible in the DOM as `<a>` tags to reflect their behavior. I suggest adapting the implementation as follows:
1. In `blog-post.tsx`: Refactor `onNextClick` and `onPreviousClick` to strings which contain the URL (also adapt the names)
2. In `pagination.tsx`: 
   - Change the prop types from a function to string. 
   - `StyledLink` should be a `Link` and should use the `to` prop instead of a `onClick` listener

I don't see any drawbacks in this suggestion. Feel free to challenge it if you find any during implementation.

hofmann-felix (approved):
Nice refactoring ✨ And the pagination is keyboard accessible now

PR #814, opened 2024-09-09
fix (a11y): add accessible name to social buttons and copy button

noa-santo:
Changes:
- add aria labels to all blog post social buttons
- add aria label to the copy button
- add translation for these labels


hofmann-felix:
Awesome 🚀 I tested the changes with VoiceOver and it works great.

However, I don't feel like approving yet because I have a question about the translation keys: Since the blog is only in English, the keys are unnecessary and potentially misleading. I see there were already other translation keys for the blog before the changes. @satellytes/website-crew Does anyone know if there is a specific reason for this?

georgiee:
I think it's historical. We aimed to provide a bilingual experience throughout the entire website but for the blog we did not bother to serve translated content and decided to keep it in a single language even though the content is mixed (we redirect people from German to English on this page: https://www.satellytes.com/de/blog/). 

I think we can keep the PR as is as the keys are just not used right now - but ideally they are going to be used in a future iteration of the implementation (which is a separate topic). Thanks for pointing this out.

hofmann-felix (approved):
@georgiee Thanks for the clarification 🙏

Then this PR ist good to go from my side

PR #815, opened 2024-09-09
fix (a11y): hide accordion chevron from screen reader

noa-santo:
Chages: 
- add aria-hidden=true to chevron icon in accordion header


hofmann-felix (changes requested):
Thank you for taking care of this 🙏 

The `aria-hidden` doesn't appear in the DOM because the attribute isn't passed to the `Icon` component. 

Bildschirmfoto 2024-09-10 um 11 08 55

hofmann-felix (approved):
Awesome 🚀 I tested the changes with VoiceOver and the icon is no longer read out.

PR #816, opened 2024-09-10
fix (a11y): make privacy checkbox keyboard accessible

noa-santo:
Changes:
- make checkbox the actual checkbox instead of having a hidden real checkbox and a fake visual checkbox


src/components/forms/checkbox/checkbox.tsx
```diff
@@ -70,29 +67,19 @@ export const Checkbox = (
...
-        <HiddenCheckbox
+        <StyledCheckbox
           aria-required={true}
           disabled={formState.isSubmitting}
           {...field}
-          value={field?.value || false}
           {...props}
           id={props.name}
-          onChange={handleChange}
+          hasError={Boolean(fieldState?.error)}
```
hofmann-felix: Can you please change the name to `$hasError`? The `$`-prefix tells styled-components to not pass the prop to the DOM Element ([reference here](https://styled-components.com/docs/basics#adapting-based-on-props)). Otherwise, React isn't happy 😕
Bildschirmfoto 2024-09-13 um 11 10 48

hofmann-felix (changes requested):
Awesome changes 🚀 The checkbox is now keyboard accessible and the code has also become much simpler. Nevertheless, I found one small issue. Could you PTAL?

hofmann-felix (approved):
Nice! I can confirm that the console error is gone now

Issue #802, opened 2024-08-26
Accessibility Issues Overview


hofmann-felix:
# Priority

- 🔴 High Priority
- 🟡 Moderate Priority
- 🟢 Low Priority

# General Issues

- [x] 🔴 **Breadcrumb**: Background and foreground colors do not have a sufficient contrast ratio. ([WCAG Reference](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html))
    - **Proposed Solution:** Make the Breadcrumbs darker and check if the result matches with https://contrastchecker.com/
    - **PR**: https://github.com/satellytes/satellytes.com/pull/805
- [ ] 🔴 **Focus Outline** is cut of in many places
    <img width="306" alt="Bildschirmfoto 2024-08-23 um 17 08 47" src="https://github.com/user-attachments/assets/cc90a57e-c87c-467c-a030-55afcbc628e8">
    - **PR**: https://github.com/satellytes/satellytes.com/pull/805
    - Not fixed in:
        - TeaserContainer
- [ ] 🔴 **Images:** Provide useful alt texts
    - **Description:** We have set `alt=“”` for almost all images so that they are ignored by the screen reader. However, these images often offer added value and should therefore not be ignored. The alt text should be set for these images
    - **Assignee:** @hofmann-felix 
- [x] 🔴 **Legal Links:** Add `<nav>` tag to “Data Protection” & “Imprint” in the Menu
    - **PR**: https://github.com/satellytes/satellytes.com/pull/806
- [x] 🔴 **Navigation:** LinkedIn & GitHub Links aren’t highlighted on keyboard focus
    - **PR**: https://github.com/satellytes/satellytes.com/pull/810
- [x] 🔴 **Navigation: Add** `[aria-hidden="true"]` to flyout when it is visually hidden
    - **PR**: https://github.com/satellytes/satellytes.com/pull/811
- [x] 🔴 **Navigation:** Other content should not be focusable when the flyout is active ([WCAG Refernce](https://www.w3.org/WAI/WCAG22/Understanding/focus-order#:~:text=A%20Web%20page%20implements%20modal,the%20element%20following%20the%20button))
    - - **PR**: https://github.com/satellytes/satellytes.com/pull/829
[...]

PR #817, opened 2024-09-10
fix (a11): add aria-invalid to career form fields 

noa-santo:
Changes:
- set `aria-invalid` to `true` for all input components if an error message is there
- add `aria-describedby` for all input components to the id of the error message element if there is an error
- add `id` to all error message elements
- add `aria-labelledby` to checkbox if the label is set
- add `id` to checkbox label

Note:
Because the useForm is set to `mode: 'onSubmit'` the `aria-invalid` flag only changes after the first submit. Changing the mode to `'onChange'` would fix that but it would also change the behaviour of the form.


src/components/forms/text-input/text-input.tsx
```diff
@@ -66,6 +71,7 @@ export const TextInput = (
       <Input
         type={'text'}
         aria-required={true}
+        aria-invalid={Boolean(props.errors?.[props.name])}
```
hofmann-felix: I would reuse the existing error logic here and in the other input components for two reasons:
1. The visual error handling and the error handling for screen readers are consistent.
2. You don't need any changes in the parent and no new logic in the input component 
```suggestion
        aria-invalid={Boolean(errorMessage)}
```

src/components/forms/text-input/text-input.tsx
```diff
@@ -66,6 +71,7 @@ export const TextInput = (
...
         $hasError={Boolean(errorMessage)}
         disabled={formState.isSubmitting}
         id={props.name}
```
hofmann-felix, line 73: Here is another suggestion, you have to know yourself whether you want to implement it: There is also the [issue](https://github.com/satellytes/satellytes.com/issues/802#:~:text=Form%20Fields%3A%20aria%2Ddescribedby%20and%20aria%2Dlabeledby%20attributes%20should%20be%20connected%20by%20ID%20to%20an%20element%20that%20either%20has%20text%20description%20or%20an%20%22aria%2Dlabel%22%20attribute.) that the relevant elements of the entire input elements should be linked by id. I think this issue fits perfectly with the addition of the aria-invalid attribute, as the user then not only knows that the field is invalid, but also why it is invalid. Feel free to add this as part of this PR if you also think it belongs together. Here is a diff as an inspiration how the implementation could look like, but it needs to be tested again. 

```diff
diff --git a/src/components/forms/text-input/text-input.tsx b/src/components/forms/text-input/text-input.tsx
--- a/src/components/forms/text-input/text-input.tsx
+++ b/src/components/forms/text-input/text-input.tsx
@@ -67,13 +67,19 @@
         type={'text'}
         aria-required={true}
         $hasError={Boolean(errorMessage)}
+        aria-invalid={Boolean(errorMessage)}
         disabled={formState.isSubmitting}
         id={props.name}
         placeholder={props.placeholder}
         {...field}
         value={(field?.value as string) || ''}
+        aria-describedby={errorMessage ? `${props.name}-error` : undefined}
       />
-      {errorMessage && <StyledErrorMessage>{errorMessage}</StyledErrorMessage>}
+      {errorMessage && (
+        <StyledErrorMessage id={`${props.name}-error`}>
+          {errorMessage}
+        </StyledErrorMessage>
+      )}
     </div>
   );
 };

```

hofmann-felix (changes requested):
Thank you for taking care of this, works great 💪

However, I think that the implementation can still be improved a little. Could you please take a look at the two comments?

hofmann-felix (approved):
Works nice 🙌

hofmann-felix (changes requested):
I just noticed that **all** fields have the `aria-required={true}` attribute even if they are not required. Can you please fix this in the scope of this PR?

hofmann-felix (approved):
Awesome 🚀

PR #821, opened 2024-09-18
fix (a11y): mark all external links with aria label

noa-santo:
Changes:
- add `aria-label` attribute to ExternalLink component to indicate that the link will open in a new tab


hofmann-felix (approved):
Works great 🚀

PR #823, opened 2024-09-29
feat: test Cookiebot script

ipeke94:
- Testing Cookiebot script (Domain is already specified as www.satellytes.com / German option is already added)
- The account is connected to "ipek.guener@satellytes.com" (that's why I'm keeping in draf)
- Pushing to the production probably will occur analytics result to be shown

Configuration script -> https://consentcdn.cookiebot.com/consentconfig/422b90b3-20ea-49d6-9658-9c722c5b4d97/satellytes.com/configuration.js

image

cookieconsent

image


ipeke94:
Since we decided to go with "https://github.com/orestbida/cookieconsent?tab=readme-ov-file" option, I'm closing this PR.

PR #825, opened 2024-10-08
feat: add cookie consent v3

ipeke94:
Based on the conversation with Georgios, I added Cookie Consent script from here -> https://github.com/orestbida/cookieconsent?tab=readme-ov-file 

Note: We can add German translation version.

Kindly checkout the gif below:

cookie-consent-trial-1234


src/components/CookieConsent.tsx
```diff
@@ -0,0 +1,92 @@
...
+    } else {
+      setAnalyticsEnabled(false);
+      clearLeadinfoCookies();
+    }
+  };
+
+  const setLeadinfoCookies = () => {
+    const twoYears = 63072000;
+    const sessionValue = new Date().toISOString();
+    document.cookie = `_li_id=some_value; max-age=${twoYears}; path=/`;
```
ipeke94: We can add more meaningful values: https://help.leadinfo.com/en/add-leadinfo-to-your-cookie-banner-solution

package.json
```diff
@@ -95,6 +95,7 @@
     "stylis": "^4.3.1",
     "unified": "^10.1.1",
     "util": "^0.12.5",
+    "vanilla-cookieconsent": "3.0.1",
```
georgiee, line 99: 👍

src/@types/cookieconsent.d.ts
georgiee, line 1: Why manually, we have the `types` project (https://github.com/DefinitelyTyped/DefinitelyTyped)

I think those should be enough:
https://www.npmjs.com/package/@types/cookieconsent

src/components/CookieConsent.tsx
```diff
@@ -0,0 +1,92 @@
...
+const CookieConsentComponentV3 = () => {
+  const [analyticsEnabled, setAnalyticsEnabled] = useState(false);
+
+  useEffect(() => {
+    CookieConsent.run({
+      categories: {
+        necessary: {
+          enabled: true,
+          readOnly: true,
+        },
```
georgiee: we don't need that category, do we ? Seems like you copied that as seen in the docs?

src/components/CookieConsent.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  linkedCategory: 'necessary',
+                },
+                {
+                  title: 'Analytics Cookies',
+                  description:
+                    'We use analytics cookies to analyze website usage and improve our services.',
+                  linkedCategory: 'analytics',
+                },
+              ],
+            },
```
georgiee: 1. Is there no default language shipped?
2. Put them into a separate en file, I think that's manageable and makes it less noisy to work with the config.

src/components/CookieConsent.tsx
georgiee: The article explains the nature of the cookies involved. I don't think that we're supposed to set those cookies.
Instead you have to pull in the matching script through the scripts management of the cookie consent js I think ?

https://cookieconsent.orestbida.com/advanced/manage-scripts.html

I'm not sure how to distinguish the two different modes (with cookies and without). Maybe there are two different scripts possible or so?

--> If that's not a given use case to allow both scenarios (cookie and without cookies) then let's enable cookies in the backend and integrate the tracker only when the consent was given. Maybe we find a solution afterwards.

src/components/layout/leadinfo-script.tsx
```diff
@@ -1,20 +1,30 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
+
+let leadinfoInitialized = false;
+
+const LeadinfoScript = ({ enable }: { enable: boolean }) => {
+  const [script, setScript] = useState<HTMLScriptElement | null>(null);
```
georgiee: as said I think it's easier to use the script manager instead of manually bootstrapping it

georgiee (changes requested):
Hi thanks for the start,
I've provided some feedback regarding the integration.

src/components/layout/leadinfo-script.tsx
```diff
@@ -2,17 +2,26 @@ import { useEffect } from 'react';
 
 const LeadinfoScript = () => {
   useEffect(() => {
-    const script = document.createElement('script');
-    script.innerHTML = `
+    const leadinfoScript = document.createElement('script');
+    leadinfoScript.setAttribute('type', 'text/plain');
+    leadinfoScript.setAttribute('data-category', 'analytics');
```
ipeke94, line 7: In the leadInfo side, the cookies are already enabled.

image

As far as I see in the [documentation](https://cookieconsent.orestbida.com/advanced/manage-scripts.html) when we add these two attributes, that prevents script execution until consent is given.

Cookies are enabled:


image
Screenshot 2024-10-15 at 3 13 10 PM
Screenshot 2024-10-15 at 3 18 28 PM


----


Cookies aren't enabled:
 

Screenshot 2024-10-15 at 3 11 41 PM
Screenshot 2024-10-15 at 3 11 48 PM

src/components/layout/leadinfo-script.tsx
ipeke94: Hello, 

Thank you for the feedback and your time🙏. I tried to block/manage script tags according to the [document](https://cookieconsent.orestbida.com/advanced/manage-scripts.html) you sent. You can see some FAQs in my comment below. Accordingly, when we do this, different cookies are set when analytics data is accepted by Leadinfo. The cookies are listed below possibly related to  [li_id and li_ses cookies](https://help.leadinfo.com/en/add-leadinfo-to-your-cookie-banner-solution) in terms of tracking user sessions and identifying visitors. (like Google Analytics(_ga cookies) or HubSpot (__hstc, hubspotutk) or _hssrc session cookie)

Screenshot 2024-10-16 at 10 43 00 PM

Do you think we need to add detailed descriptions of these cookies under analytics or the relevant section?

PR #827, opened 2024-10-17
Fix: Change slack file upload method to uploadV2

### LINKED ISSUE #818: Slack API: files.upload retires in March 2025
The files.upload function of the Slack API, which is used in the Career Form for applications, can no longer be used from March 2025. Under this link you will find all information about this, as well as the necessary steps for migration: https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay

noa-santo:
# Change:
- use the new `uploadV2` method instead of the `upload` method 

https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay#prepare

### DIFF
--- a/src/api/career-form.ts
+++ b/src/api/career-form.ts
@@ -148,7 +148,7 @@ export default async function handler(
     });
 
     for (const file of files) {
-      await client.files.upload({
+      await client.files.uploadV2({
         channels: CHANNEL_ID,
         thread_ts: postMessageResult.ts,
         file: file.buffer,


hofmann-felix (approved):
LGTM 🚀 Thank you for taking care of this

Can you please change these three things before the merge:
1. Fix the typo in the PR description
2. Assign yourself to the related issue
3. Ensure that the commit message on main is lower case (`fix: change slack file upload method to uploadV2`). We should always try to be consistent with the other commits.

PR #828, opened 2024-10-17
chore: change twitter to x

### LINKED ISSUE #813: Twitter is still called Twitter instead of X
All labels still call Twitter "Twitter" instead of the new name "X" and the old logo is used as icon.
This should be updated.

noa-santo:
## Changes:
- change twitter to x in translation keys 
- change twitter to x in url
- update `react-share` lib to newest version
- change twitter icon to x icon
- remove facebook share option (requested by Gholam)


hofmann-felix (approved):
Works great 🎉 thank you for taking care of this

PR #829, opened 2024-10-17
fix (a11y): trap focus when navigation flyout is active

noa-santo:
## Changes:
- trap focus when navigation flyout is active


src/components/layout/header/menu-flyout.tsx
```diff
@@ -60,12 +60,74 @@ interface NavigationFlyoutProp {
...
   showLanguageSwitch?: boolean;
   setIsNavigationVisible: (b: boolean) => void;
+  burgerRef: React.RefObject<HTMLButtonElement>;
 }
 
 export const NavigationFlyout: React.FC<NavigationFlyoutProp> = (props) => {
+  const overlayRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const handleFocusOutside = (visible: boolean) => {
```
hofmann-felix: I think that the name of this function is not self-explanatory. I would change it to something like `preventFocusOfNonFlyoutContent`

src/components/layout/header/menu-flyout.tsx
```diff
@@ -60,12 +60,74 @@ interface NavigationFlyoutProp {
...
+      const lastFocusableElement =
+        focusableElements?.[focusableElements.length - 1];
+
+      const handleFocus = (event: KeyboardEvent) => {
+        if (event.key !== 'Tab') {
+          return;
+        }
+        if (
+          event.shiftKey &&
+          document.activeElement === props.burgerRef.current
```
hofmann-felix: I think we would achieve better readability if we add a few variables here, e.g. `shifKeyPressed`, `burgerElement` and `activeElement`

hofmann-felix (commented):
What a brilliant solution to this problem! Nice work 🚀 I tested the behavior on different pages and it works great 👍

I left you two more suggestions to improve readability. Can you please take a look and see if you have the same opinion?

src/components/layout/header/menu-flyout.tsx
```diff
@@ -60,12 +60,65 @@ interface NavigationFlyoutProp {
...
+          event.preventDefault();
+        }
+        if (shiftKeyPressed && activeElement === firstFocusableElement) {
+          props.burgerRef.current?.focus();
+          event.preventDefault();
+        }
+        if (!shiftKeyPressed && activeElement === lastFocusableElement) {
+          props.burgerRef.current?.focus();
+          event.preventDefault();
+        }
```
hofmann-felix: Can you please add the new constant here? 

```suggestion
        if (shiftKeyPressed && activeElement === firstFocusableElement) {
          burgerElement?.focus();
          event.preventDefault();
        }
        if (!shiftKeyPressed && activeElement === lastFocusableElement) {
          burgerElement?.focus();
          event.preventDefault();
        }
```

hofmann-felix (approved):
Thanks for applying my suggestions. There are still two places left where you can add the new constants. Please take a look at my comment before you merge

PR #831, opened 2024-11-06
fix: snow works now again

### LINKED ISSUE #801: Weather Easter Egg: Snow crashes site
Updating `styled-components` caused a bug in the weather easter egg. Because of an update in the way how to implement animations the site crashes when trying to display snowy weather.

Error description:
https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/utils/errors.md#12

> It seems you are interpolating a keyframe declaration (%s) into an untagged string. This was supported in styled-components v3, but is not longer supported in v4 as keyframes are now injected on-demand. Please wrap your string in the css`` helper which ensures the styles are injected correctly. See https://www.styled-components.com/docs/api#css

noa-santo:
fixes #801

changed sway animation according to https://styled-components.com/docs/api#css

### DIFF
--- a/src/components/content/heroes/aurora/weather-easter-egg/precipitation-effect.tsx
+++ b/src/components/content/heroes/aurora/weather-easter-egg/precipitation-effect.tsx
@@ -18,7 +18,6 @@ interface SnowflakeProps {
   left: number;
   top: number;
   animation_delay: number;
-  snowflake?: PrecipitationData;
 }
 
 interface SnowflakeContainerProps {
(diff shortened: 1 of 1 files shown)


src/components/content/heroes/aurora/weather-easter-egg/precipitation-effect.tsx
```diff
@@ -13,12 +13,12 @@ interface RaindropProps {
 interface SnowflakeProps {
   fallDuration: number;
   size: number;
-  animation: any;
   blur: number;
   transparency: number;
   left: number;
   top: number;
   animation_delay: number;
+  snowflake?: PrecipitationData;
```
hofmann-felix: What is the reason for adding the snowflake here? If I understand this correctly, `snowflake` is only used in `swayAnimation`. `swayAnimation` only needs `left` & `fallDuration`, which are both already in the props.

hofmann-felix (commented):
Works great and the fix arrives just in time for winter and the first snow in Munich ❄️ I left one minor comment. Could you PTAL?

hofmann-felix (approved):
Works great ☃

PR #832, opened 2024-11-07
fix: global navigation close/open button

konclave:
The button was hidden by the block with menu on the short screens.  
https://github.com/user-attachments/assets/2a24cd9c-255e-47a4-829e-0150bdc09494

The height of the button was quite small. I've increased it and made the active area a square.
SCR-20241107-jeba
SCR-20241107-jdyh


hofmann-felix (approved):
Works great 🚀 Thank you for recognizing and fixing this bug

PR #835, opened 2024-11-21
feat: add button to reconfigure cookie consent 

noa-santo:
This PR adds a button on the data privacy page that lets the use reopen the cookie consent banner and reconfigure the preferences.
This is required according to [Art. 7 GDPR](https://gdpr-info.eu/art-7-gdpr/) point 3.

Changes: 
- added button
- added translations for the button 

image


hofmann-felix:
Nice, works great from a technical point of view 🚀 And thank you for adding the link to the legal requirements in the PR description. 

I think we can improve the content itself. Just adding a button on the top of the page is maybe a little bit too pragmatic.  I rather imagine a small text explaining that we collect cookies for certain purposes, but only if the user agrees. This consent can be withdrawn at any time. Then you can add the button below the text. I think the `Quote` or the `Callout` component are very useful for this.

src/pages/data-privacy.tsx
```diff
@@ -5,18 +5,34 @@ import { Layout } from '../components/layout/layout';
...
+import styled from 'styled-components';
 
 interface DataPrivacyPageQueryProps {
   contentfulPage: ContentfulPage;
   locales: LocalesQueryProps;
 }
 
+const StyledText = styled.p`
+  margin: 0;
+  margin-bottom: 1rem;
```
hofmann-felix: Is there a specific reason for using `rem` here? If not, I would use `px` instead. This way, we are consistent with the existing code and the margin is always the same as the top and bottom padding of the `Callout`, not dependent on the root font size. 

```suggestion
  margin-bottom: 16px;
```

hofmann-felix (approved):
Thanks for the changes. The feature works great 🚀

PR #838, opened 2024-11-29
fix: center menu icon vertically

hofmann-felix:
### What was the problem?

The changes in https://github.com/satellytes/satellytes.com/pull/832 broke the alignment of the menu icon. Unfortunately, I missed that during the review. 

### How is it fixed?

The changes of https://github.com/satellytes/satellytes.com/pull/832 that do not affect the `z-index` are reverted. This way we have the best of both worlds. 

### Preview
Before this PR:
Screenshot 2024-11-29 at 14 39 39
After this PR:
Screenshot 2024-11-29 at 14 39 46

### DIFF
--- a/src/components/legacy/icons/header-icons/burger-menu.tsx
+++ b/src/components/legacy/icons/header-icons/burger-menu.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
+import { GRID_GAP_DESKTOP, GRID_GAP_MOBILE } from '../../../layout/theme';
+import { up } from '../../../support/breakpoint';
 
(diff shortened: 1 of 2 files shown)


src/components/legacy/icons/header-icons/burger-menu.tsx
```diff
@@ -44,7 +52,6 @@ const Wrapper = styled.div<{ $transition: boolean }>`
 
   .bar:nth-child(3) {
     width: 8px;
-    margin-bottom: 0px;
```
hofmann-felix, line 47: We don't need this anymore since it is covered by `gap`

feedm3 (approved):

